PR #29, opened 2026-03-10
refactor(schema): v2→v3 migration to add FK on scope_id

### LINKED ISSUE #10: chore: rebuild tables in future migration to add FK on scope_id for v1-migrated databases
## Context

`migrate_v1_to_v2` adds `scope_id INTEGER NOT NULL DEFAULT 1` via `ALTER TABLE`, which cannot include `REFERENCES scopes(id)` due to SQLite limitations. Fresh v2 databases have the FK constraint; migrated v1 databases do not.

## Impact

Migrated databases can accept orphan `scope_id` values that wouldn't pass validation on fresh databases. In practice, this is mitigated by the engine's `ScopeStore::ensure_path` always creating valid scopes before use, but the schema-level guarantee is missing.

## Proposed Fix

A future migration (v2→v3) should rebuild affected tables:
1. CREATE TABLE facts_new with full FK constraint
2. INSERT INTO facts_new SELECT * FROM facts
3. DROP TABLE facts
4. ALTER TABLE facts_new RENAME TO facts
5. Recreate indexes and triggers

Same for events, edges, summaries.

## Source

Super-review Round 2, GPT-5.4 finding #1.

dutiona:
## Summary

- Adds `migrate_v2_to_v3` that rebuilds `events`, `facts`, `edges`, and `summaries` tables with `REFERENCES scopes(id)` on `scope_id` columns
- Extends migration framework with `disable_foreign_keys` flag for table-rebuild migrations (SQLite requires FK off during DROP+RENAME)
- Runs `PRAGMA foreign_key_check` after rebuild to verify integrity
- Fixes `TABLES_DDL` to include FK on `events`, `edges`, `summaries` for fresh databases (only `facts` had it)
- FTS5 index and triggers are dropped before rebuild, recreated and repopulated after

## Test plan

- [x] `migrate_v2_to_v3_enforces_scope_fk` — verifies orphan scope_id accepted before migration, rejected after
- [x] `migrate_v2_to_v3_preserves_fts` — verifies FTS index rebuilt and triggers fire post-migration
- [x] `migration_v1_through_v3_preserves_data` — end-to-end v1→v2→v3 with data survival check
- [x] `migrate_v1_through_v3_runs_without_error` — full migration chain completes
- [x] `fresh_db_creates_latest_schema` — fresh DB creates v3 directly
- [x] All 164 tests pass (`cargo test --all-features`)
- [x] No new clippy warnings
- [x] Super-review (pending — codex rate limit reset at 17:33)

Fixes #10

### DIFF
--- a/src/store/schema.rs
+++ b/src/store/schema.rs
@@ -171,8 +171,7 @@ fn set_foreign_keys(conn: &Connection, enabled: bool) -> Result<()> {
     } else {
         "PRAGMA foreign_keys = OFF"
     };
-    let mut stmt = conn.prepare(sql)?;
-    let _ = stmt.query([])?;
+    conn.execute_batch(sql)?;
     Ok(())
 }
 
@@ -188,7 +187,21 @@ fn check_foreign_keys(conn: &Connection) -> Result<()> {
 }
 
 fn migrate_v1_to_v2(conn: &Connection) -> Result<()> {
-    conn.execute_batch(SCOPES_DDL)?;
+    // Frozen snapshot of SCOPES_DDL at v2 — do not reference the global constant,
+    // which may evolve in future schema versions.
+    conn.execute_batch(
+        "CREATE TABLE IF NOT EXISTS scopes (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            parent_id INTEGER REFERENCES scopes(id),
+            label TEXT NOT NULL,
+            depth INTEGER NOT NULL DEFAULT 0
+        );
+        CREATE INDEX IF NOT EXISTS idx_scopes_parent ON scopes(parent_id);
+        CREATE UNIQUE INDEX IF NOT EXISTS idx_scopes_parent_label
+            ON scopes(parent_id, label);
+        -- Insert root scope (sentinel). Only root has parent_id=NULL.
+        INSERT OR IGNORE INTO scopes (id, parent_id, label, depth) VALUES (1, NULL, 'root', 0);",
+    )?;
     conn.execute_batch(
         "ALTER TABLE facts ADD COLUMN scope_id INTEGER NOT NULL DEFAULT 1;
          ALTER TABLE edges ADD COLUMN scope_id INTEGER NOT NULL DEFAULT 1;
--- a/src/store/schema.rs
+++ b/src/store/schema.rs
@@ -139,14 +139,23 @@ pub fn migrate(conn: &Connection) -> Result<()> {
             if *disable_fk {
                 set_foreign_keys(conn, false)?;
             }
-            let tx = conn.unchecked_transaction()?;
-            migration(&tx)?;
-            set_config(&tx, "schema_version", &target.to_string())?;
-            tx.commit()?;
+            let result: Result<()> = (|| {
+                let tx = conn.unchecked_transaction()?;
+                migration(&tx)?;
+                set_config(&tx, "schema_version", &target.to_string())?;
+                tx.commit()?;
+                Ok(())
+            })();
             if *disable_fk {
+                // Re-enable FK checks unconditionally, even if migration failed.
+                // Transaction rollback already restored the data, but the
+                // connection-level PRAGMA must be restored explicitly.
                 set_foreign_keys(conn, true)?;
-                check_foreign_keys(conn)?;
+                if result.is_ok() {
+                    check_foreign_keys(conn)?;
+                }
             }
+            result?;
         }
     }
     Ok(())
@@ -218,7 +227,8 @@ fn migrate_v2_to_v3(conn: &Connection) -> Result<()> {
             session_id TEXT,
             scope_id INTEGER NOT NULL DEFAULT 1 REFERENCES scopes(id)
         );
-        INSERT INTO events_new SELECT * FROM events;
+        INSERT INTO events_new (id, timestamp, event_type, payload, source, session_id, scope_id)
+            SELECT id, timestamp, event_type, payload, source, session_id, scope_id FROM events;
         DROP TABLE events;
         ALTER TABLE events_new RENAME TO events;",
     )?;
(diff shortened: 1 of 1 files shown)


src/store/schema.rs
```diff
@@ -131,18 +133,47 @@ pub fn migrate(conn: &Connection) -> Result<()> {
         )));
     }
 
-    for (i, migration) in MIGRATIONS.iter().enumerate() {
+    for (i, (migration, disable_fk)) in MIGRATIONS.iter().enumerate() {
```
dutiona, line 136: If a migration fails or the transaction rollback occurs, `foreign_keys` remains `OFF` for the connection. Consider using a RAII guard to ensure `PRAGMA foreign_keys = ON` is restored even on error.

src/store/schema.rs
```diff
@@ -131,18 +133,47 @@ pub fn migrate(conn: &Connection) -> Result<()> {
...
 
+fn set_foreign_keys(conn: &Connection, enabled: bool) -> Result<()> {
+    let sql = if enabled {
+        "PRAGMA foreign_keys = ON"
+    } else {
+        "PRAGMA foreign_keys = OFF"
+    };
+    let mut stmt = conn.prepare(sql)?;
+    let _ = stmt.query([])?;
+    Ok(())
```
dutiona, line 175: Minor style point: `PRAGMA foreign_keys = ON/OFF` does not return rows and does not accept parameters. Using `conn.execute(sql, [])` would be more idiomatic and slightly more efficient than `prepare` + `query`, though it matches existing code in `set_pragmas`.

src/store/schema.rs
```diff
@@ -160,6 +191,106 @@ fn migrate_v1_to_v2(conn: &Connection) -> Result<()> {
...
+            timestamp TEXT NOT NULL,
+            event_type TEXT NOT NULL,
+            payload TEXT NOT NULL DEFAULT '{}',
+            source TEXT NOT NULL,
+            session_id TEXT,
+            scope_id INTEGER NOT NULL DEFAULT 1 REFERENCES scopes(id)
+        );
+        INSERT INTO events_new SELECT * FROM events;
+        DROP TABLE events;
+        ALTER TABLE events_new RENAME TO events;",
```
dutiona, line 250: Using `SELECT *` in the migration is fragile as it assumes the column order in the old table perfectly matches the new `events_new` definition. Explicitly listing the columns is safer and prevents issues if the schema was previously modified or if SQLite reorders things.

src/store/schema.rs
```diff
@@ -160,6 +191,106 @@ fn migrate_v1_to_v2(conn: &Connection) -> Result<()> {
...
+    )?;
+
+    // 6. Recreate FTS5 and repopulate from rebuilt facts table
+    conn.execute_batch(FTS5_DDL)?;
+    conn.execute_batch("INSERT INTO facts_fts(rowid, content) SELECT id, content FROM facts;")?;
+
+    // 7. Recreate triggers and indexes (DROP TABLE removed them)
+    conn.execute_batch(TRIGGERS_DDL)?;
+    conn.execute_batch(INDEXES_DDL)?;
+
```
dutiona, line 364: Using global DDL constants like `INDEXES_DDL` and `TRIGGERS_DDL` inside a migration is risky. If these constants are updated in future schema versions (e.g., adding an index to a column not yet present in v3), the v2->v3 migration will fail when run on older databases. It is safer to hardcode the DDL state for v3 or use a versioned DDL system.

dutiona (commented):
The migration to v3 correctly addresses the need for foreign key constraints on `scope_id` columns, which were missing in v2. The use of table rebuilding (recreate-and-swap) is the correct approach for SQLite since `ALTER TABLE` is limited.

Key points to address:
1. **FK Safety:** If a migration fails, foreign keys remain disabled on the connection. A RAII-style guard would make this more robust.
2. **Migration Robustness:** The migration depends on current DDL constants (`INDEXES_DDL`, etc.), which may break future runs of this migration if those constants change.
3. **Data Integrity:** `SELECT *` in the migration is slightly fragile; explicit column lists are preferred for schema evolution.

The logic for FTS repopulation and trigger recreation looks correct, and the added tests provide good coverage for both fresh installs and migrations from v1/v2.

src/store/schema.rs
```diff
@@ -131,18 +133,56 @@ pub fn migrate(conn: &Connection) -> Result<()> {
...
+    let mut stmt = conn.prepare("PRAGMA foreign_key_check")?;
+    let mut rows = stmt.query([])?;
+    if rows.next()?.is_some() {
+        return Err(MemoryError::Migration(
+            "foreign key violations detected after table rebuild".to_string(),
+        ));
+    }
+    Ok(())
+}
+
```
dutiona, line 188: Minor: This migration still references the global `SCOPES_DDL` constant. To ensure this migration remains a frozen snapshot (as established for v2→v3), it would be better to inline the DDL or use a versioned constant like `SCOPES_V1_DDL`. This prevents future changes to the `scopes` table from leaking back into the v1→v2 migration logic.

dutiona (commented):
Round 2 re-review complete. 

The updates successfully address all Round 1 findings:
1. **FK Safety:** Table rebuilds are now wrapped in a closure that ensures  is restored unconditionally, even if a migration fails. This is a robust approach that prevents the database from remaining in an unsafe state.
2. **Explicit Columns:** All  statements in the migration now specify explicit column lists. This is a critical fix to prevent silent data corruption if column order or schema changes in future versions.
3. **Frozen Snapshots:** The  migration now correctly inlines the DDL for FTS5 virtual tables, triggers, and indexes as strings, ensuring it remains a frozen snapshot of the schema at that point in time.

One minor point noted:  still references the global  constant, which technically violates the frozen snapshot principle for that older migration. However, this is outside the immediate scope of the current PR's fixes.

Overall, the code is well-structured and the migration logic is safe and correct.

dutiona (commented):
Round 2 re-review complete. 

The updates successfully address all Round 1 findings:
1. FK Safety: Table rebuilds are now wrapped in a closure that ensures PRAGMA foreign_keys = ON is restored unconditionally, even if a migration fails. This is a robust approach that prevents the database from remaining in an unsafe state.
2. Explicit Columns: All INSERT INTO ... SELECT statements in the migration now specify explicit column lists. This is a critical fix to prevent silent data corruption if column order or schema changes in future versions.
3. Frozen Snapshots: The v2_to_v3 migration now correctly inlines the DDL for FTS5 virtual tables, triggers, and indexes as strings, ensuring it remains a frozen snapshot of the schema at that point in time.

One minor point noted: migrate_v1_to_v2 still references the global SCOPES_DDL constant, which technically violates the frozen snapshot principle for that older migration. However, this is outside the immediate scope of the current PR's fixes.

Overall, the code is well-structured and the migration logic is safe and correct.

dutiona:
**Codex R2 finding (posted manually — Codex couldn't reach GitHub API):**

> FK check happens AFTER `tx.commit()`. If a v2 DB has pre-existing orphan `scope_id` values, the table rebuild copies them with FK checks OFF, the commit persists v3 + bad data, and only then does `foreign_key_check` fail. DB is stuck at v3 with invalid rows.

**Fixed in 609442e:** `check_foreign_keys()` now runs inside the transaction before `tx.commit()`. A failed check rolls back the entire migration, leaving the DB at v2. Added test `migrate_v2_to_v3_rejects_orphan_scope_ids` to verify.

PR #61, opened 2026-03-13
feat(store): schema evolution discipline (#18)

### LINKED ISSUE #18: Deferred: Schema evolution discipline — versioning policy and backwards-compat
## What it solves

Formal rules for how the schema evolves across versions: migration policies, backwards-compatibility testing, event envelope versioning.

## Trigger

Before Phase 4 ships (Phase 4 introduces MCP server with external consumers who depend on stable schema).

## Scope

- Schema version numbering policy (semver? monotonic?)
- Migration rules: forward-only (current), but need backwards-compat story for downgrades
- Event envelope versioning: how to handle events from different schema versions in the same log
- Backwards-compat testing: automated tests that verify old data migrates correctly
- Documentation: migration guide for consumers upgrading across versions

## What it does NOT solve

- Does not handle data format changes in consumer-provided embeddings (that's the consumer's problem)
- Does not solve wire protocol versioning (MCP server handles that separately)

## Hints

- GPT-5.4 identified this as a blind spot
- Phase 3 already has a schema migration framework (versioned, forward-only)
- Key question: can we guarantee event log replay produces identical state across schema versions?

## Design doc

See `docs/plans/2026-03-09-future-phases-design.md` — Deferred Items, GPT-5.4 Blind Spots

dutiona:
## Summary

- **Storage epoch versioning**: coarse-grained compatibility gate (`STORAGE_EPOCH=1`) stored in config table. Rejects databases from future epochs with `MemoryError::UnsupportedEpoch`
- **WAL-safe pre-migration backup**: `VACUUM INTO` creates atomic, consistent backup before migrations. Opt-in via `EngineConfig::backup_dir`
- **Event envelope versioning**: migration v4→v5 adds `event_revision INTEGER NOT NULL DEFAULT 1` to events table. Per-event-type revision tracking via `UpcasterRegistry`
- **Raw vs upcasted reads**: `get()`/`list()` unchanged (audit-log semantics). New `get_upcasted()`/`list_upcasted()` apply upcaster chains at read time
- **UpcasterRegistry**: `fn(Value) -> Result<Value>` functions keyed by `(event_type, from_revision)`, chained sequentially for multi-step upgrades
- **Migration testing infrastructure**: insta snapshot test (deterministic DDL projection), 3 proptest property-based tests (row count, content hashes, FK integrity), frozen V4 DDL snapshot
- **Schema evolution policy**: formal document covering version axes, forward-only policy, backup, envelope versioning, and testing requirements

## Stats

- 18 files changed, 1381 insertions, 69 deletions
- 214 tests pass (187 original + 27 new)
- 3 new files: `src/store/upcaster.rs`, `docs/schema-evolution-policy.md`, insta snapshot

## Design decisions

Based on 3-way AI research consensus (Claude/Codex/Gemini) documented in issue #18:
1. Forward-only migrations with pre-migration backup (no rollback code)
2. `event_revision` on `Event` only (read side) — `NewEvent` unchanged, store stamps internally
3. Explicit raw/upcasted read separation preserves audit-log integrity
4. Storage epoch as coarse-grained break-glass for architectural changes

## Test plan

- [x] All 214 tests pass (`cargo test`)
- [x] No new clippy warnings from changed code
- [x] Migration chain v1→v2→v3→v4→v5 verified
- [x] Property-based tests: row count preserved, content hashes survive, FK integrity clean
- [x] Insta snapshot accepted for v5 schema
- [x] Upcaster chain: single, sequential, gap detection, noop at latest
- [x] Raw reads return stored data unchanged; upcasted reads apply transforms

Closes #18

### DIFF
--- a/src/store/events.rs
+++ b/src/store/events.rs
@@ -1,7 +1,8 @@
 use chrono::{DateTime, Utc};
-use rusqlite::{Connection, params};
+use rusqlite::{params, Connection};
 
 use crate::error::{MemoryError, Result};
+use crate::store::upcaster::UpcasterRegistry;
 use crate::types::{Event, EventType, NewEvent};
 
 /// Filter for querying events.
@@ -16,6 +17,7 @@ pub struct EventFilter {
 /// Store for the append-only event log.
 pub struct EventStore<'a> {
     conn: &'a Connection,
+    registry: &'a UpcasterRegistry,
 }
 
 const fn event_type_to_str(et: &EventType) -> &'static str {
@@ -74,6 +76,8 @@ fn row_to_event(row: &rusqlite::Row<'_>) -> rusqlite::Result<Event> {
         })
         .transpose()?;
 
+    let event_revision: i64 = row.get("event_revision")?;
+
     Ok(Event {
         id: row.get("id")?,
         timestamp,
@@ -85,14 +89,15 @@ fn row_to_event(row: &rusqlite::Row<'_>) -> rusqlite::Result<Event> {
         origin_node_id: row.get("origin_node_id")?,
         sequence_id: row.get("sequence_id")?,
         created_at,
+        event_revision: u16::try_from(event_revision).unwrap_or(1),
     })
 }
 
 impl<'a> EventStore<'a> {
-    /// Create a new `EventStore` borrowing the given connection.
+    /// Create a new `EventStore` borrowing the given connection and upcaster registry.
     #[must_use]
-    pub const fn new(conn: &'a Connection) -> Self {
-        Self { conn }
+    pub const fn new(conn: &'a Connection, registry: &'a UpcasterRegistry) -> Self {
+        Self { conn, registry }
     }
 
     /// Insert a new event, returning its auto-assigned id.
@@ -109,8 +114,8 @@ impl<'a> EventStore<'a> {
 
         self.conn.execute(
             "INSERT INTO events (timestamp, event_type, payload, source, session_id, scope_id,
-                origin_node_id, sequence_id, created_at)
-             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9)",
+                origin_node_id, sequence_id, created_at, event_revision)
+             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10)",
             params![
                 timestamp_str,
                 event_type_str,
@@ -121,6 +126,7 @@ impl<'a> EventStore<'a> {
                 event.origin_node_id,
                 event.sequence_id,
                 created_at_str,
+                i64::from(self.registry.latest_revision(event_type_str)),
             ],
         )?;
         Ok(self.conn.last_insert_rowid())
@@ -134,7 +140,7 @@ impl<'a> EventStore<'a> {
     pub fn get(&self, id: i64) -> Result<Event> {
         let mut stmt = self.conn.prepare(
             "SELECT id, timestamp, event_type, payload, source, session_id, scope_id,
-                    origin_node_id, sequence_id, created_at
+                    origin_node_id, sequence_id, created_at, event_revision
              FROM events WHERE id = ?1",
         )?;
         let mut rows = stmt.query_map(params![id], row_to_event)?;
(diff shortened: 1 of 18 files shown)


src/store/schema.rs
```diff
@@ -166,9 +212,54 @@ pub fn migrate(conn: &Connection) -> Result<()> {
...
+    }
+
+    let db_name = Path::new(&db_path)
+        .file_name()
+        .unwrap_or_default()
+        .to_string_lossy();
+    let backup_path = backup_dir.join(format!("{db_name}.v{current_version}.bak"));
+
+    // VACUUM INTO creates a standalone, defragmented copy — WAL-safe
+    let sql = format!("VACUUM INTO '{}'", backup_path.to_string_lossy());
```
dutiona: [Gemini Review] SQLite's `VACUUM INTO` will fail if the destination file already exists. If a migration fails (e.g. midway) and the application restarts, the next attempt will try to write to the same backup file `*.vX.bak` and fail. This will permanently block startup until the user manually deletes the backup file. Consider checking if the file exists and removing it, or adding a timestamp to the filename (e.g., `{db_name}.v{current_version}.{timestamp}.bak`).

src/store/upcaster.rs
```diff
@@ -0,0 +1,225 @@
...
+        if current_revision >= latest {
+            return Ok((payload, current_revision));
+        }
+
+        let mut value = payload;
+        let mut rev = current_revision;
+
+        while rev < latest {
+            let key = (event_type.to_string(), rev);
+            let func = self.upcasters.get(&key).ok_or_else(|| {
```
dutiona, line 100: [Gemini Review] Performance: `event_type.to_string()` allocates a new String on every iteration of this loop. Since this is called per-event during `list_upcasted`, it will cause significant memory allocation overhead for large event logs. Consider storing the event type as a `Cow<'static, str>` or finding a way to borrow the key (e.g., using a wrapper struct with `Borrow` or doing the lookup differently) to avoid per-event-per-revision allocations.

dutiona (commented):
[Gemini Review] Overall summary: Good implementation of schema evolution discipline. I left a couple of inline comments on  regarding backup failure recovery and  regarding performance. Otherwise, the testing and design look solid!

dutiona (commented):
[Gemini Review] Overall summary: Good implementation of schema evolution discipline. I left a couple of inline comments on `src/store/schema.rs` regarding backup failure recovery and `src/store/upcaster.rs` regarding performance. Otherwise, the testing and design look solid!

src/store/events.rs
```diff
@@ -85,14 +89,15 @@ fn row_to_event(row: &rusqlite::Row<'_>) -> rusqlite::Result<Event> {
         origin_node_id: row.get("origin_node_id")?,
         sequence_id: row.get("sequence_id")?,
         created_at,
+        event_revision: u16::try_from(event_revision).unwrap_or(1),
```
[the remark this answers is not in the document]
dutiona: Fixed: Changed to `u16::try_from(event_revision).map_err(|_| FromSqlConversionFailure(...))` — now fails with an explicit error instead of silently defaulting to revision 1 on corrupted data.

src/store/schema.rs
dutiona: Fixed: Now removes existing backup file with `std::fs::remove_file` before `VACUUM INTO`. Also added single-quote escaping for SQL safety (SQLite VACUUM INTO does not support parameterized paths).

src/store/upcaster.rs
dutiona: Fixed: Moved `event_type.to_string()` outside the while loop (1 allocation total instead of N). Per-iteration cost is now just a `String::clone()` for the HashMap key lookup — negligible for short event type names relative to the JSON upcaster transform work.

dutiona (commented):
[Gemini Review] LGTM. Round 2 review of commit 2375a7e confirms that all previous feedback has been addressed:
- `backup_before_migration` now handles existing backup files gracefully by removing them before calling `VACUUM INTO`.
- SQL path escaping (single quotes) is now handled for the `VACUUM INTO` command.
- Metadata and epoch parsing are now fail-closed with appropriate `MemoryError` returns.
- Event revision parsing now includes explicit column-index error reporting.
- `UpcasterRegistry::upcast` has been optimized to reduce string allocations.

All 214 tests pass. Great work on the schema evolution discipline!

PR #69, opened 2026-03-19
feat(bootstrap): session log bootstrap — JSONL to historical memory (#43)

### LINKED ISSUE #43: Phase 4a: Session log bootstrap — JSONL to historical memory
### Scope

One-time migration tool: parse existing Claude Code JSONL session logs into memory-engine as historical facts.

1. Parse session files, extract episodes with timestamps
2. Load into memory-engine as historical facts (backdated `t_created`)
3. Run retroactive dream-cycle consolidation (Phase 5 dependency for full pipeline)
4. Surface patterns for human review

### Partial standalone value

Even without Phase 5 dream-cycle, the bootstrap populates the engine with historical context for `resume_context()` and `query()`.

### Research reference

Four-layer architecture § Bootstrap Strategy.

dutiona:
## Summary

- Parse Claude Code JSONL session logs and import noteworthy episodes as historical facts
- 5-stage pipeline: parse → reconstruct turns → classify outcome → keyword pre-filter → extract facts
- Research-backed: success-gated ingestion (AWM), workflow extraction (APC), pre-warming semantics
- Savepoint-wrapped per-session transactions for crash safety
- Idempotent via marker events (source="bootstrap")
- `SessionExtractor` trait for consumer-provided LLM extraction; `KeywordExtractor` default (no LLM)
- Backdated `t_created` and `last_accessed` for correct Ebbinghaus decay on historical facts
- 47 new unit tests + 8 integration tests, 263 total (0 failures)

## Changes

| Commit | Description |
|--------|-------------|
| 1 | Extend `AddFactOptions` with `t_created`/`last_accessed`, add `source` to `EventFilter`, add `Io`/`Bootstrap` error variants |
| 2 | JSONL parser (`parse.rs`) + 3 test fixtures |
| 3 | Turn reconstruction + NeuroStack keyword pre-filter (`filter.rs`) |
| 4 | Heuristic session outcome classification (`outcome.rs`) |
| 5 | `SessionExtractor` trait + `KeywordExtractor` (`extract.rs`) |
| 6 | Pipeline orchestration + engine facade methods (`mod.rs`, `metrics.rs`) |
| 7 | Integration tests (`bootstrap_test.rs`) |
| 8 | Usage guide, ADR-0009, reference doc updates |

## Review findings addressed (from Codex gpt-5.4 + Gemini 3)

- **[C1 BLOCKER]** Added savepoint transactions for crash safety
- **[C2/G2 HIGH]** Extended `EventFilter` with `source` field for idempotency
- **[C3/G1 HIGH]** Backdated `last_accessed` alongside `t_created`
- **[C4 HIGH]** `SessionExtractor::extract()` returns `Result`
- **[C6 MEDIUM]** Added typed `Io(std::io::Error)` error variant
- **[C7 MEDIUM]** Facts use `source_event_id` for provenance

## Test plan

- [x] `cargo test bootstrap` — 47 unit tests pass
- [x] `cargo test --test bootstrap_test` — 8 integration tests pass
- [x] `cargo test` — 263 total tests, 0 failures
- [x] `cargo check` — clean compilation, 0 warnings
- [x] Run `/super-review` for multi-model PR review

Closes #43


dutiona:
All 7 review findings addressed in 352943b:
- **C-P1 (savepoint):** Added RELEASE after ROLLBACK TO
- **C-P1/G-high (tool linkage):** Rewrote extract_tool_calls with positional pairing via parent_uuid
- **C-P2 (unreadable files):** File::open now caught before session-level match
- **G-medium (contains ash):** Replaced with eq_ignore_ascii_case
- **G-medium (if/else chain):** Replaced with match
- **G-medium (weak assertion):** Now asserts turns_reconstructed <= 1

dutiona:
Implementation plan published as #74.

src/bootstrap/filter.rs
```diff
@@ -0,0 +1,697 @@
...
+                if let Some(ref u) = tr.uuid {
+                    used.insert(u.as_str());
+                }
+            }
+        }
+    }
+
+    // --- Sequential fallback for unpaired entries ---
+    let mut i = 0;
+    while i < relevant.len() {
```
dutiona, line 139: [MEDIUM] The sequential fallback logic for unpaired entries relies on find from the current index i+1. This is safer than simple pairing, but could still be fragile if the session log contains consecutive user entries (e.g., from subagents or multiple prompts). The used set helps, but a more robust turn-graph construction might be needed in Phase 5.

src/bootstrap/outcome.rs
```diff
@@ -0,0 +1,427 @@
...
+            _ => {
+                consecutive = 1;
+            }
+        }
+    }
+    false
+}
+
+/// Proxy for interruption: the very last tool call's `is_error` flag.
+fn detect_interrupted(turns: &[super::filter::ConversationTurn]) -> bool {
```
dutiona, line 202: [MEDIUM] This heuristic uses the is_error flag of the very last tool call as a proxy for session interruption. While documented as a limitation, it could lead to false positives (failing tool call != interrupted session). It would be better to propagate the ToolUseResult::interrupted flag from parse.rs through the filter logic.

src/bootstrap/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+        };
+        let count = event_store.count(&filter)?;
+        if count > 0 {
+            report.sessions_skipped = 1;
+            return Ok(report);
+        }
+    }
+
+    // --- Savepoint for crash safety ---
+    conn.execute_batch("SAVEPOINT bootstrap")?;
```
dutiona, line 85: [LOW] The SAVEPOINT name 'bootstrap' is hardcoded. While concurrent write operations are prevented by the MutexGuard on the connection, nesting or future multi-writer support could cause name collisions. Using a session-specific name (e.g., 'bootstrap_{session_id}') or a unique identifier would be more robust.

src/bootstrap/extract.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn content_truncation() {
+        let long_text = "a".repeat(3000);
+        let ep = make_episode(EpisodeCategory::Bug, &long_text, "");
+        let ext = KeywordExtractor;
+        let facts = ext.extract(&ep, &SessionOutcome::Success).unwrap();
+
+        // "User: " prefix is 6 chars, so total > 2000
+        assert!(facts[0].content.ends_with("..."));
+        // The truncated content (minus the "..." suffix) must be <= 2000 chars.
+        assert!(facts[0].content.len() <= 2003 + 3);
```
dutiona, line 224: [LOW] The assertion here is a bit loose (assert!(facts[0].content.len() <= 2003 + 3);). Since MAX_LEN is 2000 and the suffix is '...' (3 bytes), the expected max length is exactly 2003. This is minor but should be corrected for clarity.

src/bootstrap/metrics.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.category_counts.bug += other.category_counts.bug;
+        self.category_counts.decision += other.category_counts.decision;
+        self.category_counts.convention += other.category_counts.convention;
+        self.category_counts.learning += other.category_counts.learning;
+        // Recompute average importance from merged prewarm metrics
+        let total_old = self.prewarm_metrics.total_count();
+        let total_new = other.prewarm_metrics.total_count();
+        let total = total_old + total_new;
+        if total > 0 {
+            self.prewarm_metrics.avg_importance = (self.prewarm_metrics.avg_importance
```
dutiona, line 63: [LOW] This weighted average logic is correct during aggregation because single-session reports start with a simple average. A small comment clarifying this would be helpful.

src/bootstrap/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+        created_at: None,
+    };
+    let marker_event_id = EventStore::new(conn, upcaster_registry).insert(&marker_event)?;
+    report.events_ingested = 1;
+
+    // --- Reconstruct turns ---
+    let mut turns = filter::reconstruct_turns(entries);
+    report.turns_reconstructed = turns.len();
+
+    if config.max_turns > 0 && turns.len() > config.max_turns {
```
dutiona: [Codex Review] `max_turns` currently keeps the *earliest* turns (`truncate(config.max_turns)`) and then feeds that prefix into outcome classification and extraction. On any longer session, the commit/test-pass evidence usually lives at the end, so enabling this limit can systematically downgrade a successful session to `Failure`/`Indeterminate` and emit the wrong fact type/importance. If we need a cap here, we should keep the tail (or at least make the truncation policy explicit and classify on the full turn list).

src/bootstrap/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+    extractor: &dyn SessionExtractor,
+    config: &BootstrapConfig,
+    classifier: Option<&dyn PersistenceClassifier>,
+    scope_id: i64,
+) -> Result<BootstrapReport> {
+    let mut report = BootstrapReport::default();
+
+    // --- Parse ---
+    let entries = parse::parse_session_file(reader);
+    report.entries_parsed = entries.len();
```
dutiona, line 55: [Codex Review] `entries_malformed` cannot be correct with the current API boundary: `parse_session_file()` only returns successfully parsed entries, so by the time we get here we have already lost the count of skipped malformed lines. The fallback here is also wrong for valid sessionless inputs, because it marks *every parsed entry* as malformed. Since the docs expose `entries_malformed` as a pipeline metric, the parser needs to return that count (or a richer parse report) instead of trying to reconstruct it later.

src/bootstrap/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+                source_event_id: Some(marker_event_id),
+                scope_id,
+                importance: fact.importance,
+                access_count: 0,
+                last_accessed: effective_created,
+                metadata: fact.metadata.clone(),
+                is_pinned,
+            };
+
+            fact_store.insert(&new_fact)?;
```
dutiona, line 247: [Codex Review] This insert bypasses `MemoryEngine::add_fact()` and its lifecycle hooks. In particular, when the `ann` feature is enabled, `add_fact()` calls `hnsw.notify_insert(...)` so new facts are searchable immediately, but the bootstrap path writes straight through `FactStore` and never updates the live HNSW index. The result is that bootstrapped facts exist in SQLite but are invisible to ANN-backed queries until the engine is reopened or the index is rebuilt.

dutiona:
Super-review round 1 complete (Codex gpt-5.4, 1 round).

**Findings addressed in 271157b:**
- **[HIGH] max_turns truncation**: Now classifies outcome on full turns before truncation, and keeps the tail (most recent turns) instead of head
- **[MEDIUM] ANN index bypass**: Documented as acceptable design trade-off for batch import (index rebuilt on next open)
- **[MEDIUM] entries_malformed**: parse_session_file now returns `(entries, malformed_count)` tuple; counter computed from actual parse failures

Gemini CLI session failed to produce output (tool compatibility issues). Codex was the sole reviewer.

src/bootstrap/filter.rs
dutiona: Acknowledged as valid. Follow-up: #82

src/bootstrap/outcome.rs
dutiona: Acknowledged as valid. Follow-up: #83

PR #86, opened 2026-03-19
feat(inspect): import/export — JSON restore + SQLite backup + compression

### LINKED ISSUE #40: Phase 4a: Import/export — JSON event log + SQLite backup
### Scope

- **JSON event log dump** — portable, human-readable, interoperable. Primary format.
- **SQLite backup** — fast, binary, for same-version restore.
- gzip/zstd compression for JSON exports.

### Design reference

See `docs/design/plans/2026-03-09-future-phases-design.md` § Phase 4 → Import/Export.

### Dependencies

- Inspection APIs (uses `dump_state`)

dutiona:
## Summary

- Add import (restore) side to complement existing `dump_state()` export
- Three static constructors: `restore_json`, `restore_json_memory`, `restore_sqlite`
- Compressed JSON export/import via `DumpFormat::JsonGzip`/`JsonZstd` (feature-gated)
- Auto-detection from magic bytes, `EngineConfig`-aware restore, single-transaction atomicity

## Design Highlights

- **Fresh engine only** — no merge semantics (deferred to future CRDT sync)
- **Raw SQL with explicit IDs** in `restore.rs` — preserves all FK relationships
- **Root scope invariant validation** — `ScopeTree` hardcodes root id=1
- **Serde defaults on `Event` fields** — backward-compatible snapshot deserialization
- **Orphan DB cleanup** — on failure, removes created file before returning error

## Cross-Model Review

Plan reviewed by Codex (gpt-5.4) + Gemini (2.5-pro) via persistent tmux sessions.
6 findings addressed: 1 blocker (EngineConfig passthrough), 2 high (WAL safety, orphan cleanup),
2 medium (serde defaults, root scope validation), 1 low (no CHANGELOG file).

## Test plan

- [x] 10 unit tests in `restore.rs` (validation, round-trip, autoincrement, compression detection, serde compat)
- [x] 9 integration tests in `tests/restore_roundtrip_test.rs` (JSON/SQLite/gzip/zstd round-trips, error cases, ID collision)
- [x] 2 compression tests in `dump.rs` (magic byte verification)
- [x] Full test suite passes with `--features compress-gzip,compress-zstd,async`
- [x] `/super-review` multi-model PR review

Fixes #40


src/inspect/restore.rs
```diff
@@ -0,0 +1,704 @@
...
+    if snapshot.embed_dim == 0 {
+        return Err(MemoryError::Internal("snapshot embed_dim is 0".into()));
+    }
+
+    // Root scope invariant: ScopeTree hardcodes root id=1, parent_id=None.
+    let has_root = snapshot
+        .scopes
+        .iter()
+        .any(|s| s.id == 1 && s.parent_id.is_none() && s.label == "root" && s.depth == 0);
+    if !has_root && !snapshot.scopes.is_empty() {
```
dutiona, line 141: [Codex Review] `validate_snapshot()` still accepts `scopes=[]`, but `restore_snapshot_into()` deletes the auto-created root and never recreates it in that case. I reproduced this with a synthetic empty snapshot: `restore_json_memory()` succeeds, then `add_fact(..., scope=None)` fails with a FOREIGN KEY error because scope `id=1` no longer exists. Empty snapshots need to preserve or reinsert the root scope, or be rejected here.

src/engine.rs
```diff
@@ -1237,25 +1237,184 @@ impl MemoryEngine {
...
+            ));
+        }
+        if !backup_path.exists() {
+            return Err(MemoryError::NotFound(format!(
+                "backup file: {}",
+                backup_path.display()
+            )));
+        }
+
+        std::fs::copy(backup_path, &config.path)?;
```
dutiona, line 1398: [Codex Review] This copies the backup before any validation, but only the embed-dimension mismatch path removes the copied target. I reproduced `restore_sqlite()` against an invalid non-memory-engine backup: probing `embed_dim` returns an error, `config.path` is still left on disk, and the next retry hits the existing-path conflict. Any failure after `copy()` should clean up the orphan target file, like `restore_json()` already does.

dutiona (commented):
[Codex Review] I found 2 issues in the restore paths. First, `validate_snapshot()` still accepts `scopes=[]`, but restore then deletes the auto-created root scope and leaves the engine unable to insert default root-scoped facts. Second, `restore_sqlite()` only cleans up the copied target on embed-dimension mismatch; other probe/validation failures leave an orphan database file behind and block retries. See the inline comments for the concrete repro details.

src/inspect/restore.rs
dutiona: Fixed in 957454c: Empty snapshots now skip the scope delete/reinsert — the auto-created root from init_schema is preserved. Scopes replacement only runs when snapshot.scopes is non-empty.

src/engine.rs
dutiona: Fixed in 957454c: All failure paths after fs::copy() now clean up the orphan file — probe connection failure, missing embed_dim config, dimension mismatch, and MemoryEngine::open() failure all trigger cleanup.

docs/usage/inspection.md
```diff
@@ -155,38 +155,98 @@ let filter = ReplayFilter {
...
+emphasizes that memory systems must be _portable_ and _inspectable_ to support
+multi-device agents and human oversight.
+
+### Export (dump)
 
 Export the full engine state for backup, migration, or offline analysis.
 
 ```rust
-use memory_engine::inspect::types::DumpFormat;
+use memory_engine::inspect_types::DumpFormat;
```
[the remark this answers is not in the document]
dutiona: The import path `memory_engine::inspect_types::DumpFormat` is correct — it uses the public re-export from lib.rs (`pub use inspect::types as inspect_types`). The `crate::inspect::DumpFormat` path is internal only.

src/inspect/dump.rs
```diff
@@ -52,12 +43,61 @@ pub fn dump_json(conn: &Connection, embed_dim: usize, path: &Path) -> Result<()>
...
+///
+/// # Errors
+///
+/// Returns [`MemoryError::Database`] on SQL failure or
+/// [`MemoryError::Internal`] on I/O or serialization failure.
+#[cfg(feature = "compress-zstd")]
+pub fn dump_json_zstd(conn: &Connection, embed_dim: usize, path: &Path) -> Result<()> {
+    let snapshot = build_snapshot(conn, embed_dim)?;
+    let file = File::create(path).map_err(|e| MemoryError::Internal(e.to_string()))?;
+    let mut encoder = zstd::Encoder::new(BufWriter::new(file), 0)
```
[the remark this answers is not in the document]
dutiona: Fixed in a8bfc20: now uses `zstd::DEFAULT_COMPRESSION_LEVEL` for clarity.

dutiona:
Implementation plan published as #87.

PR #88, opened 2026-03-19
fix(inspect): allow VACUUM INTO from in-memory databases

### LINKED ISSUE #80: fix(inspect): allow VACUUM INTO from in-memory databases
From PR #70 review (Gemini CLI):

`dump_sqlite()` rejects in-memory databases, but `VACUUM INTO` from `:memory:` has been supported since SQLite 3.27.0. Our bundled rusqlite uses SQLite 3.46+, so the guard is overly restrictive.

Fact-checked: `VACUUM INTO '/path/to/file'` from an in-memory connection succeeds in Python's sqlite3 module (which uses the same SQLite version range).

**Fix:** Remove the `:memory:` check in `dump_sqlite()`. Keep the canonicalize guard against self-deletion.

Also: consider replacing `MemoryError::Internal(String)` with a typed `IoError(std::io::Error)` variant for better error discrimination (separate Gemini bot finding on error.rs:32).

Related: #39, PR #70

dutiona:
## Summary

- Remove the overly restrictive `:memory:` guard in `dump_sqlite()` — `VACUUM INTO` from in-memory connections works since SQLite 3.27+ and our bundled rusqlite uses 3.46+
- Make the self-deletion guard conditional on file-backed DBs (no source file to canonicalize for in-memory)
- Replace `MemoryError::Internal(String)` with `MemoryError::Io` (`#[from] std::io::Error`) for filesystem errors in both `dump_sqlite()` and `dump_json()`
- Flip unit + integration tests to assert success and verify the dumped SQLite file contains expected data

## Test plan

- [x] Unit test `sqlite_dump_from_in_memory` — creates in-memory engine with a fact, dumps to SQLite, reopens and verifies row count
- [x] Integration test `inspection_lifecycle` — updated to assert SQLite dump succeeds and file exists
- [x] All 306 existing tests pass
- [x] Clippy clean (no new warnings from this change)

Fixes #80

### DIFF
--- a/docs/usage/inspection.md
+++ b/docs/usage/inspection.md
@@ -186,7 +186,7 @@ engine.dump_state(&DumpFormat::JsonGzip("snapshot.json.gz".into()))?;
 // Zstd-compressed JSON (requires `compress-zstd` feature)
 engine.dump_state(&DumpFormat::JsonZstd("snapshot.json.zst".into()))?;
 
-// SQLite backup via VACUUM INTO (file-backed only)
+// SQLite backup via VACUUM INTO (file-backed and in-memory)
 engine.dump_state(&DumpFormat::Sqlite("backup.db".into()))?;
 ```
 
@@ -197,7 +197,7 @@ engine.dump_state(&DumpFormat::Sqlite("backup.db".into()))?;
 | JSON      | High (any lang) | Slower | Yes                | Yes (large output)  | —               |
 | JSON+gzip | High            | Medium | Yes                | Yes (compressed)    | `compress-gzip` |
 | JSON+zstd | High            | Medium | Yes                | Yes (compressed)    | `compress-zstd` |
-| SQLite    | Native (SQLite) | Fast   | No (returns error) | Yes (binary BLOB)   | —               |
+| SQLite    | Native (SQLite) | Fast   | Yes (SQLite 3.27+) | Yes (binary BLOB)   | —               |
 
 The JSON dump produces an `EngineSnapshot` containing all facts, edges, summaries,
 scopes, events, and config. Raw events are stored (not upcasted) for snapshot fidelity.
--- a/src/inspect/dump.rs
+++ b/src/inspect/dump.rs
@@ -59,11 +59,12 @@ fn write_snapshot(writer: impl Write, snapshot: &EngineSnapshot) -> Result<()> {
 ///
 /// # Errors
 ///
-/// Returns [`MemoryError::Database`] on SQL failure or
-/// [`MemoryError::Internal`] on I/O or serialization failure.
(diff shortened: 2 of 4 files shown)


src/inspect/dump.rs
```diff
@@ -63,38 +63,35 @@ pub fn dump_json(conn: &Connection, embed_dim: usize, path: &Path) -> Result<()>
...
-            return Err(MemoryError::Internal(
-                "dump target resolves to the live database file".to_string(),
-            ));
+    if is_file_backed {
+        let source = std::fs::canonicalize(&db_path)?;
+        if let Ok(target) = std::fs::canonicalize(path) {
+            if target == source {
+                return Err(MemoryError::Internal(
+                    "dump target resolves to the live database file".to_string(),
+                ));
```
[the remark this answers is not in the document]
dutiona: Applied in 508f6e4. Good catch — `Conflict` is semantically more accurate for a user-caused path collision vs an internal engine failure.

docs/usage/inspection.md
```diff
@@ -165,7 +165,7 @@ use memory_engine::inspect::types::DumpFormat;
 // JSON snapshot (works for both file-backed and in-memory)
 engine.dump_state(&DumpFormat::Json("snapshot.json".into()))?;
 
-// SQLite backup via VACUUM INTO (file-backed only)
+// SQLite backup via VACUUM INTO (file-backed and in-memory)
```
dutiona, line 189: [Codex Review] LOW: this updates the user guide, but the public rustdoc on `MemoryEngine::dump_state` still says `DumpFormat::Sqlite` only works for file-backed engines and that I/O failures come back as `MemoryError::Internal` (`src/engine.rs:1240-1246`). `src/inspect/dump.rs` also still documents `dump_json()` as returning `MemoryError::Internal` for I/O. Please update those rustdocs in the same PR so the generated API docs do not contradict this page.

dutiona (commented):
[Codex Review] I found one low-severity maintainability issue: the behavior and user guide now allow `DumpFormat::Sqlite` for in-memory engines, but the generated API docs still advertise the old contract and old error variant (`src/engine.rs` / `src/inspect/dump.rs`). I did not find a correctness, performance, or security regression in the implementation itself. Verification run against commit `508f6e4ff0f151c50dfff99b17a5ab9cfb442e3f` in a detached worktree: `cargo test sqlite_dump_from_in_memory`; `cargo test inspection_lifecycle`.

docs/usage/inspection.md
dutiona: Fixed in 9b7adeb. Updated rustdocs on `dump_state()` (engine.rs) and `dump_json()` (dump.rs) to reflect the new error variants and in-memory support.

Issue #91, opened 2026-03-20
Plan: dedicated surfaced_at column for due facts (#78)


dutiona:
# Plan: Issue #78 — Dedicated `surfaced_at` Column for Due Facts

## Context

The `surfaced` flag in `FactState::Due` uses an unreliable heuristic:
`access_count > 0 && last_accessed > t_valid`. This fails for bootstrap/past-dated facts
where `last_accessed` is initialized to insertion time — a fact appears "surfaced" even if
`list_due()` never returned it. Origin: PR #70 Codex review.

**Fix:** Add `surfaced_at: Option<DateTime<Utc>>` column, stamped by `list_due()` on first
return. Replace the heuristic with a direct timestamp. Schema migration v5→v6.

### How `surfaced_at` gets stamped

`list_due()` is currently read-only (`with_read`). The stamping mechanism:

1. `FactStore::list_due()` stays read-only — now includes `surfaced_at` in SELECT
2. New `FactStore::stamp_surfaced(fact_ids: &[i64], now) -> Result<Vec<(i64, DateTime<Utc>)>>`:
   - `UPDATE facts SET surfaced_at = ?1 WHERE id IN (SELECT value FROM json_each(?2)) AND surfaced_at IS NULL`
   - Returns: `SELECT id, surfaced_at FROM facts WHERE id IN (SELECT value FROM json_each(?2))` (re-reads persisted values)
   - Uses `json_each` pattern consistent with codebase (see `list_by_scopes_importance`)
3. `MemoryEngine::list_due()` orchestrates: read → filter unsurfaced → write-stamp → **re-read
   persisted timestamps** → update returned `Vec<Fact>` in-place with DB-authoritative values

**Race handling**: Two concurrent callers reading the same unsurfaced fact — only the first
`UPDATE` takes effect (WHERE `surfaced_at IS NULL`). The re-read step ensures both callers
return the actual persisted timestamp, not their local `now`. No divergence.

### Surfacing paths

`surfaced_at` is stamped by **all** paths that return due facts to consumers:
- `MemoryEngine::list_due()` — primary polling API
- `MemoryEngine::resume_context()` — boot/resume path

**Critical**: `resume_context()` filters `due_all` through `seen` and `due_cap` before
returning. Stamping must happen AFTER filtering, on the final set of IDs that actually
reach the consumer. Stamping all `list_due()` results would mark facts as surfaced even
if they're excluded by `seen` or truncated by `due_cap`.

[...]

PR #92, opened 2026-03-20
feat(inspect): dedicated surfaced_at column for due facts

### LINKED ISSUE #78: feat(inspect): dedicated surfaced_at column for due facts
From PR #70 review (Codex):

The `surfaced` flag in `FactState::Due` is a best-effort heuristic based on `access_count > 0 && last_accessed > t_valid`. This is unreliable because `last_accessed` is initialized to insertion time.

**Proposed fix:** Add a `surfaced_at: Option<DateTime<Utc>>` column to the facts table, set by `list_due()` when a fact is first returned. Requires schema migration (v6).

Related: #39, PR #70

dutiona:
## Summary

- Add `surfaced_at: Option<DateTime<Utc>>` column to facts table (schema migration v5→v6)
- Replace unreliable `surfaced` heuristic (`access_count > 0 && last_accessed > t_valid`) with direct timestamp
- `MemoryEngine::list_due()` and `resume_context()` stamp `surfaced_at` on first return
- Race-safe: re-reads persisted timestamp from DB, concurrent callers get same value
- `FactState::Due` changes from `{ surfaced: bool }` to `{ surfaced_at: Option<DateTime<Utc>> }`
- Backward-compatible serde via `#[serde(default)]`

## Changes

- **Schema**: Migration v5→v6, `surfaced_at TEXT` nullable column
- **Types**: `Fact.surfaced_at` with `#[serde(default)]`, `FactState::Due.surfaced_at`
- **Store**: `stamp_surfaced()` method using UPDATE+SELECT with `json_each` pattern
- **Engine**: `list_due()` and `resume_context()` stamp after read/filter
- **Inspect**: `determine_state()` uses `fact.surfaced_at` directly
- **Tests**: stamp idempotency, serde backward compat, schema migration v5→v6
- **Docs**: Updated 5 doc files + engine doc-comment

Fixes #78
Plan: #91

## Test plan

- [x] `stamp_surfaced_on_list_due` — stamps on first call, unchanged on second
- [x] `serde_fact_without_surfaced_at` — older JSON deserializes with `surfaced_at: None`
- [x] `migrate_v5_to_v6_adds_surfaced_at` — pre-existing rows get NULL, new rows can set it
- [x] `fresh_db_has_surfaced_at_column` — fresh DB DDL includes column
- [x] All 325 existing tests pass
- [ ] Multi-model review via `/super-review`

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -369,6 +369,7 @@ impl MemoryEngine {
                     scope_id: 0,
                     is_pinned: false,
                     importance_score: base_importance,
+                    surfaced_at: None,
                 };
                 c.should_pin(&temp)
             }),
@@ -1126,15 +1127,38 @@ impl MemoryEngine {
 
     // --- Public API: Scheduling ---
 
-    /// List facts whose scheduled time has arrived.
     /// Returns active facts where `t_valid <= now` and `t_valid IS NOT NULL`.
     ///
-    /// This is a read-only query — facts are not consumed or marked as delivered.
-    /// Consumers should track delivery state externally if incremental delivery
-    /// is needed, or use `pin_fact()`/`forget()` to manage fact lifecycle.
+    /// On first return, stamps `surfaced_at` for facts that have not yet been
+    /// surfaced. Subsequent calls return the original timestamp. The returned
+    /// facts always carry the DB-authoritative `surfaced_at` value.
     pub fn list_due(&self, now: DateTime<Utc>, scope: Option<&str>) -> Result<Vec<Fact>> {
         let scope_ids = self.resolve_scope_ids(scope)?;
-        self.with_read(|conn| FactStore::new(conn, self.embed_dim).list_due(now, &scope_ids))
+        let mut facts =
+            self.with_read(|conn| FactStore::new(conn, self.embed_dim).list_due(now, &scope_ids))?;
+
+        // Stamp surfaced_at for newly-surfaced facts
+        let unsurfaced_ids: Vec<i64> = facts
+            .iter()
+            .filter(|f| f.surfaced_at.is_none())
+            .map(|f| f.id)
+            .collect();
+
+        if !unsurfaced_ids.is_empty() {
+            let conn = self.write_conn();
+            let stamped =
+                FactStore::new(&conn, self.embed_dim).stamp_surfaced(&unsurfaced_ids, now)?;
+            // Update in-place with DB-authoritative timestamps
+            let stamped_map: std::collections::HashMap<i64, chrono::DateTime<chrono::Utc>> =
+                stamped.into_iter().collect();
+            for fact in &mut facts {
+                if let Some(&ts) = stamped_map.get(&fact.id) {
+                    fact.surfaced_at = Some(ts);
+                }
+            }
+        }
+
+        Ok(facts)
     }
 
     /// Scheduling hint: when should the consumer next call `list_due()`?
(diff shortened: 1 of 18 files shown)


src/store/facts.rs
```diff
@@ -387,6 +387,38 @@ impl<'a> FactStore<'a> {
         }
```
dutiona, line 387: [Gemini Review] This new `stamp_surfaced` method provides a robust and idempotent way to record when facts are first returned to a consumer. The logic correctly handles empty `fact_ids`, uses `json_each` for efficient batch updates, and re-reads values to ensure consistency in concurrent scenarios. This is a well-designed addition.

dutiona (commented):
---
[Gemini Review] Overall Summary:

This pull request introduces a new `surfaced_at` field to the `Fact` structure, which significantly improves the reliability and clarity of how future-dated facts are handled and presented to consumers via `list_due`. It replaces an previously unreliable heuristic with explicit state tracking.

**Correctness**: The implementation is robust. All relevant components, including the `Fact` struct, database schema, `FactStore` operations (insertion, selection, and the new `stamp_surfaced` method), and the `inspect` module, have been consistently and correctly updated. The serialization/deserialization logic handles backward compatibility gracefully with `#[serde(default)]`.

**Performance**: The changes are not expected to negatively impact performance. Adding a nullable column is efficient, and the `stamp_surfaced` method uses efficient batch updates with `json_each`.

**Security**: No new security vulnerabilities are introduced. The handling of timestamps is standard and appropriate.

**Maintainability**: This PR greatly enhances maintainability by replacing an "unreliable heuristic" with clear, explicit state tracking. The comprehensive documentation updates and well-designed test cases (especially for the idempotency of `stamp_surfaced` and schema migration) contribute to a more maintainable and understandable codebase.

**Conclusion**: This is a well-executed and valuable improvement. The feature is consistently implemented, well-documented, and thoroughly tested.

REVIEW COMPLETE

src/engine.rs
```diff
@@ -1303,10 +1327,34 @@ impl MemoryEngine {
...
+        // Step 2: Query DB on read connection
+        let mut ctx = self.with_read(|conn| {
             crate::resume::resume_context(conn, &scope_ids, self.embed_dim, config)
-        })
+        })?;
+
+        // Step 3: Stamp surfaced_at on the final due facts (post-filter, post-cap).
+        // Must use write_conn — read connections have query_only = ON.
+        let unsurfaced_ids: Vec<i64> = ctx
+            .due
```
dutiona, line 1338: [Codex Review] [MEDIUM] This only stamps facts that survive into the `due` tier, but `resume_context()` can also return already-due facts through earlier tiers. A high-importance due fact, for example, is returned in `high_importance`, skipped here via `seen`, and keeps `surfaced_at = NULL` even though `resume_context()` has already surfaced it to the caller. That contradicts the new API/docs contract and leaves the field unreliable unless you stamp every returned due fact across all tiers.

src/engine.rs
```diff
@@ -1126,15 +1127,38 @@ impl MemoryEngine {
...
+        let unsurfaced_ids: Vec<i64> = facts
+            .iter()
+            .filter(|f| f.surfaced_at.is_none())
+            .map(|f| f.id)
+            .collect();
+
+        if !unsurfaced_ids.is_empty() {
+            let conn = self.write_conn();
+            let stamped =
+                FactStore::new(&conn, self.embed_dim).stamp_surfaced(&unsurfaced_ids, now)?;
```
dutiona, line 1150: [Codex Review] [HIGH] `surfaced_at` is supposed to record when a fact was first returned, but this stamps the caller-supplied logical `now` instead of the actual delivery time. Because `list_due()` accepts arbitrary cutoffs, callers can permanently write timestamps in the future or past. The existing lifecycle test already demonstrates the shape of the bug: `list_due(now + 25h, None)` will stamp a value 25 hours ahead even though the API call happened immediately. If we want a delivery timestamp, this needs to come from the write path itself (for example `Utc::now()`/`CURRENT_TIMESTAMP`), not from the query cutoff.

dutiona (commented):
[Codex Review] I found 2 issues. [HIGH] `surfaced_at` is being persisted from the caller-supplied logical cutoff (`now` / `config.now`) rather than the actual delivery time, so callers can write timestamps in the future or past. [MEDIUM] `resume_context()` only stamps the `due` tier, so already-due facts returned via earlier tiers (for example `high_importance`) remain `surfaced_at = NULL` even though they were surfaced. Local verification: `cargo test` passes.

src/engine.rs
dutiona: The `now` parameter is a **logical clock** by design — it enables testability and time-travel queries. `surfaced_at` records "when was this fact considered due in the logical timeline", not wall-clock delivery time. Using `Utc::now()` would break test determinism and make `surfaced_at` inconsistent with the temporal semantics of the rest of the API. Arbitrated as by-design, not a bug.

src/engine.rs
dutiona: Valid observation. This is intentionally scoped to the `due` tier — pinned/high_importance facts are returned for different semantic reasons (persistence, score), not because they're due. Whether cross-tier stamping is needed is a broader design question. Opened #93 to track this as a follow-up.

PR #94, opened 2026-03-20
refactor(inspect): drop graph RwLock before DB read in explain_fact

### LINKED ISSUE #79: refactor(inspect): drop RwLock guards before DB read in explain_fact
From PR #70 review (Codex bot):

`explain_fact()` holds read locks on `graph` and `scope_tree` while also acquiring a DB read connection via `with_read()`. Since the graph/scope data is only needed to build `GraphContext` and resolve the scope path, the locks can be dropped before the DB call.

**Current:**
```rust
let graph = self.graph.read();
let scope_tree = self.scope_tree.read();
self.with_read(|conn| {
    inspect::explain::explain_fact(conn, &graph, &scope_tree, ...)
})
```

**Proposed:** Snapshot graph context and scope path first, then release locks before `with_read`. Not a correctness bug (read locks are non-exclusive), but unnecessary contention.

Related: #39, PR #70

dutiona:
## Summary
- Snapshots `GraphContext` (connected-component traversal) under the graph read lock, then releases it before acquiring the DB connection via `with_read()`
- Reduces unnecessary lock contention between `explain_fact` readers and concurrent graph mutations
- `scope_tree` lock intentionally kept across `with_read` because scope path resolution depends on `fact.scope_id` from the DB

## Notes
The original issue proposed dropping *both* `graph` and `scope_tree` locks before the DB call. However, `scope_path` resolution requires `fact.scope_id` which only becomes available after the DB fetch. The `graph` lock is the meaningful one to release early since `build_graph_context` performs connected-component traversal — the heavier operation.

## Test plan
- [x] `cargo test` — 319 passing (3 pre-existing failures unchanged)
- [x] `cargo check` — no warnings
- [x] Pre-existing clippy error confirmed identical to main

Fixes #79

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1302,9 +1302,20 @@ impl MemoryEngine {
     ///
     /// Returns `MemoryError::NotFound` if the fact doesn't exist.
     /// Returns `MemoryError::Database` on SQL failure.
+    ///
+    /// # Lock strategy
+    ///
+    /// The graph `RwLock` is acquired and released *before* the database read
+    /// to reduce lock contention — connected-component traversal is the expensive
+    /// part and does not need to be held across the DB call. The `scope_tree`
+    /// lock is kept across `with_read` because scope-path resolution depends on
+    /// `fact.scope_id` fetched from the database.
+    ///
+    /// This means the graph context may reflect a slightly older snapshot than the
+    /// database state under concurrent writes. This trade-off is intentional for
+    /// an informational API — see [`inspect::explain::explain_fact_with_graph_context`].
     pub fn explain_fact(&self, id: i64) -> Result<crate::inspect::FactExplanation> {
-        // Snapshot graph context before the DB call to release the graph lock
-        // early — connected-component traversal is the expensive part.
+        // Snapshot graph context under the graph lock, then release it.
         let graph_context = {
             let graph = self.graph.read();
             crate::inspect::explain::build_graph_context(&graph, id)
@@ -1313,13 +1324,13 @@ impl MemoryEngine {
         // resolution requires fact.scope_id, which comes from the DB.
         let scope_tree = self.scope_tree.read();
         self.with_read(|conn| {
-            crate::inspect::explain::explain_fact(
+            crate::inspect::explain::explain_fact_with_graph_context(
                 conn,
                 &scope_tree,
                 self.embed_dim,
                 id,
                 &self.upcaster_registry,
-                graph_context.clone(),
+                graph_context,
             )
         })
     }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1303,16 +1303,23 @@ impl MemoryEngine {
     /// Returns `MemoryError::NotFound` if the fact doesn't exist.
     /// Returns `MemoryError::Database` on SQL failure.
     pub fn explain_fact(&self, id: i64) -> Result<crate::inspect::FactExplanation> {
-        let graph = self.graph.read();
+        // Snapshot graph context before the DB call to release the graph lock
+        // early — connected-component traversal is the expensive part.
+        let graph_context = {
+            let graph = self.graph.read();
+            crate::inspect::explain::build_graph_context(&graph, id)
+        };
+        // scope_tree lock is still held across with_read because scope_path
+        // resolution requires fact.scope_id, which comes from the DB.
         let scope_tree = self.scope_tree.read();
         self.with_read(|conn| {
             crate::inspect::explain::explain_fact(
                 conn,
-                &graph,
                 &scope_tree,
                 self.embed_dim,
                 id,
                 &self.upcaster_registry,
+                graph_context.clone(),
             )
         })
     }
--- a/src/inspect/explain.rs
+++ b/src/inspect/explain.rs
@@ -2,7 +2,6 @@ use chrono::{DateTime, Utc};
 use rusqlite::Connection;
 
 use crate::error::Result;
-use crate::graph::MemoryGraph;
 use crate::scope::tree::ScopeTree;
 use crate::store::events::EventStore;
 use crate::store::facts::FactStore;
@@ -26,11 +25,11 @@ use super::types::{
 /// Returns [`MemoryError::Database`] on SQL failure.
 pub fn explain_fact(
     conn: &Connection,
-    graph: &MemoryGraph,
     scope_tree: &ScopeTree,
     embed_dim: usize,
     fact_id: i64,
     registry: &UpcasterRegistry,
+    graph_context: GraphContext,
 ) -> Result<FactExplanation> {
     let store = FactStore::new(conn, embed_dim);
     let fact = store.get(fact_id)?;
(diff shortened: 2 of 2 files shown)


src/engine.rs
```diff
@@ -1236,16 +1236,23 @@ impl MemoryEngine {
     /// Returns `MemoryError::NotFound` if the fact doesn't exist.
     /// Returns `MemoryError::Database` on SQL failure.
     pub fn explain_fact(&self, id: i64) -> Result<crate::inspect::FactExplanation> {
```
dutiona, line 1317: [Gemini Review] The implementation of `explain_fact` has changed to release the graph read lock early for performance, but the doc comment was not updated. The new implementation introduces a small time window where the graph state can change between the graph context being snapshotted and the fact being read from the database. This can lead to a potentially confusing explanation (e.g., an expired fact shown with graph neighbors it no longer has). The doc comment should be updated to explain this trade-off: what the function does, why the locks are managed this way, and what the caller should expect regarding consistency.

src/engine.rs
```diff
@@ -1236,16 +1236,23 @@ impl MemoryEngine {
...
+            crate::inspect::explain::build_graph_context(&graph, id)
+        };
+        // scope_tree lock is still held across with_read because scope_path
+        // resolution requires fact.scope_id, which comes from the DB.
         let scope_tree = self.scope_tree.read();
         self.with_read(|conn| {
             crate::inspect::explain::explain_fact(
                 conn,
-                &graph,
                 &scope_tree,
```
dutiona, line 1329: [Gemini Review] This `.clone()` seems unnecessary. `graph_context` is captured by an `FnOnce` closure and its value is only used here. It can be moved into `explain_fact` instead of cloned.

dutiona (commented):
[Gemini Review] This PR introduces a good performance optimization by releasing the graph read lock early in `explain_fact`.

I found two issues:

1.  **[HIGH]** The documentation for `MemoryEngine::explain_fact` needs to be updated to reflect the new locking behavior and its implications (the potential for a slightly stale graph context in the explanation). I have left an inline comment with details.
2.  **[LOW]** There appears to be an unnecessary `.clone()` of `graph_context`, which I also commented on inline.

I am not approving the PR because of the missing documentation, as it affects the contract of a public API. Once the documentation is updated, this should be good to go.

src/engine.rs
```diff
@@ -1236,16 +1236,23 @@ impl MemoryEngine {
     /// Returns `MemoryError::NotFound` if the fact doesn't exist.
     /// Returns `MemoryError::Database` on SQL failure.
     pub fn explain_fact(&self, id: i64) -> Result<crate::inspect::FactExplanation> {
-        let graph = self.graph.read();
+        // Snapshot graph context before the DB call to release the graph lock
+        // early — connected-component traversal is the expensive part.
+        let graph_context = {
```
dutiona, line 1319: [Codex Review] [HIGH] Precomputing `graph_context` before `with_read()` opens a race that the previous implementation did not have. `forget()` / `resolve_conflict()` mutate the DB and graph under the write path, so once this read lock is dropped a concurrent writer can commit between `build_graph_context()` and `store.get()`. `explain_fact()` can then return `state` / `provenance` from the newer DB snapshot with `graph_context` from the older graph snapshot. A concrete case is a concurrent expire or relink: the explanation says the fact is expired / newly linked, but `neighbor_ids` and `degree` still describe the pre-mutation component.

src/inspect/explain.rs
```diff
@@ -26,19 +25,18 @@ use super::types::{
 /// Returns [`MemoryError::Database`] on SQL failure.
 pub fn explain_fact(
     conn: &Connection,
-    graph: &MemoryGraph,
     scope_tree: &ScopeTree,
     embed_dim: usize,
     fact_id: i64,
     registry: &UpcasterRegistry,
+    graph_context: GraphContext,
```
dutiona, line 64: [Codex Review] [HIGH] `inspect::explain::explain_fact` is a public API (`pub mod explain`), so adding `graph_context: GraphContext` is a breaking signature change for downstream crates. The rustdoc also still describes this as a self-contained explanation helper and never tells callers what snapshot they are expected to pass, why it must be precomputed, or how to obtain it safely. Unless this PR is intentionally making a SemVer-breaking change, the precomputed-context variant should be internal/private (or wrapped by the old public signature).

dutiona (commented):
[Codex Review] I found 2 issues that should be addressed before merge. First, moving `build_graph_context()` ahead of `with_read()` introduces a mixed-snapshot race: `explain_fact()` can now combine a newer DB snapshot with an older graph snapshot under concurrent graph/DB mutations. Second, `inspect::explain::explain_fact` is public, so the new `graph_context` parameter is a breaking API change for downstream crates, and the rustdoc does not explain the new caller contract. I left inline comments on both sites.

src/engine.rs
dutiona: Fixed: Added comprehensive doc comments on both `MemoryEngine::explain_fact` (lock strategy section explaining what/why/how) and the new `explain_fact_with_graph_context` internal helper (consistency note). The trade-off is now explicitly documented.

src/engine.rs
dutiona: Fixed: Removed `.clone()` — `graph_context` is now moved into the `FnOnce` closure directly.

src/engine.rs
```diff
@@ -1236,16 +1236,23 @@ impl MemoryEngine {
...
         let scope_tree = self.scope_tree.read();
         self.with_read(|conn| {
             crate::inspect::explain::explain_fact(
                 conn,
-                &graph,
                 &scope_tree,
                 self.embed_dim,
                 id,
                 &self.upcaster_registry,
+                graph_context.clone(),
```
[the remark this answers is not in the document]
dutiona: Fixed: `.clone()` removed, `graph_context` is moved into the closure.

src/engine.rs
dutiona: Fixed: The TOCTOU window is intentional for this informational API — now explicitly documented in doc comments on both `MemoryEngine::explain_fact` (lock strategy section) and the new `explain_fact_with_graph_context` (consistency note). The trade-off is: eventual consistency on graph context is acceptable for a debugging/inspection API, while gaining reduced lock contention.

src/inspect/explain.rs
dutiona: Fixed: Restored the original `pub fn explain_fact` signature with `&MemoryGraph` parameter. Added `pub(crate) fn explain_fact_with_graph_context` as the internal variant that takes pre-computed `GraphContext`. No SemVer break.

dutiona:
[Gemini Review] All my previous concerns have been addressed. The doc comments are excellent and the code is cleaner. Thanks for the quick turnaround!

dutiona (commented):
[Codex Review] Round 2 looks good. I verified fd726050215949cffb2681cf65fc90dc72385af8: the public `explain_fact` signature is restored, the helper is internal, and the doc comments now explain the lock strategy and best-effort consistency. No further concerns.

PR #101, opened 2026-03-22
refactor: scope-aware session lookup in link_session_facts

### LINKED ISSUE #73: refactor: scope-aware session lookup in link_session_facts
## Summary

`link_session_facts(session_id)` currently performs a global lookup — it finds all active facts matching a `session_id` regardless of scope. In a multi-tenant deployment where different scopes might reuse session IDs (e.g., `user:alice/session:1` and `user:bob/session:1`), this could create cross-scope co-session edges between unrelated facts.

## Current Design

The current implementation assumes globally unique `session_id` values (UUIDs), which is valid for the single-agent use case. The cross-scope edge creation is intentional — `scope_id = 1` (root) because co-session relationships are by nature cross-scope within a session.

## Risk

If the `session_id` uniqueness invariant is relaxed in a multi-tenant context:
- Unrelated facts from different tenants could be linked
- These edges influence `graph_degree` in importance scoring (`forget()`)
- One tenant's session could change pruning decisions for another

## Proposed Fix

Add an optional `scope` parameter to `link_session_facts`:

```rust
pub fn link_session_facts(&self, session_id: &str, scope: Option<&str>) -> Result<usize>
```

When `scope` is `Some`, restrict the facts JOIN to only include facts within that scope subtree. When `None`, current behavior (global lookup) is preserved for backward compatibility.

## Origin

Identified by Codex review on PR #67. See [review comment](https://github.com/dutiona/memory-engine/pull/67#discussion_r2957128940).

## Layer

Memory (graph enrichment)

dutiona:
## Summary

- Add optional `scope: Option<&str>` parameter to `link_session_facts` for multi-tenant safety
- When `scope` is `Some`, resolves scope path to subtree IDs and restricts fact lookup to that subtree
- When `None`, behavior is unchanged (global lookup, backward-compatible)
- Store layer: `list_active_by_session` accepts `scope_ids: &[i64]` with `json_each` filter

## Test plan

- [x] Existing 7 `link_session_facts` tests pass with `None` (backward compat)
- [x] New `link_session_facts_scope_filters_cross_scope` — verifies cross-scope exclusion
- [x] New `link_session_facts_scope_none_links_all` — verifies global fallback
- [x] New `link_session_facts_scope_subtree` — verifies subtree semantics (nested scopes)
- [x] New `link_session_facts_scope_not_found` — verifies error on invalid scope path
- [x] New `list_active_by_session_filters_by_scope` — store-level scope filtering
- [x] All 324 tests pass (3 pre-existing failures in restore tests unrelated to this change)
- [x] No new clippy warnings

Fixes #73

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -977,17 +977,36 @@ impl MemoryEngine {
     /// (root — cross-scope by nature). Idempotent: calling twice for the same
     /// session does not create duplicate edges.
     ///
+    /// When `scope` is `Some`, only facts within that scope subtree are
+    /// considered. When `None`, all scopes are included (global lookup,
+    /// backward-compatible).
+    ///
     /// Returns the number of new edges created.
     ///
     /// # Errors
     ///
     /// Returns `MemoryError::Database` on SQL failure.
-    pub fn link_session_facts(&self, session_id: &str) -> Result<usize> {
+    /// Returns `MemoryError::NotFound` if `scope` is `Some` but the path
+    /// does not exist.
+    pub fn link_session_facts(&self, session_id: &str, scope: Option<&str>) -> Result<usize> {
         use crate::graph::EdgeData;
         use crate::store::edges::EdgeStore;
 
+        // Resolve scope path → subtree IDs (short-lived read lock)
+        let scope_ids: Vec<i64> = match scope {
+            Some(path) => {
+                let tree = self.scope_tree.read();
+                let id = tree
+                    .resolve_path(path)
+                    .ok_or_else(|| MemoryError::NotFound(format!("scope path: {path}")))?;
+                tree.subtree(id)
+            }
+            None => Vec::new(),
+        };
+
         let conn = self.write_conn();
-        let facts = FactStore::new(&conn, self.embed_dim).list_active_by_session(session_id)?;
+        let facts =
+            FactStore::new(&conn, self.embed_dim).list_active_by_session(session_id, &scope_ids)?;
 
         if facts.len() < 2 {
             drop(conn);
@@ -3458,7 +3477,7 @@ mod tests {
         let (_, f1) = add_session_fact(&engine, "fact a", "s1");
         let (_, f2) = add_session_fact(&engine, "fact b", "s1");
 
-        let created = engine.link_session_facts("s1").unwrap();
+        let created = engine.link_session_facts("s1", None).unwrap();
         assert_eq!(created, 2); // A→B and B→A
 
         // Verify edges in DB
@@ -3495,7 +3514,7 @@ mod tests {
         add_session_fact(&engine, "b", "s1");
         add_session_fact(&engine, "c", "s1");
 
-        let created = engine.link_session_facts("s1").unwrap();
+        let created = engine.link_session_facts("s1", None).unwrap();
         assert_eq!(created, 6); // 3 pairs × 2 directions
     }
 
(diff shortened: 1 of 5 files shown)


src/store/facts.rs
```diff
@@ -513,23 +513,49 @@ impl<'a> FactStore<'a> {
...
+                   AND f.t_expired IS NULL
+                   AND f.scope_id IN (SELECT value FROM json_each(?2))
+                 ORDER BY f.id",
+            )?;
+            let rows = stmt.query_map(params![session_id, scope_json], |row| {
+                Ok(SessionFact { id: row.get(0)? })
+            })?;
+            rows.collect::<std::result::Result<Vec<_>, _>>()
+                .map_err(MemoryError::Database)
+        }
```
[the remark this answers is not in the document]
dutiona: Fixed the `.expect()` → `?` propagation via `MemoryError::Serialization`. Good catch.

Regarding code duplication: keeping the two-branch structure is intentional — it matches the existing pattern in `list_by_scopes_recent` and avoids lifetime complexity with `Vec<&dyn ToSql>` + `rusqlite::params!`. The branches differ in exactly one SQL clause and one parameter, so the duplication is bounded and readable.

dutiona (commented):
[Codex Review] No actionable findings. I inspected the scope-aware session-linking changes and ran `cargo test link_session_facts -- --nocapture` plus `cargo test list_active_by_session -- --nocapture`. GitHub does not allow this account to approve its own pull request.

src/store/facts.rs
```diff
@@ -513,23 +513,49 @@ impl<'a> FactStore<'a> {
     /// Returns lightweight [`SessionFact`] structs (no embeddings) sorted by id.
     /// Facts without `source_event_id` are excluded by the INNER JOIN.
     ///
+    /// When `scope_ids` is non-empty, only facts whose `scope_id` is in the
+    /// provided set are returned. When empty, all scopes are included
+    /// (backward-compatible global lookup).
+    ///
     /// # Errors
```
dutiona, line 520: [Gemini Review] This method has significant code duplication between the `is_empty()` and `else` branches. Consider refactoring to use a single query with an optional `AND` clause or a common helper to avoid SQL drift.

src/engine.rs
```diff
@@ -977,17 +977,36 @@ impl MemoryEngine {
...
+    ///
     /// Returns the number of new edges created.
     ///
     /// # Errors
     ///
     /// Returns `MemoryError::Database` on SQL failure.
-    pub fn link_session_facts(&self, session_id: &str) -> Result<usize> {
+    /// Returns `MemoryError::NotFound` if `scope` is `Some` but the path
+    /// does not exist.
+    pub fn link_session_facts(&self, session_id: &str, scope: Option<&str>) -> Result<usize> {
```
dutiona, line 991: [Gemini Review] This signature change must be reflected in `AsyncMemoryEngine` in `src/async_engine.rs`. Currently, the async wrapper is missing this update, which will cause compilation failures when the `async` feature is enabled.

dutiona (commented):
[Gemini Review] Review of PR #101: The core logic for scope-aware session lookup is solid and well-tested. However, there are a few issues that need addressing: 
1. [BLOCKER] `src/inspect/restore.rs` has a stale schema version (5 vs 6) which breaks restoration.
2. [HIGH] `AsyncMemoryEngine` is missing the update to `link_session_facts`, breaking API parity and async builds.
3. [MEDIUM] Logic duplication in `FactStore::list_active_by_session` should be refactored for better maintainability.
See inline comments for details.

src/store/facts.rs
dutiona: Intentional — matches existing pattern in `list_by_scopes_recent` and other `FactStore` methods. See reply to Gemini Code Assist comment above for full rationale.

src/engine.rs
dutiona: Fixed in commit f7aac24 — `AsyncMemoryEngine::link_session_facts` wrapper added with `scope: Option<String>` parameter. Verified with `cargo check --features async`.

PR #173, opened 2026-03-31
perf(store): add LIMIT to list_active_facts query

### LINKED ISSUE #104: perf(store): add LIMIT to list_active_facts query
## Context

Identified during PR #99 (CLI inspector) code review.

`list_active_facts()` loads and deserializes ALL active facts (including embeddings) into memory before any truncation. For the CLI `dump facts --limit 10`, this means a full table scan even when only 10 rows are needed.

## Proposed

Add an optional `limit` parameter to `list_active_facts(limit: Option<usize>)` that pushes `LIMIT N` into the SQL query. This avoids materializing the full corpus for operator/debug commands.

## Labels

performance, Phase 4b

dutiona:
## Summary

- Adds `limit: Option<usize>` parameter to `FactStore::list_active()` and `MemoryEngine::list_active_facts()`, pushing SQL `LIMIT N` into the query when `Some(n)` is provided
- CLI `dump facts --limit N` now materializes only N rows instead of the full corpus + post-hoc truncation
- All internal callers (consolidation, forgetting, conflict resolution) pass `None` for unchanged semantics
- Includes store-level test covering limit boundary conditions (exact, over, zero)

Fixes #104

## Test plan

- [x] New test `list_active_with_limit` verifies: `None` returns all, `Some(2)` returns 2, `Some(100)` returns all (oversized), `Some(0)` returns empty
- [x] All 381 existing tests pass (352 lib + 29 subcrates)
- [x] `cargo clippy --all-targets` — no new warnings

## Files changed (12)

| File | Change |
|------|--------|
| `src/store/facts.rs` | Core: add `limit` param, format SQL with `LIMIT N`, add test |
| `src/engine.rs` | Public API: propagate `limit` param |
| `src/async_engine.rs` | Async wrapper: propagate `limit` param |
| `memory-engine-cli/src/commands/dump.rs` | CLI: pass `Some(limit)`, remove `.truncate()` |
| `src/consolidation/dedup.rs` | Internal caller: pass `None` |
| `src/consolidation/cluster.rs` | Internal caller: pass `None` |
| `src/forgetting/policy.rs` | Internal caller: pass `None` |
| `src/conflict/temporal.rs` | Internal caller: pass `None` |
| `examples/basic_roundtrip.rs` | Example: pass `None` |
| `examples/custom_traits.rs` | Example: pass `None` |
| `docs/reference/api.md` | Update API signature |
| `docs/reference/cli-inspector.md` | Remove known-limitation note |

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1196,13 +1196,16 @@ impl MemoryEngine {
         self.with_read(|conn| FactStore::new(conn, self.embed_dim).get(id))
     }
 
-    /// List all active (non-expired) facts.
+    /// List active (non-expired) facts, optionally limited.
+    ///
+    /// When `limit` is `Some(n)`, a SQL `LIMIT` clause avoids materializing
+    /// the entire corpus. `None` returns all active facts.
     ///
     /// # Errors
     ///
     /// Returns `MemoryError::Database` on query failure.
-    pub fn list_active_facts(&self) -> Result<Vec<Fact>> {
-        self.with_read(|conn| FactStore::new(conn, self.embed_dim).list_active())
+    pub fn list_active_facts(&self, limit: Option<usize>) -> Result<Vec<Fact>> {
+        self.with_read(|conn| FactStore::new(conn, self.embed_dim).list_active(limit))
     }
 
     /// List summaries by consolidation level.
@@ -1919,7 +1922,7 @@ mod tests {
         let stats = engine.consolidate(&MockGen, &config).unwrap();
         assert_eq!(stats.duplicates_removed, 1);
 
-        let active = engine.list_active_facts().unwrap();
+        let active = engine.list_active_facts(None).unwrap();
         assert_eq!(active.len(), 1);
     }
 
@@ -1946,7 +1949,7 @@ mod tests {
         // Second run should find 0 new duplicates
         assert_eq!(stats2.duplicates_removed, 0);
         // Both facts still active
-        assert_eq!(engine.list_active_facts().unwrap().len(), 2);
+        assert_eq!(engine.list_active_facts(None).unwrap().len(), 2);
     }
 
     #[test]
--- a/src/store/facts.rs
+++ b/src/store/facts.rs
@@ -119,19 +119,32 @@ impl<'a> FactStore<'a> {
         }
     }
 
-    /// List all active facts (`t_expired IS NULL`).
+    /// List active facts (`t_expired IS NULL`), optionally limited.
+    ///
+    /// When `limit` is `Some(n)`, a SQL `LIMIT n` clause is pushed into the
+    /// query so that at most `n` rows are materialized. `None` returns all.
     ///
     /// # Errors
     ///
     /// Returns `MemoryError::Database` on query failure.
-    pub fn list_active(&self) -> Result<Vec<Fact>> {
-        let mut stmt = self.conn.prepare(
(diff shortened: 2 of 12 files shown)


src/engine.rs
```diff
@@ -1196,13 +1196,16 @@ impl MemoryEngine {
...
+    ///
+    /// When `limit` is `Some(n)`, a SQL `LIMIT` clause avoids materializing
+    /// the entire corpus. `None` returns all active facts.
     ///
     /// # Errors
     ///
     /// Returns `MemoryError::Database` on query failure.
-    pub fn list_active_facts(&self) -> Result<Vec<Fact>> {
-        self.with_read(|conn| FactStore::new(conn, self.embed_dim).list_active())
+    pub fn list_active_facts(&self, limit: Option<usize>) -> Result<Vec<Fact>> {
```
dutiona, line 1207: [Codex Review] [MEDIUM] This turns a previously zero-arg public API into a breaking signature change for every downstream caller, even though the motivating optimization is CLI-only. Please keep `list_active_facts()` as the compatibility-preserving all-facts entry point and add a separate limited variant/helper for the dump path.

src/store/facts.rs
```diff
@@ -119,19 +119,32 @@ impl<'a> FactStore<'a> {
...
     ///
     /// # Errors
     ///
     /// Returns `MemoryError::Database` on query failure.
-    pub fn list_active(&self) -> Result<Vec<Fact>> {
-        let mut stmt = self.conn.prepare(
+    pub fn list_active(&self, limit: Option<usize>) -> Result<Vec<Fact>> {
+        let sql = if let Some(n) = limit {
+            format!(
+                "SELECT id, content, content_hash, embedding, fact_type,
```
dutiona, line 133: [Codex Review] [MEDIUM] Pushing LIMIT into SQL reduces row count, but this query still selects `embedding` and routes every row through `row_to_fact()`, so `memory-engine-cli dump` continues to read and deserialize each returned embedding even for table/plain output where embeddings are unused. A lightweight projection query would capture the real perf win without widening the core API.

dutiona (commented):
[Codex Review] I left two medium-severity comments. Main issues: (1) the CLI optimization is exposed as a breaking public API change via `list_active_facts(limit: Option<usize>)`, and (2) the limited path still selects/deserializes embeddings, so the performance win is only partial while widening the core API surface. I did not find separate correctness or security regressions in the diff. Local verification on `a7b4e089df60402f295500abbe7e42747bb780e6`: `cargo fmt --check` passed, `cargo test --all-features` passed, and `cargo clippy --all-targets` exited 0 but still emitted many pre-existing warnings on this branch.

src/engine.rs
dutiona: Acknowledged. The `limit: Option<usize>` signature was explicitly specified in #104 ("Add an optional limit parameter to list_active_facts(limit: Option<usize>)"). Since this is a pre-1.0 library with no external consumers, the breaking change is acceptable. All internal callers have been updated to pass `None`.

src/store/facts.rs
dutiona: Valid observation — a lightweight projection query that skips embedding deserialization would capture a bigger win for CLI table/plain output. This is out of scope for #104 (which specifically targets the LIMIT clause). Filed as a follow-up opportunity.

PR #180, opened 2026-03-31
feat(query)!: expose QueryDiagnostics in execute_query results (#152)

### LINKED ISSUE #152: feat(query): abstention type exposure in Query results (four-type taxonomy)
## Context (from research note 18)

The RAG contradictions debate established a four-type abstention taxonomy. Currently ME query results don't expose why results might be insufficient.

## What

Expose abstention type classification in Query results:

| Type | Signal | Meaning |
|------|--------|---------|
| Retrieval | No relevant facts found | Pipeline failure |
| Evidence | Facts found but answer absent | Correct behavior (out-of-scope query) |
| Reasoning | Answer in facts but consumer can't synthesize | Reasoning failure (consumer issue) |
| **Decay** | Facts existed but decayed below threshold | **Deliberate forgetting** — unique to memory systems |

The fourth type (decay) is unique to memory systems: "I knew this once but it's decayed." Unlike retrieval failure, this is by design — but the consumer should know the difference.

## Why

Better consumer diagnostics. The ConflictArbiter trait resolves contradicting facts — it could also expose failure classification.

## Source

- Research note 18 §6 P2, §4.3 (joint consensus)
- Cross-ref: autonomous-agent-project thesis doc update A7

## Phase

4b (expose via MCP server)

dutiona:
## Summary

- Add `QueryDiagnostics` and `QueryResponse` types exposing query pipeline signals for consumer-side abstention classification (four-type taxonomy: Retrieval / Evidence / Reasoning / Decay)
- `execute_query()` now returns `Result<QueryResponse>` wrapping `Vec<SearchResult>` + `QueryDiagnostics`
- Opt-in expired-facts FTS5 COUNT probe via `MemoryQuery::include_expired_probe()`
- MCP `memory_query` tool gains `include_expired_probe` param and returns diagnostics shaped by depth tier

## Breaking Changes

| API | Change |
|-----|--------|
| `execute_query()` | Returns `Result<QueryResponse>` (was `Result<Vec<SearchResult>>`) |
| `AsyncMemoryEngine::execute_query()` | Same |
| `hybrid_search()` | Returns `(Vec<SearchResult>, QueryDiagnostics)` tuple |
| `MemoryQuery` struct | New `include_expired_probe: bool` field |

## Design Decisions

1. **Raw signals, no classification helper** — `QueryDiagnostics` exposes counts. No `AbstentionSignal` enum — expired ≠ decayed (`t_expired` is shared by prune, conflict resolution, and deduplication). The four-type taxonomy is the consumer's job.
2. **`expired_matches` not `decay_matches`** — Honest naming: counts ALL expired facts matching query, regardless of expiry reason.
3. **Opt-in FTS-only probe** — `include_expired_probe: false` by default. Vector-only queries → `expired_matches: None`.

## Test plan

- [x] 5 new unit tests for `fts_count_expired` (zero, match, scope, fact_type, syntax error)
- [x] All ~20 existing `execute_query` tests updated to destructure `.results`
- [x] `cargo test --all-features` — 411 tests pass
- [x] `cargo clippy --all-targets` — no new warnings

Fixes #152

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -9,7 +9,10 @@ use crate::graph::MemoryGraph;
 use crate::pool::ConnectionPool;
 use crate::resume::context::{ResumeConfig, ResumeContext};
 use crate::scope::ScopeTree;
-use crate::search::hybrid::{MatchType, SearchMode, SearchQuery, SearchResult, hybrid_search};
+use crate::search::hybrid::{
+    hybrid_search, MatchType, QueryDiagnostics, QueryResponse, SearchMode, SearchQuery,
+    SearchResult,
+};
 use crate::search::query::MemoryQuery;
 use crate::search::strategy::{BruteForce, SearchConfig, VectorSearchStrategy};
 use crate::store::events::EventStore;
@@ -515,7 +518,7 @@ impl MemoryEngine {
         #[cfg(not(feature = "ann"))]
         let strategy: &dyn VectorSearchStrategy = &*self.vector_strategy;
 
-        let mut results = self.with_read(|conn| {
+        let (mut results, _diagnostics) = self.with_read(|conn| {
             hybrid_search(conn, query, self.embed_dim, scope_ids.as_deref(), strategy)
         })?;
 
@@ -560,7 +563,7 @@ impl MemoryEngine {
     /// - `search_mode` conflicts with available text/embedding inputs
     ///
     /// Returns `MemoryError::Database` on query failure.
-    pub fn execute_query(&self, query: &MemoryQuery) -> Result<Vec<SearchResult>> {
+    pub fn execute_query(&self, query: &MemoryQuery) -> Result<QueryResponse> {
         // --- Validation ---
         self.validate_memory_query(query)?;
 
@@ -570,7 +573,12 @@ impl MemoryEngine {
                 let resolved = self.scope_tree.read().resolve_query(sq);
                 match resolved {
                     Some(ids) => Some(ids),
-                    None => return Ok(vec![]), // scope doesn't exist → no results
+                    None => {
+                        return Ok(QueryResponse {
+                            results: vec![],
+                            diagnostics: QueryDiagnostics::default(),
+                        });
+                    }
                 }
             }
             None => None,
@@ -661,7 +669,7 @@ impl MemoryEngine {
         scope_ids: Option<&[i64]>,
         effective_cutoff: Option<DateTime<Utc>>,
         limit: usize,
-    ) -> Result<Vec<SearchResult>> {
+    ) -> Result<QueryResponse> {
         let mode = Self::infer_search_mode(query);
 
         // When period is set, effective_cutoff is None — but hybrid_search defaults
@@ -698,7 +706,7 @@ impl MemoryEngine {
         #[cfg(not(feature = "ann"))]
         let strategy: &dyn VectorSearchStrategy = &*self.vector_strategy;
 
-        let mut results = self.with_read(|conn| {
+        let (mut results, mut diagnostics) = self.with_read(|conn| {
             hybrid_search(conn, &search_query, self.embed_dim, scope_ids, strategy)
         })?;
 
@@ -718,7 +726,24 @@ impl MemoryEngine {
         }
 
         results.truncate(limit);
-        Ok(results)
+        diagnostics.results_returned = results.len();
+
+        // Expired-facts probe (opt-in, FTS-only).
+        if query.include_expired_probe {
+            if let Some(text) = &query.text {
+                let fact_type_ref = query.fact_type.as_ref();
+                let expired_count = self.with_read(|conn| {
+                    crate::search::fts::fts_count_expired(conn, text, fact_type_ref, scope_ids)
+                })?;
+                diagnostics.expired_matches = Some(expired_count);
+            }
+            // Vector-only queries: expired_matches stays None (documented limitation).
+        }
+
+        Ok(QueryResponse {
+            results,
+            diagnostics,
+        })
     }
 
     /// Store path: no text/vector search, use FactStore directly.
@@ -731,7 +756,7 @@ impl MemoryEngine {
         scope_ids: Option<&[i64]>,
         effective_cutoff: Option<DateTime<Utc>>,
         limit: usize,
-    ) -> Result<Vec<SearchResult>> {
+    ) -> Result<QueryResponse> {
         let scope_slice = scope_ids.unwrap_or(&[]);
 
         // Fetch a broad candidate set. Use the most selective SQL query available,
@@ -792,6 +817,7 @@ impl MemoryEngine {
         }
 
         // Sort by importance_score DESC and truncate
+        let candidates_before_filter = facts.len();
         facts.sort_by(|a, b| {
             b.importance_score
                 .partial_cmp(&a.importance_score)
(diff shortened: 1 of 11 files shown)


src/search/hybrid.rs
```diff
@@ -52,6 +52,43 @@ pub struct SearchResult {
...
+/// of expiry reason (Ebbinghaus decay, conflict resolution, deduplication).
+/// The engine does not currently track expiry provenance — `t_expired` is a
+/// generic tombstone. Consumers wanting true decay-only counts should
+/// cross-reference with `ExpiredReason` via `explain_fact()`.
+#[derive(Debug, Clone, Default, PartialEq, Eq, serde::Serialize)]
+pub struct QueryDiagnostics {
+    /// Total candidates found before post-filters (temporal, importance, pinned).
+    pub candidates_before_filter: usize,
+    /// Total results returned after all filters and truncation.
+    pub results_returned: usize,
```
dutiona, line 74: [Gemini Review] [MEDIUM] `candidates_before_filter` calculation is inconsistent between Search and Store paths. In Search path, it is the count of matches before post-filters (importance, pinned, period). In Store path (`execute_store_path`), it is calculated after most filters because they are applied in SQL. This makes it difficult for consumers to consistently distinguish between Retrieval and Evidence failures as described in Research Note 18.

src/engine.rs
```diff
@@ -515,7 +518,7 @@ impl MemoryEngine {
         #[cfg(not(feature = "ann"))]
         let strategy: &dyn VectorSearchStrategy = &*self.vector_strategy;
 
-        let mut results = self.with_read(|conn| {
+        let (mut results, _diagnostics) = self.with_read(|conn| {
             hybrid_search(conn, query, self.embed_dim, scope_ids.as_deref(), strategy)
```
dutiona, line 522: [Gemini Review] [LOW] API inconsistency: The simpler `query` method still returns `Vec<SearchResult>` and silently discards diagnostics (`_diagnostics`). While this maintains backward compatibility, it means users of this API miss out on the new diagnostic metadata.

src/search/fts.rs
```diff
@@ -82,6 +82,50 @@ pub fn fts_search(
...
+        "SELECT COUNT(*) \
+         FROM facts_fts \
+         JOIN facts AS f ON f.id = facts_fts.rowid \
+         WHERE facts_fts MATCH ?1 \
+           AND f.t_expired IS NOT NULL \
+           AND (?2 IS NULL OR f.fact_type = ?2) \
+           AND (?3 IS NULL OR f.scope_id IN (SELECT value FROM json_each(?3)))",
+    )?;
+
+    let fact_type_str: Option<&str> = fact_type.map(fact_type_to_str);
```
dutiona, line 111: [Gemini Review] [LOW] Error handling: `fts_count_expired` catches all `rusqlite::Error` and returns `Ok(0)`. While this matches `fts_search` behavior for FTS5 syntax errors, it might also hide legitimate database errors (e.g. storage failures, schema mismatches). Consider only falling back to 0 on specific FTS5 syntax errors if possible.

dutiona (commented):
[Gemini Review] This PR successfully introduces query diagnostics metadata to enable advanced abstention classification (Retrieval/Evidence/Reasoning/Decay). Key improvements include the addition of `QueryDiagnostics` and `QueryResponse`, and the opt-in `include_expired_probe` for FTS.

I have raised a few [MEDIUM] and [LOW] concerns regarding API consistency (legacy `query` vs new `execute_query`) and the definition of `candidates_before_filter` in the store-only retrieval path. Overall, the implementation is solid and well-documented.

src/engine.rs
```diff
@@ -718,7 +726,24 @@ impl MemoryEngine {
...
         results.truncate(limit);
-        Ok(results)
+        diagnostics.results_returned = results.len();
+
+        // Expired-facts probe (opt-in, FTS-only).
+        if query.include_expired_probe {
+            if let Some(text) = &query.text {
+                let fact_type_ref = query.fact_type.as_ref();
+                let expired_count = self.with_read(|conn| {
+                    crate::search::fts::fts_count_expired(conn, text, fact_type_ref, scope_ids)
```
dutiona, line 736: [Codex Review] [MEDIUM] The expired probe only reapplies `fact_type` and `scope`. Queries that also set `min_importance_score` or `pinned_only` will still report expired matches that the composed query would have rejected, so `diagnostics.expired_matches` stops describing the composed query. Repro: `text("Rust").pinned_only().include_expired_probe()` still counts expired unpinned facts.

src/search/hybrid.rs
```diff
@@ -190,7 +230,14 @@ pub fn hybrid_search(
         });
     }
 
-    Ok(results)
+    let diagnostics = QueryDiagnostics {
+        candidates_before_filter: results.len(),
```
dutiona: [Codex Review] [MEDIUM] `candidates_before_filter` is documented as the pre-post-filter candidate count, but this assignment happens after the temporal cutoff has already pruned results and after the loop has stopped at `effective_limit`. That makes the field undercount dropped candidates for search queries, and it no longer has the same meaning as the store-path value.

dutiona (commented):
[Codex Review] Found 2 medium-severity correctness issues in the new query diagnostics. First, `expired_matches` ignores `min_importance_score` and `pinned_only`, so it can report expired matches that the composed query would have filtered out. Second, `candidates_before_filter` is populated after temporal pruning and early truncation on the search path, so the same field now has different semantics across code paths and undercounts dropped candidates. Verification: `cargo test --all-features` passed, `cargo fmt --check` passed, and `cargo clippy --all-targets` completed with many pre-existing unrelated warnings.

src/search/hybrid.rs
dutiona: Fixed in f61fbbc: `candidates_before_filter` now uses `ranked.len()` — captured after RRF merge but before temporal filtering and effective_limit truncation.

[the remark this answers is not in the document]
dutiona: Applied this suggestion in f61fbbc. Used `ranked_count` (captured before the consuming `for` loop) since `ranked` is moved into the iterator.

src/engine.rs
```diff
@@ -792,14 +817,25 @@ impl MemoryEngine {
         }
 
         // Sort by importance_score DESC and truncate
+        let candidates_before_filter = facts.len();
         facts.sort_by(|a, b| {
```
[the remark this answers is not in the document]
dutiona: Fixed in f61fbbc: `candidates_before_filter` now captured before all `retain()` calls in the store path.

src/engine.rs
```diff
@@ -792,14 +817,25 @@ impl MemoryEngine {
         }
 
         // Sort by importance_score DESC and truncate
+        let candidates_before_filter = facts.len();
```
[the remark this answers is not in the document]
dutiona: Fixed in f61fbbc: count now captured before post-filters. Re: redundant filters — the retain calls are intentional for AND semantics (the SQL primary query only applies the most selective filter, post-filters guarantee the rest).

src/engine.rs
dutiona: By design: the expired probe answers "how many expired facts match the search terms?" not "how many would survive the full filter chain?" Added explicit doc clarification in f61fbbc. The alternative (replicating pinned_only/min_importance in the COUNT query) would couple the probe to every future filter addition.

src/search/fts.rs
```diff
@@ -82,6 +82,50 @@ pub fn fts_search(
...
+    ) {
+        Ok(n) => Ok(n as usize),
+        Err(rusqlite::Error::QueryReturnedNoRows) => Ok(0),
+        Err(e) => {
+            // FTS5 syntax errors are caught here, same as fts_search.
+            tracing::warn!("FTS5 count_expired query failed (likely syntax error): {e}");
+            Ok(0)
+        }
+    }
+}
```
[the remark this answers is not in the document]
dutiona: Valid point — the SQL and param prep is duplicated. Out of scope for this PR; opened as a follow-up.

dutiona:
Plan archived as #183

PR #185, opened 2026-04-01
test: unit tests for ForgetPolicy and MemoryQuery

### LINKED ISSUE #128: [super-qa] high: traits.rs and query.rs have zero unit tests
## Finding: Missing Unit Tests for Core Types

- **ID:** H6 + H7
- **Severity:** high
- **Category:** testing
- **Location:** src/traits.rs:1-211, src/search/query.rs:1-188

## Description

**traits.rs** has no inline test module. ForgetPolicy::validate() has 5 error paths, default values, and trait object safety — all untested at unit level.

**query.rs** has MemoryQuery builder with 15 public methods and zero tests. Builder defaults, effective_limit() fallback, has_search(), has_period() are not verified.

### traits.rs
- ForgetPolicy::validate() tests for each error branch
- ForgetPolicy::default() field value assertions
- Trait object safety compile-time checks

### query.rs
- new() produces all-None defaults
- effective_limit() returns 50 when None
- has_search() true iff text/embedding set
- Builder chaining preserves fields

Note: These are auto-fixable but were not in Option A scope (correctness-first).

dutiona:
## Summary

- Add 22 unit tests to `src/traits.rs`: `ForgetPolicy::validate()` error paths (5 branches), `ForgetPolicy::default()` field assertions, trait object safety for all 5 consumer traits, `embed_batch` default impl, `PersistenceClassifier` default
- Add 23 unit tests to `src/search/query.rs`: `MemoryQuery::new()` defaults, `effective_limit()` fallback to 50, `has_search()`/`has_period()` predicates, scope builder variants + last-wins, builder chaining field preservation

Closes #128

## Test plan

- [x] `cargo test --all-features` — 465 tests pass (45 new)
- [x] `cargo clippy --all-targets` — clean (pre-existing warnings only)

### DIFF
--- a/src/search/query.rs
+++ b/src/search/query.rs
@@ -204,3 +204,196 @@ impl MemoryQuery {
         self
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // --- new() / Default ---
+
+    #[test]
+    fn new_produces_all_none_defaults() {
+        let q = MemoryQuery::new();
+        assert!(q.scope.is_none());
+        assert!(q.period_start.is_none());
+        assert!(q.period_end.is_none());
+        assert!(q.text.is_none());
+        assert!(q.embedding.is_none());
+        assert!(q.search_mode.is_none());
+        assert!(q.fact_type.is_none());
+        assert!(q.min_importance_score.is_none());
+        assert!(!q.pinned_only);
+        assert!(q.limit.is_none());
+        assert!(q.valid_at.is_none());
+        assert!(!q.include_expired_probe);
+    }
+
+    #[test]
+    fn new_equals_default() {
+        let new = MemoryQuery::new();
+        let def = MemoryQuery::default();
+        // Both should have identical field values
+        assert_eq!(format!("{new:?}"), format!("{def:?}"));
+    }
+
+    // --- effective_limit ---
+
+    #[test]
+    fn effective_limit_returns_50_when_unset() {
+        assert_eq!(MemoryQuery::new().effective_limit(), 50);
+    }
+
+    #[test]
+    fn effective_limit_returns_set_value() {
+        let q = MemoryQuery::new().limit(10);
+        assert_eq!(q.effective_limit(), 10);
+    }
+
+    #[test]
+    fn effective_limit_respects_zero() {
+        let q = MemoryQuery::new().limit(0);
+        assert_eq!(q.effective_limit(), 0);
+    }
+
+    // --- has_search ---
+
+    #[test]
+    fn has_search_false_by_default() {
+        assert!(!MemoryQuery::new().has_search());
+    }
+
+    #[test]
+    fn has_search_true_with_text() {
+        assert!(MemoryQuery::new().text("hello").has_search());
+    }
+
+    #[test]
+    fn has_search_true_with_embedding() {
+        assert!(MemoryQuery::new().embedding(vec![1.0]).has_search());
+    }
+
+    #[test]
+    fn has_search_true_with_both() {
+        let q = MemoryQuery::new().text("hello").embedding(vec![1.0]);
+        assert!(q.has_search());
+    }
+
+    // --- has_period ---
+
+    #[test]
+    fn has_period_false_by_default() {
+        assert!(!MemoryQuery::new().has_period());
+    }
+
+    #[test]
+    fn has_period_true_with_period() {
+        let now = chrono::Utc::now();
+        let q = MemoryQuery::new().period(now - chrono::Duration::hours(1), now);
+        assert!(q.has_period());
+    }
+
+    // --- Scope builders ---
+
+    #[test]
+    fn scope_exact_sets_exact_variant() {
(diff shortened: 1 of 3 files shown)


src/traits.rs
```diff
@@ -244,3 +244,318 @@ impl ForgetPolicy {
...
+
+    #[test]
+    fn default_validates_ok() {
+        ForgetPolicy::default().validate().unwrap();
+    }
+
+    // --- ForgetPolicy::validate() error paths ---
+
+    #[test]
+    fn validate_rejects_zero_half_life() {
```
dutiona, line 303: [Codex Review] [HIGH] These new validation tests still miss the NaN cases, and `ForgetPolicy::validate()` currently accepts them because it only checks `<= 0.0` / `< 0.0`. I verified that with a local repro against the built crate: `half_life_days = f64::NAN` and `recency_weight = f64::NAN` both return `Ok(())`. That leaves a real correctness hole in prune/materialized importance scoring, so this suite should cover NaN rejection too.

src/search/query.rs
```diff
@@ -204,3 +204,196 @@ impl MemoryQuery {
...
+
+    // --- has_period ---
+
+    #[test]
+    fn has_period_false_by_default() {
+        assert!(!MemoryQuery::new().has_period());
+    }
+
+    #[test]
+    fn has_period_true_with_period() {
```
dutiona, line 289: [Codex Review] [MEDIUM] This test only checks `has_period()`, but `has_period()` is just `period_start.is_some() || period_end.is_some()`. A regression where `period()` populated only one bound would still pass here and only fail later in `validate_memory_query()`. Since this PR is meant to lock down builder behavior, please assert that both `period_start` and `period_end` are set to the input timestamps.

dutiona (commented):
[Codex Review] Found 2 issues. [HIGH] The new `ForgetPolicy::validate()` tests miss NaN inputs, and the validator currently accepts `NaN` half-life/weight values; I reproduced that locally against the built crate, which means prune/materialized importance can propagate NaNs. [MEDIUM] `has_period_true_with_period` only asserts `has_period()`, so a `period()` regression that stores only one bound would still pass. Verification: `cargo fmt --check` passed, `cargo test --all-features` unit suite passed, the long-running ANN recall integration test did not finish within the review window, and `cargo clippy --all-targets --all-features -- -D warnings` fails on the current branch with many unrelated baseline issues.

src/traits.rs
dutiona: Fixed: Added `is_finite()` guards to `validate()` for `half_life_days`, override half-lives, and all weight fields. Added 5 NaN/Inf tests (NaN half-life, Inf half-life, NaN override, NaN weights, NaN min_importance).

src/search/query.rs
dutiona: Fixed: Test now asserts both `period_start` and `period_end` match the input timestamps.

src/search/query.rs
```diff
@@ -204,3 +204,196 @@ impl MemoryQuery {
...
+        let def = MemoryQuery::default();
+        // Both should have identical field values
+        assert_eq!(format!("{new:?}"), format!("{def:?}"));
+    }
+
+    // --- effective_limit ---
+
+    #[test]
+    fn effective_limit_returns_50_when_unset() {
+        assert_eq!(MemoryQuery::new().effective_limit(), 50);
```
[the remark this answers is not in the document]
dutiona: Fixed: Using `DEFAULT_LIMIT` constant instead of hardcoded 50.

src/search/query.rs
```diff
@@ -204,3 +204,196 @@ impl MemoryQuery {
...
+    fn has_period_false_by_default() {
+        assert!(!MemoryQuery::new().has_period());
+    }
+
+    #[test]
+    fn has_period_true_with_period() {
+        let now = chrono::Utc::now();
+        let q = MemoryQuery::new().period(now - chrono::Duration::hours(1), now);
+        assert!(q.has_period());
+    }
```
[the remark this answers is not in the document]
dutiona: Fixed: Added `period_start` and `period_end` field assertions.

src/traits.rs
```diff
@@ -244,3 +244,318 @@ impl ForgetPolicy {
...
+            fn embed(&self, _text: &str) -> crate::error::Result<Vec<f32>> {
+                self.0.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+                Ok(vec![1.0, 2.0])
+            }
+        }
+        let c = Counter(std::sync::atomic::AtomicUsize::new(0));
+        let results = c.embed_batch(&["a", "b", "c"]).unwrap();
+        assert_eq!(results.len(), 3);
+        assert_eq!(c.0.load(std::sync::atomic::Ordering::Relaxed), 3);
+        assert_eq!(results[0], vec![1.0, 2.0]);
```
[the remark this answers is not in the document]
dutiona: Fixed: Using tolerance-based comparison with `f32::EPSILON`.

PR #194, opened 2026-04-01
refactor: introduce AddFactRequest, reduce add_fact to 3 params

### LINKED ISSUE #109: [super-qa] high: add_fact takes 8 parameters
## Finding: add_fact Too Many Arguments

- **ID:** H3
- **Severity:** high
- **Category:** design
- **Location:** src/engine.rs:328-422
- **Detected by:** code-reviewer agent

## Description

\`MemoryEngine::add_fact\` takes 8 parameters: self, content, fact_type, source_event_id, embedder, scope, opts, classifier. The \`#[allow(clippy::too_many_arguments)]\` suppression confirms this is a known smell. Same pattern in async wrapper and bootstrap_session_inner.

## Suggested Approach

Introduce an \`AddFactRequest\` builder struct:

\`\`\`rust
pub struct AddFactRequest<'a> {
    pub content: &'a str,
    pub fact_type: FactType,
    pub source_event_id: Option<i64>,
    pub scope: Option<&'a str>,
    pub opts: Option<AddFactOptions>,
    pub classifier: Option<&'a dyn PersistenceClassifier>,
}
\`\`\`

Signature becomes: \`add_fact(&self, req: &AddFactRequest, embedder: &dyn EmbeddingProvider)\`

## Risk

- **Breaking API change** for all consumers
- Coordinate with Phase 4b CLI/MCP work

dutiona:
## Summary

- Rename `BatchFactEntry` → `AddFactRequest` — single input type for both `add_fact` and `add_facts_batch`
- Reduce `add_fact` from 8 parameters to 3: `add_fact(&AddFactRequest, &dyn EmbeddingProvider, Option<&dyn PersistenceClassifier>)`
- Same change propagated to `AsyncMemoryEngine::add_fact`
- Fix pre-existing `flush_insights` embedding integration test failures (missing `index` in mock responses, single-threaded tokio deadlock)

## Motivation

The `#[allow(clippy::too_many_arguments)]` suppression on `add_fact` was a known smell (issue #109). The new `AddFactRequest` struct bundles all data-level parameters (content, fact_type, source_event_id, scope, opts) while keeping infrastructure concerns (embedder, classifier) as separate method arguments.

## Changes

| Area | Files | Description |
|------|-------|-------------|
| Core type | `src/types.rs` | Rename `BatchFactEntry` → `AddFactRequest`, update doc comments |
| Engine API | `src/engine.rs` | New `add_fact(&AddFactRequest, ...)` signature, remove `#[allow(clippy::too_many_arguments)]` |
| Async wrapper | `src/async_engine.rs` | Mirror new signature |
| MCP handler | `memory-engine-mcp/src/tools/mod.rs` | `handle_add_fact` + `handle_flush_insights` use `AddFactRequest` |
| Tests | 19 test/example/bench files | ~170 call sites updated |
| Test fix | `embedding_integration.rs` | Add `index` fields to mock responses, use `multi_thread` flavor |

## Test plan

- [x] `cargo test --all-features` — 420 passed
- [x] `cargo test -p memory-engine-mcp` — 116 passed
- [x] `cargo clippy --all-targets --all-features` — clean

Closes #109

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -184,14 +184,15 @@ pub struct AddFactOptions {
     pub last_accessed: Option<DateTime<Utc>>,
 }
 
-// --- Batch input ---
+// --- Fact request ---
 
-/// Input descriptor for [`crate::engine::MemoryEngine::add_facts_batch`].
+/// Input descriptor for [`crate::engine::MemoryEngine::add_fact`] and
+/// [`crate::engine::MemoryEngine::add_facts_batch`].
 ///
-/// Each entry describes one fact to insert. The engine handles embedding,
-/// classification, scope resolution, and DB insertion atomically for the batch.
+/// Bundles all data-level parameters for fact insertion. Infrastructure
+/// concerns (embedder, classifier) remain separate method parameters.
 #[derive(Debug, Clone)]
-pub struct BatchFactEntry {
+pub struct AddFactRequest {
     /// The fact text to embed and store.
     pub content: String,
     /// Semantic, episodic, procedural, etc.
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Consumers bring their own embedding model, summarizer, and conflict resolver via
 ## Quick Example
 
 ```rust
-use memory_engine::{EngineConfig, MemoryEngine, FactType, EmbeddingProvider, MemoryError};
+use memory_engine::{EngineConfig, MemoryEngine, FactType, AddFactRequest, EmbeddingProvider, MemoryError};
 
 // Consumers implement EmbeddingProvider with their model of choice
 struct DummyEmbedder;
@@ -37,13 +37,15 @@ fn main() -> Result<(), MemoryError> {
 
     // Add a fact (embedding computed automatically)
     let id = engine.add_fact(
-        "Rust's ownership model prevents data races at compile time",
-        FactType::Semantic,
-        None,       // no source event
+        &AddFactRequest {
+            content: "Rust's ownership model prevents data races at compile time".into(),
+            fact_type: FactType::Semantic,
+            source_event_id: None,
+            scope: None,
+            opts: None,
+        },
         &embedder,
-        None,       // root scope
-        None,       // default options
-        None,       // no persistence classifier
+        None, // no persistence classifier
     )?;
 
     // Query with hybrid search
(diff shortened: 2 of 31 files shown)


src/engine.rs
```diff
@@ -422,15 +417,15 @@ impl MemoryEngine {
             None => classifier.is_some_and(|c| {
```
dutiona: [Gemini Review] Constructing a full synthetic `Fact` object and cloning `content`, `embedding`, `fact_type`, and `metadata` just to call `should_pin` can be costly for large batches or large metadata. Consider a more lightweight way to pass fact data to the classifier, or lazily construct the `Fact` object if and only if the classifier is present and `opts.pinned` is `None`.

src/types.rs
```diff
@@ -184,14 +184,15 @@ pub struct AddFactOptions {
...
-// --- Batch input ---
+// --- Fact request ---
 
-/// Input descriptor for [`crate::engine::MemoryEngine::add_facts_batch`].
+/// Input descriptor for [`crate::engine::MemoryEngine::add_fact`] and
+/// [`crate::engine::MemoryEngine::add_facts_batch`].
 ///
-/// Each entry describes one fact to insert. The engine handles embedding,
-/// classification, scope resolution, and DB insertion atomically for the batch.
+/// Bundles all data-level parameters for fact insertion. Infrastructure
```
dutiona, line 192: [Gemini Review] This rename and generalization of `BatchFactEntry` to `AddFactRequest` is a great improvement for API consistency. Consider using `Cow<'a, str>` for the `content` and `scope` fields to allow zero-copy usage when the caller has `&str` and doesn't need to own the request.

dutiona (commented):
[Gemini Review] This PR introduces a significant improvement to the ingest API by consolidating the previously large number of positional arguments into a single `AddFactRequest` struct. This improves maintainability and makes the API more resilient to future changes.

Key observations:
- API consistency is greatly improved by using the same request struct for both single and batch insertions.
- Minor performance concerns identified regarding unnecessary clones during synthetic `Fact` construction for classification.
- Potential for further optimization using `Cow` for string fields in `AddFactRequest`.

Overall, this is a positive refactoring that aligns well with the architectural goals of the project.

src/engine.rs
dutiona: Acknowledged — valid observation but pre-existing: the Fact construction for should_pin was already there before this refactor (same clones, just from req.content instead of content). Fixing this properly requires changing the PersistenceClassifier trait to accept a lighter-weight struct, which is a separate design change. Filed as a follow-up concern, not blocking this PR.

src/types.rs
dutiona: Good suggestion for a follow-up. Cow<str> would avoid allocations in the common case where callers pass literals. Out of scope for this refactor PR — the goal here is strictly reducing parameter count, not optimizing the ownership model.

PR #195, opened 2026-04-01
feat: fast cold-start via sidecar snapshot

### LINKED ISSUE #31: Deferred: Fast cold-start via snapshot + incremental replay
## What it solves

`MemoryEngine::open()` rebuilds `MemoryGraph`, `ScopeTree`, and (after #30) the HNSW index from SQLite on every startup. This is O(N) in fact count — acceptable today but will degrade as data grows.

## Trigger

Cold-start time exceeds ~5s (estimated at ~200K+ facts). Benchmark as part of the work.

## Current behavior

```
SQLite (source of truth)
    ↓ full scan
MemoryGraph + ScopeTree + HNSW  (in-memory, derived)
```

Every startup replays the entire fact table. No caching, no incremental path.

## Proposed approach: Snapshot + Incremental Replay

Event-sourcing gives us the natural optimization: **checkpoint the derived state, replay only the tail.**

### Core mechanism

1. **On shutdown** (or periodically): serialize MemoryGraph + ScopeTree + HNSW state to a sidecar file, tagged with the last processed `event_id`
2. **On startup**: load sidecar → serve immediately → replay only events after the snapshot's `event_id`
3. **Freshness validation**: compare snapshot's `event_id` against SQLite's max `event_id`. If gap is small, incremental replay. If too large or corrupt, fall back to full rebuild.

```
Sidecar file (snapshot @ event_id=N)
    ↓ deserialize (fast, O(1))
Stale-but-usable in-memory state
    ↓ replay events N+1..M from SQLite (O(delta))
Current in-memory state
```

### Startup strategies to evaluate

| Strategy | Latency | Complexity | Correctness risk |
|----------|---------|------------|------------------|
| **Snapshot + tail replay** | O(1) + O(delta) | Low | Low — event log is append-only |
| **Serve stale, rebuild in background** (RCU-style swap) | O(1) immediate, O(N) background | Medium | Queries see stale data until swap |
| **Lazy per-scope reconstruction** | O(1) startup, O(scope) on first query | High | Latency jitter on first query per scope |
| **Scope-bucketed cache** | O(scopes touched) | High | Partial validity tracking nightmare |

**Recommendation:** Start with snapshot + tail replay. It's the simplest, leverages the event-sourced architecture directly, and has no correctness risk (event log is append-only, replay is idempotent).

The RCU-style background rebuild is a good secondary option if we want zero-latency startup with eventual consistency. Could combine: load snapshot → serve → replay tail → atomic swap.

### Open questions / research needed

- **Serialization format**: bincode vs rmp-serde vs custom. Trade-off: speed vs forward-compatibility.
- **Snapshot frequency**: On shutdown only? Periodic? After N events? Need to benchmark snapshot cost.
- **HNSW snapshot**: The `hnsw` crate may or may not support serialization. If not, HNSW-only rebuild from SQLite while MemoryGraph/ScopeTree load from cache.
- **Incremental view maintenance (IVM)**: Database literature on maintaining materialized views incrementally. Worth studying if replay becomes complex.
- **WAL-based freshness**: Can we use SQLite's WAL frame number instead of event_id for cheaper freshness checks?

### Related patterns in literature

- **Event sourcing snapshots**: Standard pattern — Kafka consumer offsets, EventStoreDB projections
- **RCU (Read-Copy-Update)**: Linux kernel pattern for lock-free reads during concurrent updates
- **LSM-tree compaction**: Similar tiered structure (fast recent tier + slow historical tier)
- **Database checkpoint/recovery**: WAL checkpointing in SQLite itself uses the same principle

## Benchmark requirements

Before and after implementation, measure:

| Metric | Sizes |
|--------|-------|
| Cold-start time (full rebuild) | 10K, 50K, 100K, 500K facts |
| Cold-start time (snapshot + replay, delta=0) | Same |
| Cold-start time (snapshot + replay, delta=1K) | Same |
| Snapshot write time | Same |
| Snapshot file size | Same |
| First-query latency (stale vs fresh) | Same |

## Scope

- MemoryGraph, ScopeTree, and HNSW index (if serializable)
- Sidecar file management (write, read, validate, garbage collect)
- Graceful degradation: corrupt/missing sidecar → full rebuild (current behavior)
- Does NOT change the event-sourcing model or SQLite schema

## References

- Issue #12 — LanceDB decision document (notes on cold-start in "Startup Optimization" section)
- Issue #30 — HNSW implementation (adds another structure to rebuild)
- `src/engine.rs` — `init_from_pool()` is the current rebuild entry point

dutiona:
## Summary

- On graceful shutdown, serialize `MemoryGraph` + `ScopeTree` + HNSW state into a `.snapshot` sidecar file (named MessagePack + blake3 checksum)
- On startup, validate against a composite `DbFingerprint` (max_id + active_count from facts/edges/scopes) — load from snapshot or silently fall back to full rebuild
- Atomic write via `tempfile` + `fsync` + rename for crash safety; no-panic contract in `Drop` path

## Design

Dedicated snapshot types decoupled from petgraph/hnsw internals (the "pickle lesson"). Composite fingerprint instead of `MAX(events.id)` — Codex review caught that many mutators (`add_fact`, `forget`, `consolidate`) modify source tables without appending events.

**New file:** `src/engine/snapshot.rs` — types, I/O, validation
**Modified:** `init_from_pool` (snapshot fast path), `MemoryGraph`/`ScopeTree`/`HnswStrategy` (to/from_snapshot), `Drop` impl, `AsyncMemoryEngine`

## Test plan

- [x] 10 unit tests in `snapshot.rs` (path derivation, fingerprint, round-trip, corrupt/stale/wrong-version/checksum rejection, cross-feature serde)
- [x] 7 integration tests in `tests.rs` (full round-trip, stale fingerprint fallback, missing file fallback, in-memory skip, read-only skip, Drop writes, snapshot vs rebuild agree)
- [x] `cargo test --all-features` — 506 passed
- [x] `cargo clippy --all-targets --all-features` — 0 warnings

Fixes #31

### DIFF
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -29,6 +29,15 @@ mod query;
 mod restore;
 mod resume;
 mod scheduling;
+pub(crate) mod snapshot;
+
+/// Data loaded from a sidecar snapshot, ready to be assembled into a `MemoryEngine`.
+struct SnapshotData {
+    graph: MemoryGraph,
+    scope_tree: ScopeTree,
+    #[cfg(feature = "ann")]
+    hnsw_strategy: Option<crate::search::ann::HnswStrategy>,
+}
 
 #[cfg(feature = "archive")]
 mod archive;
@@ -213,6 +222,10 @@ impl MemoryEngine {
     }
 
     /// Shared constructor logic: validate embed_dim, load graph and scope tree.
+    ///
+    /// Tries the snapshot fast path first (file-backed engines only). If the
+    /// sidecar validates against the current DB fingerprint, loads from it.
+    /// Otherwise falls back to full SQLite scan.
     fn init_from_pool(
         pool: ConnectionPool,
         embed_dim: usize,
@@ -220,17 +233,35 @@ impl MemoryEngine {
         upcaster_registry: UpcasterRegistry,
         reranker: Option<Box<dyn Reranker>>,
     ) -> Result<Self> {
-        // Scope the guard so it drops before we move `pool` into the struct.
-        let (graph, scope_tree) = if pool.is_read_only() {
-            // Read-only: use read connection, validate only (no set)
+        // 1. Validate embed_dim (must happen first — ensures schema is ready).
+        if pool.is_read_only() {
             let conn = pool.read();
             Self::validate_embed_dim(&conn, embed_dim)?;
-            let graph = MemoryGraph::load_from_db(&conn)?;
-            let scope_tree = ScopeTree::load(&conn)?;
-            (graph, scope_tree)
         } else {
             let conn = pool.write();
             Self::validate_or_set_embed_dim(&conn, embed_dim)?;
+        }
+
+        // 2. Try snapshot fast path (file-backed engines only).
+        if let Some(loaded) = Self::try_load_snapshot(&pool, embed_dim, &search_config)? {
+            tracing::info!("loaded from snapshot (fingerprint match)");
+            return Ok(Self {
+                pool,
+                embed_dim,
+                graph: RwLock::new(loaded.graph),
+                scope_tree: RwLock::new(loaded.scope_tree),
+                vector_strategy: Box::new(BruteForce),
+                reranker,
+                #[cfg(feature = "ann")]
+                hnsw_strategy: loaded.hnsw_strategy,
+                search_config,
+                upcaster_registry,
+            });
+        }
+
+        // 3. Full rebuild from SQLite (current behavior).
+        let (graph, scope_tree) = {
+            let conn = pool.read();
             let graph = MemoryGraph::load_from_db(&conn)?;
             let scope_tree = ScopeTree::load(&conn)?;
             (graph, scope_tree)
@@ -263,6 +294,55 @@ impl MemoryEngine {
             search_config,
             upcaster_registry,
         })
+    }
+
+    /// Attempt to load in-memory structures from a sidecar snapshot.
+    ///
+    /// Returns `Ok(Some(data))` if the snapshot validated against the current
+    /// DB fingerprint. Returns `Ok(None)` if no snapshot exists or it's
+    /// stale/invalid. Returns `Err` only for unexpected DB query failures.
+    fn try_load_snapshot(
+        pool: &ConnectionPool,
+        embed_dim: usize,
+        #[cfg_attr(not(feature = "ann"), allow(unused_variables))] search_config: &Option<
+            SearchConfig,
+        >,
+    ) -> Result<Option<SnapshotData>> {
+        let Some(db_path) = pool.path() else {
+            return Ok(None); // in-memory engine
+        };
+
+        let snap_path = snapshot::snapshot_path(db_path);
+        let Some((header, payload)) = snapshot::load_from_file(&snap_path, embed_dim) else {
+            return Ok(None);
+        };
+
+        let conn = pool.read();
+        let current_fp = snapshot::read_fingerprint(&conn)?;
+        drop(conn);
+
+        if header.fingerprint != current_fp {
+            tracing::info!("snapshot stale (fingerprint mismatch), falling back to full rebuild");
+            return Ok(None);
+        }
+
+        let graph = MemoryGraph::from_snapshot(&payload.graph);
+        let scope_tree = ScopeTree::from_snapshot(&payload.scope_tree);
+
+        #[cfg(feature = "ann")]
+        let hnsw_strategy = match (search_config, payload.hnsw) {
+            (Some(cfg), Some(ref hnsw_snap)) if cfg.ann_threshold < usize::MAX => Some(
+                crate::search::ann::HnswStrategy::from_snapshot(hnsw_snap, embed_dim)?,
+            ),
+            _ => None,
+        };
+
+        Ok(Some(SnapshotData {
+            graph,
+            scope_tree,
+            #[cfg(feature = "ann")]
+            hnsw_strategy,
+        }))
     }
 
     /// Returns the name of the active reranker, if any.
@@ -336,6 +416,54 @@ impl MemoryEngine {
     #[must_use]
     pub fn is_read_only(&self) -> bool {
         self.pool.is_read_only()
+    }
+
+    /// Write a snapshot of in-memory state to the sidecar file.
+    ///
+    /// No-op for in-memory engines or read-only engines.
+    /// Returns `Ok(false)` if skipped, `Ok(true)` if written.
+    ///
+    /// # No-panic contract
+    ///
+    /// This method never panics. All internal operations use checked access
+    /// and propagate errors via `Result`. Safe to call from `Drop`.
+    pub fn write_snapshot(&self) -> Result<bool> {
+        let Some(db_path) = self.pool.path() else {
+            return Ok(false);
+        };
+        if self.pool.is_read_only() {
+            return Ok(false);
+        }
+
+        let conn = self.pool.read();
+        let fingerprint = snapshot::read_fingerprint(&conn)?;
+
+        let graph_snap = self.graph.read().to_snapshot();
+        let scope_snap = self.scope_tree.read().to_snapshot();
+
+        #[cfg(feature = "ann")]
+        let hnsw_snap = self
+            .hnsw_strategy
+            .as_ref()
+            .map(|h| h.to_snapshot(&conn, self.embed_dim))
+            .transpose()?;
+        #[cfg(not(feature = "ann"))]
+        let hnsw_snap: Option<snapshot::HnswSnapshot> = None;
+
+        let header = snapshot::SnapshotHeader {
+            format_version: snapshot::FORMAT_VERSION,
+            fingerprint,
+            embed_dim: self.embed_dim,
+            engine_version: env!("CARGO_PKG_VERSION").to_string(),
+        };
+        let payload = snapshot::SnapshotPayload {
+            graph: graph_snap,
+            scope_tree: scope_snap,
+            hnsw: hnsw_snap,
+        };
+
+        snapshot::write_to_file(&header, &payload, &snapshot::snapshot_path(db_path))?;
+        Ok(true)
     }
 
     // --- Private connection dispatch helpers ---
@@ -562,5 +690,13 @@ fn fact_to_search_result(fact: Fact) -> SearchResult {
         score: fact.importance_score,
         match_type: MatchType::ImportanceRank,
         fact,
+    }
+}
+
+impl Drop for MemoryEngine {
+    fn drop(&mut self) {
+        if let Err(e) = self.write_snapshot() {
+            tracing::warn!(error = %e, "failed to write snapshot on shutdown");
+        }
     }
 }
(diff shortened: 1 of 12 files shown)


src/engine/mod.rs
```diff
@@ -262,6 +293,55 @@ impl MemoryEngine {
...
+        let graph = MemoryGraph::from_snapshot(&payload.graph);
+        let scope_tree = ScopeTree::from_snapshot(&payload.scope_tree);
+
+        #[cfg(feature = "ann")]
+        let hnsw_strategy = match (search_config, payload.hnsw) {
+            (Some(cfg), Some(ref hnsw_snap)) if cfg.ann_threshold < usize::MAX => Some(
+                crate::search::ann::HnswStrategy::from_snapshot(hnsw_snap, embed_dim)?,
+            ),
+            _ => None,
+        };
```
dutiona, line 346: [Gemini Review] [BLOCKER] Potential HNSW degradation. If `search_config` requires HNSW (threshold < MAX) but the loaded snapshot lacks HNSW data, `hnsw_strategy` will be `None`. This forces the engine into brute-force mode for its entire lifetime even if the fact count exceeds the threshold. `try_load_snapshot` should either return `Ok(None)` (forcing a full rebuild from DB) or trigger a `build_from_db` if HNSW is missing but requested.

src/search/ann.rs
```diff
@@ -136,6 +136,85 @@ impl HnswStrategy {
             embed_dim,
         })
     }
+
+    /// Snapshot active embeddings for fast cold-start.
+    ///
+    /// Reads embeddings from DB (vectors are not kept in memory). Returns
+    /// compact data: only active facts, ordered by fact_id. On load,
+    /// `from_snapshot` rebuilds a fresh compact HNSW index from this data.
+    pub(crate) fn to_snapshot(
```
dutiona, line 145: [Gemini Review] [HIGH] Performance regression on shutdown. `to_snapshot` re-scans the `facts` table to retrieve embeddings. For large databases, this makes `Drop` / `write_snapshot` slow, defeating the purpose of a sidecar by moving the cost to shutdown. Since `HnswStrategy` already has the vectors in memory (the `hnsw` crate owns them), we should extract them directly from the index instead of querying SQLite.

src/search/ann.rs
```diff
@@ -136,6 +136,85 @@ impl HnswStrategy {
...
+
+        Ok(HnswSnapshot { entries })
+    }
+
+    /// Rebuild a compact HNSW index from snapshot data (no DB I/O needed).
+    ///
+    /// Uses the same seed and parameters as `build_from_db` for deterministic
+    /// topology.
+    pub(crate) fn from_snapshot(
+        snap: &crate::engine::snapshot::HnswSnapshot,
```
dutiona, line 175: [Gemini Review] [MEDIUM] Suboptimal cold-start. `from_snapshot` rebuilds the entire HNSW graph topology from embeddings (`O(N log N)`). For large N (e.g., 100k+ facts), this CPU cost limits the effectiveness of the snapshot. If the `hnsw` crate supports it, we should ideally serialize the HNSW graph structure (links) directly into the snapshot payload.

src/engine/snapshot.rs
```diff
@@ -0,0 +1,521 @@
...
+// ---------------------------------------------------------------------------
+
+/// Read a composite fingerprint from the three source-of-truth tables.
+///
+/// Uses a single query with subselects for atomicity within the SQLite
+/// read transaction.
+pub(crate) fn read_fingerprint(conn: &Connection) -> Result<DbFingerprint> {
+    conn.query_row(
+        "SELECT \
+            COALESCE((SELECT MAX(id) FROM facts), 0), \
```
dutiona, line 130: [Gemini Review] [LOW] Fingerprint robustness. The current fingerprint only tracks `MAX(id)` and `COUNT(*)` for facts/edges/scopes. This can miss metadata-only updates (e.g., pinning a fact, changing a scope label if that feature existed). While not critical for correctness (pinned status is read from DB during search), including the SQLite `schema_version` or a hash of the `scopes` table would make the snapshot validation more bulletproof.

src/engine/snapshot.rs
```diff
@@ -0,0 +1,521 @@
...
+// ---------------------------------------------------------------------------
+// File I/O
+// ---------------------------------------------------------------------------
+
+/// Write a snapshot to disk atomically.
+///
+/// 1. Serialize header and payload with `rmp_serde::to_vec_named` (named
+///    MessagePack for forward/backward compatibility).
+/// 2. Compute blake3 hash over the payload bytes.
+/// 3. Write to a temp file in the same directory (`NamedTempFile` uses
```
dutiona, line 160: [Gemini Review] [LOW] Snapshot compression. For large deployments (e.g., 100k facts with 1536-dim embeddings), the snapshot payload can exceed several hundred megabytes. Since `compress-zstd` is already a project feature, we should consider wrapping the payload MessagePack in a zstd stream to reduce disk I/O and storage footprint.

dutiona (commented):
[Gemini Review] The PR introduces a robust sidecar snapshot mechanism for fast cold-starts. However, there is a critical bug where loading a snapshot created without HNSW data (e.g. from a session where it was disabled) will silently force the engine into brute-force mode for its entire lifetime if HNSW is requested. Additionally, shutdown performance is regressed by a redundant DB scan for embeddings that are already in memory. See inline comments for details and other minor findings.

src/engine/mod.rs
```diff
@@ -335,6 +415,54 @@ impl MemoryEngine {
...
+    pub fn write_snapshot(&self) -> Result<bool> {
+        let Some(db_path) = self.pool.path() else {
+            return Ok(false);
+        };
+        if self.pool.is_read_only() {
+            return Ok(false);
+        }
+
+        let conn = self.pool.read();
+        let fingerprint = snapshot::read_fingerprint(&conn)?;
```
dutiona, line 453: [Codex Review] [BLOCKER] `write_snapshot()` fingerprints the live DB here, but the payload comes from this engine instance's in-memory `graph`/`scope_tree`. That becomes unsound as soon as two `MemoryEngine`s share one DB: engine A can add a new scope or edge, engine B stays stale, and if B writes the snapshot (or just drops last) it emits a snapshot whose fingerprint matches the DB but whose caches are missing A's update. I reproduced this locally with two file-backed engines on one DB: A added `user:alice/project:demo`, B dropped last, and the next reopen failed `next_due_time(Some("user:alice/project:demo"))` with `not found`. This needs a write-side rebuild/epoch check before persisting, otherwise stale instances can poison future cold starts.

src/engine/mod.rs
```diff
@@ -262,6 +293,55 @@ impl MemoryEngine {
...
+        if header.fingerprint != current_fp {
+            tracing::info!("snapshot stale (fingerprint mismatch), falling back to full rebuild");
+            return Ok(None);
+        }
+
+        let graph = MemoryGraph::from_snapshot(&payload.graph);
+        let scope_tree = ScopeTree::from_snapshot(&payload.scope_tree);
+
+        #[cfg(feature = "ann")]
+        let hnsw_strategy = match (search_config, payload.hnsw) {
```
dutiona, line 333: [Codex Review] [HIGH] This accepts a snapshot written by a non-`ann` build, but then silently disables ANN instead of rebuilding it from SQLite. When `search_config` enables ANN and `payload.hnsw` is `None`, the fast path returns `hnsw_strategy = None`, so the engine falls back to brute force even though a cold open without the snapshot would build HNSW. I reproduced this by writing a snapshot from a default build, reopening with `--features ann` and `ann_threshold = 0`, and observing `active_strategy_name() == "brute_force"`; deleting the snapshot and reopening yielded `"hnsw"`. If cross-feature snapshots are meant to be supported, the missing-`hnsw` case needs to trigger a DB rebuild instead of silently degrading search.

dutiona (commented):
[Codex Review] Found 2 issues: 1 blocker and 1 high-severity regression. The blocker is that stale engine instances can write snapshots whose DB fingerprint is current but whose `graph`/`scope_tree` caches are stale; I reproduced that with two file-backed engines on the same DB and a missing scope on reopen. The second is that ANN builds loading a non-ANN snapshot silently fall back to brute-force instead of rebuilding HNSW; I reproduced `brute_force` with the snapshot present and `hnsw` after deleting it. Verification: `cargo test --all-features` passed, `cargo fmt --check` passed, and `cargo clippy --all-targets --all-features -- -D warnings` still fails in this branch (including new unused-import warnings).

src/engine/mod.rs
dutiona: Fixed in 1b7f170: when snapshot lacks HNSW data but config requires ANN, now falls back to `HnswStrategy::build_from_db()` instead of accepting `None`. Graph + scope tree still load from snapshot.

src/engine/mod.rs
dutiona: Fixed in 1b7f170: same fix as the Gemini finding above — detects missing HNSW in snapshot when ANN is required and rebuilds from DB.

src/search/ann.rs
dutiona: Acknowledged as known V1 limitation. Shutdown cost is bounded by fact count and amortized over the session. Could be optimized in V2 by caching vectors in memory or using HNSW native serde.

src/search/ann.rs
dutiona: By design: decoupled snapshot types avoid coupling to hnsw crate internals. The rebuild is still faster than full DB scan (no I/O). format_version header allows upgrading to direct HNSW serde later if profiling warrants.

src/engine/mod.rs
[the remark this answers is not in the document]
dutiona: Fixed in 1b7f170: when snapshot lacks HNSW data but config requires ANN, now falls back to `HnswStrategy::build_from_db()` instead of accepting `None`.

src/search/ann.rs
dutiona: Tracked as #199 for V2. Added to ROADMAP.

src/search/ann.rs
dutiona: Tracked as #200 for V2. Added to ROADMAP.

src/engine/snapshot.rs
dutiona: Out of scope for V1. The snapshot structures (MemoryGraph, ScopeTree, HNSW) don't store metadata fields like `is_pinned` or `importance_score` — those are read from DB at query time. A stale snapshot with matching fingerprint but different metadata is still correct because the snapshot only caches the graph topology and scope hierarchy, not per-fact metadata.

src/engine/snapshot.rs
dutiona: Tracked as #201 for V2. Added to ROADMAP.

src/engine/mod.rs
dutiona: Single-writer is a documented constraint of `MemoryEngine` — opening multiple engines against the same SQLite file is unsupported and already dangerous at the SQLite level (WAL mode allows concurrent readers but only one writer, and in-memory caches would diverge regardless of snapshots). This is not a snapshot-specific issue. Added a doc comment to `write_snapshot()` noting this precondition.

PR #196, opened 2026-04-01
feat(archive): Phase 4c archival compression — cold storage .pak files (#46)

### LINKED ISSUE #46: Phase 4c: Archival compression — cold storage .pak files
### Scope

Cold storage for long-running agents (10 agents x 10 years).

- Extract old, non-pinned memory chunks into compressed `.pak` files (zstd)
- Not automatic — consumer explicitly triggers archival
- Live DB retains: all pinned facts, recent facts, active edges, summaries
- Archived `.pak` files are a slow fallback — queries warn consumer with ETA
- Archival is compaction that preserves the event log (compressed, not discarded)

### Design reference

See `docs/design/plans/2026-03-09-future-phases-design.md` § Phase 4 → Archival Compression.

### Dependencies

- Phase 4a import/export (#40)
- #31 fast cold-start (related — snapshot + replay)

dutiona:
## Summary

- Add consumer-triggered archival compression that moves expired, non-pinned facts into zstd-compressed `.pak` files alongside the live SQLite database
- New `archive` feature flag (implies `compress-zstd`)
- Schema v6→v7 adds `archive_manifest` table to track `.pak` files
- Opt-in archive query fallback via `MemoryQuery::include_archives()`
- Async mirror for `archive()`, `list_archives()`, `verify_archives()`

### Key design decisions
- **Consumer-triggered only** — the engine never archives automatically
- **Read lock for candidate selection, write lock only for the transaction** — minimizes blocking
- **blake3 hash computed during write** via `HashingWriter` — eliminates TOCTOU race
- **4 GiB decompression bomb guard** — `io::Take` on zstd decoder
- **Hard-delete from live DB** after archival — `.pak` file IS the preservation; events stay in live DB
- **Edges archived only when both endpoints are in the candidate set** — edges with one live endpoint stay

### New public API (gated behind `archive` feature)
- `MemoryEngine::archive(&self, policy: &ArchivePolicy) -> Result<Option<ArchiveStats>>`
- `MemoryEngine::list_archives(&self) -> Result<Vec<ArchiveManifestEntry>>`
- `MemoryEngine::verify_archives(&self) -> Result<Vec<ArchiveVerifyResult>>`
- `MemoryQuery::include_archives()` — opt-in slow fallback search
- `MatchType::Archive` variant + `QueryDiagnostics` archive fields

## Test plan

- [x] 493 tests pass with `cargo test --features archive`
- [x] 476 tests pass without feature (no regressions)
- [x] Integration tests: lifecycle roundtrip, pinned-facts skip, min_facts guard, archive search fallback
- [x] Unit tests: pak write/read roundtrip, blake3 hash consistency, atomic write, FTS5 cleanup on hard-delete, manifest CRUD, edge deletion semantics
- [ ] CI pipeline

Fixes #46


src/engine/archive.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+        Ok(results)
+    }
+
+    // --- Private helpers ---
+
+    /// Select expired, non-pinned facts and their internal edges.
+    fn select_archive_candidates(&self, policy: &ArchivePolicy) -> Result<(Vec<Fact>, Vec<Edge>)> {
+        let conn = self.pool.read();
+        let all_facts = FactStore::new(&conn, self.embed_dim).list_all()?;
```
dutiona, line 153: [Code Review] [MEDIUM] `select_archive_candidates` calls `list_all()` on both facts and edges, loading the entire dataset into memory to filter in Rust. For a database with millions of facts where only a few thousand are expired, this is wasteful.

Consider a SQL-side filter instead:

```sql
SELECT ... FROM facts WHERE is_pinned = 0 AND t_expired IS NOT NULL AND t_expired < ?1
```

Similarly for edges, you could join against the candidate fact IDs rather than loading all edges. This would reduce memory usage from O(total_facts) to O(candidates).

Not blocking since this is consumer-triggered and the embedded use case likely has manageable dataset sizes, but worth noting for scalability.

src/archive/search.rs
```diff
@@ -0,0 +1,113 @@
...
+    manifest_entries: &[ArchiveManifestEntry],
+    query: &MemoryQuery,
+    limit: usize,
+) -> Result<ArchiveSearchResult> {
+    let start = Instant::now();
+    let mut all_results: Vec<SearchResult> = Vec::new();
+    let mut paks_scanned = 0usize;
+
+    for entry in manifest_entries {
+        let pak_path = archive_dir.join(&entry.pak_path);
```
dutiona, line 45: [Code Review] [MEDIUM] Path traversal risk: `archive_dir.join(&entry.pak_path)` uses `pak_path` from the database without validation. If the manifest is corrupted or tampered with, a crafted `pak_path` like `../../sensitive_file` would cause `read_pak` to open files outside the archive directory.

Suggest adding a check that `pak_path` contains no path separators (or that the resolved path starts with `archive_dir`):

```rust
let pak_path = archive_dir.join(&entry.pak_path);
if !pak_path.starts_with(archive_dir) {
    // skip or error: path traversal detected
    continue;
}
```

This applies here and also in `verify_archives` (line 117 of `engine/archive.rs`).

src/engine/query.rs
```diff
@@ -261,6 +261,29 @@ impl MemoryEngine {
             // Vector-only queries: expired_matches stays None (documented limitation).
         }
 
+        // Archive fallback (opt-in, best-effort).
+        #[cfg(feature = "archive")]
+        if query.include_archives {
```
dutiona, line 266: [Code Review] [LOW] Archive search fallback is only wired into the search path (text/embedding queries). The store path (`execute_store_path`, used when neither text nor embedding is set) does not check `include_archives`. A `MemoryQuery::new().include_archives()` call with no search terms silently skips archive search. Either document this limitation on the `include_archives()` builder, or add the archive fallback to the store path as well.

src/archive/pak.rs
```diff
@@ -0,0 +1,160 @@
...
+#[allow(dead_code)]
+pub fn write_pak(pak: &ArchivePak, path: &Path) -> Result<()> {
+    write_pak_and_hash(pak, path)?;
+    Ok(())
+}
+
+/// Write a `.pak` file and return its blake3 hash.
+/// Hash is computed during write (no TOCTOU). Atomic write via tmp+rename.
+pub fn write_pak_and_hash(pak: &ArchivePak, path: &Path) -> Result<String> {
+    let tmp_path = path.with_extension("pak.tmp");
```
dutiona, line 24: [Code Review] [LOW] If `write_pak_and_hash` fails after creating the tmp file (e.g. during zstd encoding or JSON serialization) but before the `fs::rename`, the `.pak.tmp` file is left behind. Consider cleaning it up in the error path, or using a RAII guard that removes the tmp file on drop unless explicitly disarmed after successful rename.

dutiona (commented):
[Code Review] Phase 4c archival compression -- solid implementation with no blocking issues.

## Verdict: APPROVE (posted as comment -- cannot approve own PR)

## What was done well

- **Crash safety sequence** is correct: write .pak -> commit SQLite transaction (manifest insert + hard-delete) -> update in-memory graph. A crash at any point either leaves facts in the DB (safe) or leaves an orphan .pak file (harmless, verifiable).
- **HashingWriter** computes blake3 during write, eliminating TOCTOU between write and hash verification. Clean design.
- **Atomic write** via tmp+rename prevents partial .pak files from being visible to readers.
- **Edge archival semantics** are correct: only edges where both endpoints are in the candidate set are archived, preventing dangling references.
- **Feature gating** is well done: `archive_manifest` table exists unconditionally in the schema (schema stability), but all code paths are behind `#[cfg(feature = "archive")]`.
- **Schema migration** v6->v7 is clean and tested, including frozen v6 DDL for migration testing.
- **Test coverage** is thorough: lifecycle roundtrip, pinned-fact exclusion, min_facts guard, archive search fallback, FTS5 cleanup on hard-delete, edge deletion semantics, pak write/read roundtrip, blake3 consistency.
- **API design** is conservative: consumer-triggered only, opt-in search, clear `ArchivePolicy` controls.

## Findings

### [MEDIUM] `select_archive_candidates` loads all facts/edges into memory
See inline comment on `src/engine/archive.rs:138`. `list_all()` is O(total_facts), should be a SQL-side filter. Not blocking for the embedded use case but worth a follow-up.

### [MEDIUM] Path traversal in archive search and verify
See inline comment on `src/archive/search.rs:45`. `archive_dir.join(&entry.pak_path)` trusts the DB value without validation. Add `pak_path.starts_with(archive_dir)` check. Also applies to `verify_archives` in `engine/archive.rs:117`.

### [LOW] Orphaned .pak.tmp on write failure
See inline comment on `src/archive/pak.rs:24`. No cleanup of the temp file on error.

### [LOW] Archive fallback missing in store path
See inline comment on `src/engine/query.rs:266`. `include_archives` is silently ignored when no text/embedding is set.

## Not flagged (reviewed and OK)

- Read-only guard: `archive()` calls `write_conn()` which checks `try_write()` -- correctly rejects read-only engines.
- Decompression bomb: 4 GiB limit via `io::Take` is reasonable for single-pak safety.
- Concurrency: read lock for candidate selection, write lock only for the commit transaction. Graph write lock is scoped tightly.
- `MatchType::Archive` variant is `#[non_exhaustive]` -- existing match arms won't break.
- `ArchivePak.pak_version` field enables future format evolution.
- Event log is preserved (only facts+edges are archived, not events) -- consistent with event-sourcing design.

493 tests pass with `--features archive`, 476 without. Clippy clean. No regressions.

Issue #229, opened 2026-04-08
plan: #224 activity stream + session lifecycle MCP endpoints


dutiona:
# Plan: #224 — Activity Stream + Session Lifecycle MCP Endpoints

## Context

Issue #224 (sub-issue of #221 umbrella) adds hook-driven activity recording, session checkpointing, and project-scoped context loading to the MCP server. This is the foundational layer that enables Claude Code hooks to stream tool invocations into memory-engine and retrieve project context at session start.

**Why now:** This is the only part of #221 that has no phase dependencies. The cognitive endpoints (#225) and cross-layer linking (#226) are blocked on Phase 5a and Phase 6 respectively.

**Worktree:** `.worktrees/feat-224-activity-stream` (branch: `feat/224-activity-stream`)

---

## Design Decisions

1. **Separate `activities` table** (not reusing `events`): Activities are high-volume (every tool call), need dedup counters (`occurrence_count`), status tracking, and args hashing — none of which fit the event schema. Activities have a shorter lifecycle than events (which are the permanent source of truth for memory ops).

2. **Pre-storage filtering split**: Ignore happens pre-storage (never persisted). Dedup happens at INSERT time (counter increment or new row). Promote happens post-insert, **gated on `was_deduplicated == false`** — only fresh activities can promote to facts. Cluster is deferred — requires embeddings/LLM.

3. **`load_context` complements `resume_context`**: `resume_context` is session-scoped, fact-only (5-tier cognitive boot). `load_context` is scope-path-scoped, cross-session, and includes activities + checkpoints + facts. Both take a `scope` parameter — `load_context` uses it to find the most recent checkpoint and recent activities for that scope subtree.

4. **`session_checkpoints` uses UPSERT**: Last-write-wins via `ON CONFLICT(session_id) DO UPDATE`, matching the Stop hook pattern where sessions get at most one active checkpoint. The `last_activity_id` FK is nullable with `ON DELETE SET NULL` so activities can be GC'd without corrupting checkpoints.

5. **`args_hash` via blake3, 32 hex chars**: Same 128-bit truncation as `content_hash` on facts (see `src/store/facts.rs:33`). Dedup key includes an **outcome class** signal alongside `(session_id, tool_name, args_hash)` to distinguish semantically different results (e.g., test pass vs fail).

[...]

Issue #221, opened 2026-04-08
feat: expose MCP endpoints for hook-based activity and session lifecycle


dutiona:
## Umbrella: MCP endpoints for hook-based activity and session lifecycle

Extracted from dutiona/knowledge-base#102 (umbrella: hook-based memory-engine integration).

The two-stream architecture requires memory-engine to expose MCP endpoints that hooks and the model can call for activity recording, session lifecycle, and context retrieval.

## Sub-issues

| Issue | Scope | Blocked on |
|-------|-------|-----------|
| #224 | Activity stream + session lifecycle (`record_activity`, `checkpoint_session`, `load_context`) + server-side filtering | **Nothing** — ready now |
| #225 | Cognitive pipeline (`dream_cycle`, `get_recent_insights`) | Phase 5a: #48, #49, #55, #63 |
| #226 | Cross-layer linking (`link`, `query_linked`) | Phase 6: #50 |

## Critical Path to Close

```
#224 (activity stream)          ─── can start now
     │
Phase 5a (#55,#63 → #49,#48)   ─── in progress / next
     │
#225 (cognitive endpoints)      ─── after Phase 5a
     │
Phase 6 (#50)                   ─── after Phase 5b
     │
#226 (cross-layer linking)      ─── after Phase 6
     │
#221 closes                     ─── when #224 + #225 + #226 all done
```

## Original Design

See below for the full endpoint list and server-side filtering strategy from the original issue.

### Activity Stream (hook-driven, high-volume)

- **`record_activity(tool, args, result, session_id, timestamp)`** — PostToolUse hook calls this for every tool invocation. ME filters server-side (dedup, ignore, promote, cluster).
- **`checkpoint_session(session_id)`** — Stop hook calls this for lightweight last-write-wins snapshot when a session ends or pauses.

### Consolidation

- **`dream_cycle()`** — SessionStart hook (daily) triggers Ollama-based dedup/pruning, replacing the current flat-file `memory-consolidate.sh`. Infers intent from activity patterns to bridge compaction gaps.

### Context Retrieval (session bootstrap)

- **`load_context(project_path)`** — SessionStart hook calls this to inject active project context into a new session.
- **`resume_context(session_id?)`** — Returns recent insights and activity summary for session continuity.
- **`get_recent_insights(project_path, limit?)`** — Returns high-value model-logged insights (insight stream).

### Bidirectional Linking (cross-layer)

[...]

PR #230, opened 2026-04-08
feat: activity stream + session lifecycle MCP endpoints (#224)

### LINKED ISSUE #224: feat(mcp): activity stream + session lifecycle endpoints
## Parent

Umbrella: #221

## Context

Subset of #221 that can be built on current infrastructure. These endpoints enable hook-based activity recording and session lifecycle management without depending on Phase 5a (DreamCycle, InsightStream) or Phase 6 (KnowledgeBaseConnector).

### Activity Stream (hook-driven, high-volume)

- **`record_activity(tool, args, result, session_id, timestamp)`** — PostToolUse hook calls this for every tool invocation. ME filters server-side (dedup, ignore, promote, cluster).
- **`checkpoint_session(session_id)`** — Stop hook calls this for lightweight last-write-wins snapshot when a session ends or pauses.

### Context Retrieval (session bootstrap)

- **`load_context(project_path)`** — SessionStart hook calls this to inject active project context into a new session. Project-scoped (vs existing `resume_context` which is session-scoped).

### Server-Side Filtering (Granularity Strategy)

`record_activity` receives everything. ME filters:
- **Dedup:** 10× "read file X" → 1 event "explored file X"
- **Ignore:** formatting edits, lint fixes
- **Promote:** test failures, commits, new files
- **Cluster:** related edits → "refactored search pipeline"

## Schema Impact

New tables likely needed:
- `activities` — raw activity events (tool, args hash, result summary, session_id, timestamp)
- `session_snapshots` — checkpoint data (session_id, project_path, snapshot blob, timestamp)

## Acceptance Criteria

- [ ] `record_activity` accepts tool events and persists them
- [ ] `checkpoint_session` writes a lightweight session snapshot
- [ ] `load_context` returns project-scoped context
- [ ] Server-side filtering reduces raw activity volume
- [ ] All endpoints exposed as MCP tools
- [ ] Integration tests for each endpoint

## Dependencies

- **No prerequisites** — builds on existing MCP server infrastructure (#45) and schema evolution policy (#18)

## Refs

- Umbrella: #221
- NYX12 architecture: dutiona/knowledge-base#102

dutiona:
## Summary

- Add 3 new MCP endpoints: `memory_record_activity`, `memory_checkpoint_session`, `memory_load_context`
- Schema v8→v9 migration adding `activities` and `session_checkpoints` tables
- Server-side activity filtering: dedup (outcome-class-aware), ignore, and promote pipeline
- Adapter-specific Claude Code policy in `activity_policy.rs` (core stays generic)
- Full depth shaping (sparse/standard/full) for activity and context responses

Closes #224 (sub-issue of #221 umbrella). Plan: #229.

## Key Design Decisions

1. **Separate `activities` table** — not reusing `events` (different lifecycle, dedup counters)
2. **Outcome-class in dedup key** — `cargo test` pass vs fail within window are NOT collapsed
3. **Promotion gated on `was_deduplicated == false`** — prevents duplicate fact creation (Codex BLOCKER fix)
4. **Scope paths, not filesystem paths** — aligns with existing `ScopeTree` model
5. **Filtering policy in MCP adapter** — core engine is generic, adapter owns tool-name heuristics
6. **Single read snapshot** for `load_context` — consistent cross-query results

## Test plan

- [x] Schema migration test (`migrate_v8_to_v9_adds_activities_and_checkpoints`)
- [x] Fresh DB test (`fresh_db_has_activities_and_checkpoints`)
- [x] ActivityStore unit tests (insert, dedup, outcome differentiation, list, update)
- [x] CheckpointStore unit tests (upsert, overwrite, scope lookup)
- [x] Activity filter unit tests (ignore, promote, precedence, truncation)
- [x] Integration test: full record → checkpoint → load_context cycle
- [x] MCP tool count assertion updated (20 → 23)
- [x] All existing tests updated for new dispatch signature
- [x] `cargo build --workspace && cargo test --workspace && cargo clippy --workspace --all-targets` passes

### DIFF
--- a/memory-engine-mcp/src/tools/mod.rs
+++ b/memory-engine-mcp/src/tools/mod.rs
@@ -4,7 +4,6 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use chrono::{DateTime, Utc};
-use memory_engine::ResumeConfig;
 use memory_engine::bootstrap::{BootstrapConfig, KeywordExtractor};
 use memory_engine::engine::MemoryEngine;
 use memory_engine::inspect_types::{DumpFormat, FactExplanation, ReplayFilter, ReplayOrder};
@@ -15,12 +14,13 @@ use memory_engine::traits::{
 use memory_engine::types::{
     AddFactOptions, AddFactRequest, EventType, FactType, NewEvent, Outcome,
 };
+use memory_engine::ResumeConfig;
 use rmcp::model::{CallToolResult, Content, ErrorData, Tool};
-use serde_json::{Map, Value, json};
+use serde_json::{json, Map, Value};
 
 use crate::depth::{self, Depth};
 use crate::embedding::{HttpEmbeddingProvider, PassthroughEmbedder};
-use crate::error::{ValidationError, to_mcp_error};
+use crate::error::{to_mcp_error, ValidationError};
 
 // ---------------------------------------------------------------------------
 // Tool definitions (JSON schemas)
@@ -1374,8 +1374,8 @@ fn handle_load_context(
 ) -> Result<CallToolResult, ErrorData> {
     let scope =
         get_str(&args, "scope").ok_or_else(|| ErrorData::invalid_params("missing scope", None))?;
-    let activity_limit = get_i64(&args, "activity_limit").unwrap_or(20) as usize;
-    let fact_limit = get_i64(&args, "fact_limit").unwrap_or(10) as usize;
+    let activity_limit = get_usize(&args, "activity_limit").unwrap_or(20);
+    let fact_limit = get_usize(&args, "fact_limit").unwrap_or(10);
     let depth_level = get_depth(&args);
 
     let ctx = engine
--- a/memory-engine-mcp/src/tools/mod.rs
+++ b/memory-engine-mcp/src/tools/mod.rs
@@ -312,6 +312,52 @@ pub fn all_tool_definitions() -> Vec<Tool> {
                 "required": ["fact_id"]
             }),
         ),
+        // Activity stream + session lifecycle (#224)
+        tool_def(
+            "memory_record_activity",
+            "Record a tool invocation activity. Server-side filtering deduplicates, ignores formatting, and can promote significant actions to facts.",
+            json!({
+                "type": "object",
+                "properties": {
+                    "tool": { "type": "string", "description": "Tool name that was invoked" },
+                    "args": { "description": "Tool arguments (arbitrary JSON)" },
+                    "result": { "type": "string", "description": "Tool result summary (truncated at 512 chars)" },
+                    "session_id": { "type": "string", "description": "Current session ID" },
+                    "timestamp": { "type": "string", "format": "date-time", "description": "ISO 8601 timestamp. Defaults to now." },
+                    "scope": { "type": "string", "description": "Scope path for the activity" },
+                    "outcome_class": { "type": "string", "description": "Outcome class (e.g. 'success', 'error', 'test_failure'). Default: 'success'" }
+                },
+                "required": ["tool", "session_id"]
+            }),
+        ),
+        tool_def(
+            "memory_checkpoint_session",
+            "Checkpoint the current session (last-write-wins). Called by Stop hook.",
+            json!({
+                "type": "object",
+                "properties": {
+                    "session_id": { "type": "string", "description": "Session ID to checkpoint" },
+                    "scope": { "type": "string", "description": "Scope path (e.g. 'project:memory-engine')" },
+                    "summary": { "type": "string", "description": "Free-form session summary" },
+                    "metadata": { "description": "Arbitrary JSON metadata" }
+                },
+                "required": ["session_id"]
+            }),
+        ),
+        tool_def(
+            "memory_load_context",
+            "Load active project context for session start. Returns recent activities, last checkpoint, and relevant facts.",
+            json!({
+                "type": "object",
+                "properties": {
+                    "scope": { "type": "string", "description": "Scope path (e.g. 'project:memory-engine')" },
+                    "activity_limit": { "type": "integer", "default": 20, "description": "Max recent activities to return" },
+                    "fact_limit": { "type": "integer", "default": 10, "description": "Max relevant facts to return" },
+                    "depth": { "type": "string", "enum": ["sparse", "standard", "full"], "default": "standard" }
+                },
+                "required": ["scope"]
+            }),
+        ),
     ]
 }
 
@@ -335,6 +381,7 @@ pub fn dispatch(
     embedder: Option<&HttpEmbeddingProvider>,
     summary_gen: Option<&(dyn SummaryGenerator + Send + Sync)>,
     embed_dim: usize,
+    filter_config: &memory_engine::ActivityFilterConfig,
 ) -> Result<CallToolResult, ErrorData> {
     match name {
         "memory_ingest" => handle_ingest(args, engine),
@@ -360,6 +407,10 @@ pub fn dispatch(
         // Phase 5a: Outcome tracking
         "memory_record_outcome" => handle_record_outcome(args, engine),
         "memory_outcome_counts" => handle_outcome_counts(args, engine),
+        // Activity stream + session lifecycle (#224)
+        "memory_record_activity" => handle_record_activity(args, engine, embedder, filter_config),
+        "memory_checkpoint_session" => handle_checkpoint_session(args, engine),
+        "memory_load_context" => handle_load_context(args, engine),
         _ => Err(ErrorData::invalid_params(
             format!("unknown tool: {name}"),
             None,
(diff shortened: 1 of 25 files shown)


dutiona:
## Implementation Status

All 5 implementation tasks complete (5 commits, ~1960 lines added):

| Commit | Scope |
|--------|-------|
| Schema v8→v9 | `activities` + `session_checkpoints` tables, 5 indexes, 7 new types |
| Store layer | `ActivityStore` (dedup-aware insert, scope queries) + `CheckpointStore` (UPSERT) |
| Activity filter | Generic `ActivityFilterConfig` + `apply_filter()` with ignore/promote patterns |
| Engine methods | `record_activity`, `checkpoint_session`, `load_context` on `MemoryEngine` |
| MCP endpoints | 3 tools + Claude Code adapter policy + depth shaping |

### Test coverage

- **38 new tests** (schema migration, store CRUD, filter logic, integration cycle, tool count)
- **794 total** workspace tests passing, 0 failures
- `cargo build --workspace && cargo test --workspace && cargo clippy --workspace --all-targets` ✅

### Remaining before merge

- [x] `/super-review` — multi-model PR review (Codex + Gemini)
- [x] `/finish-pr` — pre-merge validation checklist

Blocked on Codex/Gemini usage limit reset. Will resume later.

src/engine/activity_filter.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    ActivityFilterDecision::Record
+}
+
+fn format_promote_content(tool_name: &str, result: Option<&str>) -> String {
+    match result {
+        Some(r) if !r.is_empty() => {
+            let truncated = if r.len() > 200 { &r[..200] } else { r };
```
dutiona: [Codex Review] [HIGH] `format_promote_content()` slices `result` by raw byte index (`&r[..200]`). A valid UTF-8 summary longer than 200 bytes can panic here if byte 200 lands mid-codepoint, which turns `memory_record_activity` into a crash on perfectly valid tool output. Please reuse the UTF-8-safe truncation logic from `activity.rs` (or a shared helper) before formatting promoted fact content.

src/store/activities.rs
```diff
@@ -0,0 +1,398 @@
...
+        dedup_window_secs: i64,
+    ) -> Result<(i64, bool)> {
+        let ts = activity.timestamp.to_rfc3339();
+
+        // Check for existing activity within the dedup window.
+        let existing: Option<(i64, i64)> = self
+            .conn
+            .query_row(
+                "SELECT id, occurrence_count FROM activities
+                 WHERE session_id = ?1
```
dutiona, line 38: [Codex Review] [MEDIUM] `scope_id` is not part of this dedup lookup. Two identical tool calls in the same session but different scopes will collapse into one row, and `load_context()` for the second scope will never see its own activity. Because context loading is scope-filtered, dedup should be scope-aware as well (query plus supporting index).

src/store/activities.rs
```diff
@@ -0,0 +1,398 @@
...
+                   AND tool_name = ?2
+                   AND args_hash = ?3
+                   AND outcome_class = ?4
+                   AND status IN ('recorded', 'deduplicated')
```
dutiona: [Codex Review] [HIGH] This dedup query excludes rows once they have been promoted (`status = promoted`). A second identical `git_commit`/`git_push` inside the dedup window will therefore miss the existing row, insert a fresh activity, and promote a second duplicate fact. Dedup needs to consider promoted rows too, otherwise the "deduplicate before promotion" contract stops working as soon as the first insert is promoted.

memory-engine-mcp/src/tools/mod.rs
```diff
@@ -1250,3 +1301,86 @@ fn handle_outcome_counts(
...
+    ok_json(json!({
+        "session_id": session_id,
+        "checkpointed": true,
+    }))
+}
+
+fn handle_load_context(
+    args: Map<String, Value>,
+    engine: &MemoryEngine,
+) -> Result<CallToolResult, ErrorData> {
```
dutiona, line 1374: [Codex Review] [MEDIUM] These limits cast raw `i64` values straight to `usize`. A caller can send `-1`, which wraps to a huge `usize` and effectively removes the bound on activities/facts returned. There is already a `get_usize()` helper above that rejects negatives; this path should use it (or explicit validation) instead of `as usize`.

dutiona (commented):
[Codex Review] I found 4 issues in this patch.

- [HIGH] `format_promote_content()` can panic on valid UTF-8 result summaries because it truncates by raw bytes.
- [HIGH] Promoted activities are excluded from the dedup lookup, so repeated promoted actions can create duplicate activity rows and duplicate promoted facts.
- [MEDIUM] The dedup key is not scope-aware, so identical actions in different scopes can collapse together and disappear from the second scope's context.
- [MEDIUM] `memory_load_context` casts signed limits directly to `usize`, so negative inputs wrap to huge values and can remove the intended response bounds.

I left inline comments on each point.

src/engine/activity_filter.rs
dutiona: Fixed: replaced `&r[..200]` with `&r[..r.floor_char_boundary(200)]` which finds the last valid UTF-8 boundary at or before byte 200. Stable since Rust 1.80, our MSRV is 1.85.

src/store/activities.rs
dutiona: Fixed: added `AND scope_id = ?5` to the dedup lookup query. Activities in different scopes are now distinct dedup candidates.

src/store/activities.rs
dutiona: Fixed: added 'promoted' to the IN clause — `status IN ('recorded', 'deduplicated', 'promoted')`. Promoted activities now participate in dedup, preventing duplicate fact creation.

memory-engine-mcp/src/tools/mod.rs
dutiona: Fixed: swapped `get_i64(...) as usize` to `get_usize(...)` which already uses `usize::try_from().ok()` to reject negatives. Returns `None` on negative input, falling through to the default.

src/engine/activity.rs
```diff
@@ -0,0 +1,250 @@
...
+        // Step 3: Resolve scope.
+        let scope_id = match req.scope_path.as_deref() {
+            Some(path) => {
+                let conn = self.write_conn()?;
+                self.ensure_scope_with_conn(&conn, path)?
+            }
+            None => 1, // root scope
+        };
+
+        // Step 4: Insert or dedup.
```
dutiona: [Gemini Review] [HIGH] Multiple write lock acquisitions in `record_activity`. The method calls `self.write_conn()?` three times (lines 83, 102, 131). This is inefficient and non-atomic. Consider acquiring the write lock once and passing the connection to internal methods or re-using it.

src/store/activities.rs
```diff
@@ -0,0 +1,398 @@
...
+        if scope_ids.is_empty() {
+            return Ok(Vec::new());
+        }
+        let placeholders: String = scope_ids.iter().map(|_| "?").collect::<Vec<_>>().join(",");
+        let sql = format!(
+            "SELECT id, session_id, tool_name, args_hash, args, result_summary,
+                    outcome_class, status, occurrence_count, first_seen, last_seen,
+                    scope_id, promoted_fact_id
+             FROM activities
+             WHERE scope_id IN ({placeholders})
```
dutiona: [Gemini Review] [MEDIUM] `list_recent_by_scope` uses manual placeholder generation and `params_from_iter` (line 158). This is inconsistent with the rest of the project (e.g., `FactStore::list_by_scopes_recent`) and may hit SQLite parameter limits for large scope trees. Recommendation: Use `json_each(?1)` with a serialized JSON array of IDs.

src/engine/activity_filter.rs
```diff
@@ -0,0 +1,205 @@
...
+            format!("[{tool_name}] {truncated}")
+        }
+        _ => format!("[{tool_name}] (no result summary)"),
+    }
+}
+
```
dutiona, line 109: [Gemini Review] [MEDIUM] Panic risk in `format_promote_content`: `&r[..200]` (line 103) will panic if index 200 falls in the middle of a multi-byte UTF-8 character. Recommendation: Use the `truncate` helper from `src/engine/activity.rs` (consider moving it to a common utility location) or check `is_char_boundary` before slicing.

src/store/schema.rs
```diff
@@ -661,6 +662,49 @@ fn migrate_v7_to_v8(conn: &Connection) -> Result<()> {
...
+            result_summary   TEXT,
+            outcome_class    TEXT    NOT NULL DEFAULT 'success',
+            status           TEXT    NOT NULL DEFAULT 'recorded'
+                             CHECK(status IN ('recorded', 'deduplicated', 'ignored', 'promoted')),
+            occurrence_count INTEGER NOT NULL DEFAULT 1,
+            first_seen       TEXT    NOT NULL,
+            last_seen        TEXT    NOT NULL,
+            scope_id         INTEGER NOT NULL DEFAULT 1 REFERENCES scopes(id),
+            promoted_fact_id INTEGER REFERENCES facts(id)
+        );
```
dutiona, line 682: [Gemini Review] [LOW] Missing index for `checkpoint_session`. The query `SELECT id FROM activities WHERE session_id = ?1 ORDER BY last_seen DESC LIMIT 1` (used in `MemoryEngine::checkpoint_session`) would benefit from a composite index on `(session_id, last_seen DESC)`. Consider updating the `idx_activities_session` index or adding a new one.

dutiona (commented):
[Gemini Review] Summary of findings:
1. [HIGH] `record_activity` acquires write locks multiple times, which is inefficient.
2. [MEDIUM] `list_recent_by_scope` uses manual placeholders instead of `json_each`, risking errors with large scope trees.
3. [MEDIUM] Potential panic in `format_promote_content` due to unsafe UTF-8 string slicing.
4. [LOW] Missing composite index for session checkpoint activity lookup.

Overall, the feature is well-implemented and follows the project patterns, but these technical debt/correctness issues should be addressed.

src/engine/activity_filter.rs
dutiona: Already addressed in commit 0ff0b2c — replaced with `r.floor_char_boundary(200)`. Good catch confirming the Codex finding.

src/engine/activity.rs
dutiona: Partially addressed in 15a2f19: consolidated scope resolution + insert into one lock acquisition (2 calls → 1). The third call (line 125, post-embedding update) stays separate by design — add_fact does embedding, which is expensive network I/O that must not hold the write lock. The non-atomicity between insert and promote is acceptable: promotion failure already falls back gracefully to Recorded status.

src/store/activities.rs
dutiona: Fixed in 15a2f19: switched to json_each(?1) pattern, matching FactStore and EdgeStore. Also parametrized the LIMIT clause.

src/store/schema.rs
dutiona: The existing idx_activities_session already covers session_id. Added scope_id to idx_activities_dedup instead (15a2f19) — more impactful since the dedup query now uses scope_id. The checkpoint query perf is adequate with the existing index for typical session sizes.

src/store/activities.rs
```diff
@@ -0,0 +1,398 @@
...
+    let status_str: String = row.get(7)?;
+    let status = status_str.parse::<ActivityStatus>().map_err(|e| {
+        rusqlite::Error::FromSqlConversionFailure(
+            7,
+            rusqlite::types::Type::Text,
+            Box::new(std::io::Error::new(std::io::ErrorKind::InvalidData, e)),
+        )
+    })?;
+    let args_str: String = row.get(4)?;
+    let args: serde_json::Value = serde_json::from_str(&args_str).unwrap_or_default();
```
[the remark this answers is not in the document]
dutiona: The `status` field already returns an explicit error via `parse::<ActivityStatus>().map_err(...)` (line 210-216). The `args` field uses `unwrap_or_default()` on line 218, but the DB has `CHECK(json_valid(args))` so malformed JSON cannot exist in the column. The CHECK constraint is the boundary guard — adding a redundant parse error path for an impossible condition adds complexity without value.

Issue #516, opened 2026-06-13
chore(build): clippy -D warnings debt in async/ann feature modules


dutiona:
## Summary

`cargo clippy --all-targets -- -D warnings` is clean under **default features** (fixed in #515). But ~40 **pre-existing** lints remain in the feature-gated modules `src/async_engine.rs` (`async` feature) and `src/search/ann.rs` (`ann` feature). These surface only under:

- `cargo clippy --all-features --all-targets -- -D warnings`, and
- `cargo clippy --workspace --all-targets -- -D warnings` (the `memory-engine-mcp` crate enables `features = ["async"]`, so the workspace gate compiles `async_engine.rs`).

They predate #515 and are outside its scope (gap #2 was the default-features root-crate gate). `cargo test --all-features` **compiles and runs green** — this is lint hygiene only, not a build break.

## Breakdown (~40)

**`src/async_engine.rs` (34):**
- `missing_errors_doc` ×26 — async wrappers (`pub async fn` / `spawn_blocking`) mirroring sync engine methods, missing `# Errors` sections.
- `doc_markdown` ×4 — un-backticked identifiers in doc comments.
- `needless_pass_by_value` ×1 (`:41`), `missing_const_for_fn` ×1 (`:56`).

**`src/search/ann.rs` (6):**
- `doc_markdown` ×2, `unnecessary_literal_bound` ×1 (`HnswStrategy::name` `-> &str` → `&'static str`), `significant_drop_tightening` ×1 (`:245`), `unused_self` ×1 (`:146`).

## Acceptance

`cargo clippy --all-features --all-targets -- -D warnings` and `cargo clippy --workspace --all-targets -- -D warnings` both exit 0.

## Notes

Most are mechanical (`# Errors` docs + backticks). The same patterns and scoped-`#[allow]` conventions established in #515's PR apply (e.g. the `ann` `significant_drop_tightening` is likely the same store-wrapper-borrow false positive).

_Collateral finding from work on #515._

Issue #291, opened 2026-06-01
[super-qa] high: cargo test --all-features broken via #1


dutiona:
> ✅ **Verified against primary source.** VERIFIED

**cargo test --all-features broken via #1**

- **ID:** `workspace/test-allfeatures-broken`
- **Severity:** high
- **Area:** build · **Type:** chore
- **Location:** `workspace`
- **Auto-fixable:** no

## Description
(none)

## Suggested approach
—

## References
—

dutiona:
Resolved by #517 (consolidated under #515). `cargo test --all-features` now compiles and runs green.

Issue #290, opened 2026-06-01
[super-qa] high: clippy -D warnings gate fails (default)


dutiona:
> ✅ **Verified against primary source.** VERIFIED exit101

**clippy -D warnings gate fails (default)**

- **ID:** `workspace/clippy-gate-broken`
- **Severity:** high
- **Area:** build · **Type:** chore
- **Location:** `workspace`
- **Auto-fixable:** no

## Description
(none)

## Suggested approach
—

## References
—

dutiona:
Resolved by #517 (consolidated under #515). `cargo clippy --all-targets -- -D warnings` now passes.

Issue #95, opened 2026-03-20
MCP server: consolidate, forget, dump_state, pin/unpin tools (P1)


dutiona:
### Scope

Additional MCP tools for memory management operations, deferred from the P0 MVP.

### Tools

| MCP Tool | Engine Method | Notes |
|----------|--------------|-------|
| `consolidate` | `consolidate()` | Requires `SummaryGenerator` trait configured at server startup. Tool unavailable if not configured — returns clear error. |
| `forget` | `forget()` | Takes `ForgetPolicy` parameters (half_life_days, min_importance, weights). |
| `dump_state` | `dump_state()` | Export engine state as JSON/SQLite/compressed. Returns file path. |
| `pin_fact` | `pin_fact()` | Make a fact unforgettable. |
| `unpin_fact` | `unpin_fact()` | Remove unforgettable flag. |

### Design notes

- `consolidate` is the only tool that requires a non-trivial trait (`SummaryGenerator`). The MCP server should report its availability via server capabilities/metadata so agents know whether consolidation is possible.
- `forget` policy parameters should have sensible defaults matching the engine's `ForgetPolicy::default()`.
- `dump_state` format selection via parameter: `format: "json" | "json_gzip" | "json_zstd" | "sqlite"`.
- `pin_fact` / `unpin_fact` are trivial pass-through — just fact ID.

### Dependencies

- #45 (MCP server P0 must land first)

PR #517, opened 2026-06-13
refactor(build): fix --all-features compile, clear clippy -D warnings, fix fmt drift (#515)

### LINKED ISSUE #515: build: --all-features does not compile; clippy --all-targets -D warnings has 195 errors; fmt drift in 2 files
Three pre-existing hygiene gaps, all verified on untouched `main` (2bab8ff) during PR #514 verification, 2026-06-12:

## 1. `--all-features` does not compile

```
cargo test --all-features
# error: duplicate definitions — `verify_archives` defined twice
# candidate #1 / #2 both in an impl for `engine::MemoryEngine` at src/engine/archive.rs:116
```

Two cfg-gated impls of the same method collide when mutually-exclusive features are enabled together. Goal: `cargo test --all-features` compiles **and runs green**, so the full feature matrix is CI-coverable.

## 2. `clippy --all-targets -- -D warnings`: 195 errors

```
cargo clippy --all-targets -- -D warnings
# error: could not compile `memory-engine` (lib test) due to 195 previous errors
```

Identical count on main and on PR #514's branch (the PR adds zero new lints). Presumably CI uses a narrower invocation; goal is the strict one passing.

## 3. `cargo fmt --check`: drift in 2 files

`src/engine/activity.rs`, `src/store/activities.rs` — import-ordering drift (looks like rustfmt style-edition difference).

---
Found during #514 (decay exemption); excluded from that PR to keep it atomic. Owner will address in a dedicated session.

dutiona:
Fixes #515. Also resolves the duplicate super-qa findings #290 (default `clippy -D` fails) and #291 (`cargo test --all-features` broken).

Closes the three build-hygiene gaps so the strict CI gates pass. Three atomic commits, one per gap.

## Commits

1. **`style(build)`** — gap #3: `cargo fmt --check`. Normalized rustfmt import-ordering drift (stable rustfmt 1.9 / the CI toolchain).
2. **`fix(build)`** — gap #1: `cargo build/test --all-features`. The root cause was **not** the duplicate `verify_archives` the issue guessed, but a **duplicated `#[cfg(feature = "archive")] mod archive;`** in `src/engine/mod.rs` (same feature gate) → 13 cascading duplicate-definition errors. Plus a second latent error: `cognitive.rs` called `hnsw.notify_insert(..)` without `VectorSearchStrategy` in scope.
3. **`refactor(build)`** — gap #2: `cargo clippy --all-targets -- -D warnings`. ~195 default-feature lints + a deny-level `explicit_counter_loop` + an MSRV-incompatible API, then a further batch in `examples/`/`tests/eval/` that only surfaced once the deny-level errors were cleared.

## Verification (all green)

| Gate | Result |
|------|--------|
| `cargo fmt --check` | ✅ |
| `cargo clippy --all-targets -- -D warnings` (default) | ✅ exit 0 |
| `cargo test` | ✅ 537+ pass |
| `cargo test --all-features` | ✅ compiles + runs green |
| `cargo build --workspace` | ✅ |

## Please review these judgment calls

- **Scoped `#[allow]` (genuine false positives only):** `significant_drop_tightening` ×6 — the write `conn` is borrowed through a store wrapper (`FactStore`/`ScopeStore`/`EventStore`/`CheckpointStore`) across the whole atomic op; clippy misses the transitive borrow and its own machine suggestion **would not compile**. Plus `missing_const_for_fn` on `active_strategy_name` (const under default features, non-const under `ann` — can't be const across the matrix). Each has a `// reason` comment.
- **`dead_code` → `#[cfg(test)]`:** the store methods `get`/`list_recent`/`list_by_session`/`count_by_session` are exercised only by unit tests (not yet wired into the engine). Gated `#[cfg(test)]` with a note pointing at the MCP P1/P2 tools (#95/#96). **If these are intended as live API, say so and I'll wire/expose them instead.**
- **`struct_excessive_bools`:** `OutcomeSignals.tests_passed: bool` → `TestOutcome` enum (one bool → enum is enough to clear the >3-bools threshold; also lets it distinguish "failed" from "not run").
- **`too_many_lines`:** extracted helpers from `add_facts_batch`, `bootstrap_within_savepoint`, `migrate_v2_to_v3`, `restore_snapshot_into` (behavior-preserving). The 2 example/test long fns kept linear with a scoped allow.

## Collateral (not in this PR)

`#516` — ~40 **pre-existing** clippy lints in the `async`/`ann` feature modules (`async_engine.rs`, `ann.rs`), surfaced only by `--all-features`/`--workspace` clippy. Out of #515's default-features scope; `--all-features` **test** is green. Filed separately per the repo's collateral-issues policy.


dutiona (commented):
[Gemini Review] LGTM. Audited extracted helpers (ingest.rs, bootstrap/mod.rs, schema.rs, restore.rs) and verified behavior parity. Drop-tightening in mod.rs/activity.rs preserves transaction boundaries while improving concurrency. Feature-matrix fixes confirmed.

dutiona (commented):
[Antigravity Review] LGTM! Reviewed all extracted helpers, transaction lock boundaries, cast refactorings, and clippy hygiene adjustments. All tests compile and pass on --all-features.

src/bootstrap/outcome.rs
```diff
@@ -22,14 +22,27 @@ pub enum SessionOutcome {
...
+///
+/// An enum (not a `bool`) so "no passing-test signal" is not conflated with a
+/// definite failure — and to keep [`OutcomeSignals`] within clippy's bool budget.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum TestOutcome {
+    /// A passing-test signal (`test result: ok` / `passed`) was detected.
+    Passed,
+    /// No passing-test signal detected (tests failed or did not run).
+    #[default]
+    Failed,
```
dutiona: [Codex Review] [MEDIUM] `Failed` is assigned whenever no passing-test signal is detected, including when tests never ran. That contradicts the enum rationale above and turns the old `false` meaning ("not observed passing") into a definite failure signal for downstream consumers. Either use a neutral variant such as `NotObservedPassing`, or add a third `NotRun`/`Unknown` state and only emit `Failed` when an actual failure is detected; `classify_outcome` can continue using the original local bool so its classification remains unchanged.

src/bootstrap/metrics.rs
```diff
@@ -62,10 +62,16 @@ impl BootstrapReport {
         let total_new = other.prewarm_metrics.total_count();
         let total = total_old + total_new;
         if total > 0 {
-            self.prewarm_metrics.avg_importance = (self.prewarm_metrics.avg_importance
-                * total_old as f64
-                + other.prewarm_metrics.avg_importance * total_new as f64)
-                / total as f64;
+            // Counts are small (episode tallies); convert losslessly via u32 so
+            // clippy's cast_precision_loss does not fire (f64::from is lossless).
+            let w_new = f64::from(u32::try_from(total_new).unwrap_or(u32::MAX));
```
dutiona: [Codex Review] [LOW] These `unwrap_or(u32::MAX)` fallbacks silently saturate independently, so `merge` no longer computes the same weighted average once any public count exceeds `u32::MAX`. For example, an old count of `u32::MAX + 1` and a new count of 1 use weights `u32::MAX` and 1 but a denominator also clamped to `u32::MAX`, yielding a value outside the true weighted mean. The preceding comment also calls the conversion lossless, which the fallback makes false. Keep the original `usize` ratio under a narrowly scoped cast allow, or convert through a wider numeric representation without saturation.

dutiona (commented):
[Codex Review] Found 2 correctness issues: [MEDIUM] `TestOutcome::Failed` misrepresents sessions where tests did not run, despite the stated non-conflation goal; [LOW] the new `u32::MAX` saturation changes weighted-average behavior for large public report counts. The four helper extractions preserve ordering/error propagation/savepoint ownership; the explicit drops do not move transaction boundaries; FTS/vector serialization errors now propagate correctly; dead-code gates have no non-test callers; and the feature-gated module/import fixes are correct. Verification: `cargo fmt --check`, `cargo build --workspace`, `cargo test --workspace`, and `cargo test --workspace --all-features` passed. `cargo clippy --workspace --all-targets -- -D warnings` failed locally under Clippy 1.96 on 31 pre-existing async-wrapper diagnostics, so I did not treat those toolchain-version-dependent style warnings as findings under this review brief.

dutiona:
## Super-review round 1 — resolutions

Three-model review (Codex + Gemini + agy). **No blocker/high findings** — all three confirmed the helper extractions are semantically identical, the drop reorders preserve transaction boundaries, and the `--all-features` fixes are correct.

**Addressed in 36d08c4:**
- **[MEDIUM] Codex — `TestOutcome::Failed` semantics:** renamed `Failed` → `NotObserved`. The heuristic only detects a *passing* signal, so its absence isn't a definite failure (the old name contradicted the enum's own doc and overclaimed vs the original `bool`'s `false`).
- **[LOW] Codex + Gemini — saturating metrics cast:** restored the direct `usize as f64` under a scoped `cast_precision_loss` allow instead of `try_from(..).unwrap_or(u32::MAX)` (episode tallies are tiny → lossless in practice; the saturation was "not lossless as claimed").

**Verified non-issues (no change):**
- **[MEDIUM] Gemini — "missing AFTER UPDATE FTS trigger":** `facts_fts_au` **is** present in `recreate_facts_fts_v3` (`schema.rs:463`) — all three triggers (ai/ad/au) were extracted.
- **Codex — `clippy --workspace` async-wrapper failures:** pre-existing `async`/`ann` debt, unrelated to this PR; tracked separately in #516.

Re-verified green: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test`. Soft-exiting the review loop — only resolved/invalid findings remained.

Issue #51, opened 2026-03-12
Phase 6: Knowledge change notification — KB updates trigger memory re-evaluation


dutiona:
### Scope

When KB content is superseded or updated, notify memory-engine to re-evaluate facts that reference it.

**Scenarios:**
- Paper retracted → facts citing it should be flagged for review
- Paper superseded → facts may need conflict resolution with newer findings
- KB entry updated → dependent summaries may be stale

**API sketch:**
```rust
pub trait KnowledgeBaseConnector: Send + Sync {
    // ... existing methods ...
    fn on_change(&self, uri: &str, event: KBChangeEvent) -> Result<(), KBError>;
}
```

**Engine behavior:**
- Find all facts with matching `knowledge_ref`
- Mark them for re-evaluation (flag, not auto-modify)
- Consumer decides what to do (re-consolidate, expire, update)

### Research reference

Four-layer architecture: partial gap — Knowledge ↔ Memory bridge missing bidirectional notification.

### Dependencies

- Phase 6 KnowledgeBaseConnector (#50)

dutiona:
**Gap analysis update (notes 23–28)**

Expand scope beyond KB→ME notification. Note 27 identifies the reverse direction too: when ME appends a fact, scope-subscribed agents should receive notification (stigmergy pattern). This pairs with new issue #164 (pub/sub event emission on fact append).

Consider whether #51 and #164 should be merged or kept as complementary KB→ME and ME→agents directions.

Issue #45, opened 2026-03-12
Phase 4b: MCP server — memory-engine-mcp


dutiona:
### Scope

New workspace binary: `memory-engine-mcp`. Maps 1:1 to engine API.

**MCP tools exposed:**
- `ingest`, `add_fact`, `query`, `resume_context`, `list_due`
- `consolidate`, `forget`, `explain_fact`
- `dump_state`, `statistics`

**Pre-compaction flush endpoint:** Push-based capture — receives model's last-minute insights before context window compaction. Addresses the Intelligence → Memory partial gap from the four-layer architecture.

**Versioning:** Explicit supported MCP version range in server metadata. Reject incompatible clients with clear error.

**Deployment:** Initially in-tree as workspace member. Eventually separate repo once Rust public API stabilizes.

### Design reference

See `docs/design/plans/2026-03-09-future-phases-design.md` § Phase 4 → MCP Server Adapter.

### Research reference

Four-layer architecture § Pre-Compaction Hook, § Surgeon vs. Scrub Nurse.

### Dependencies

- Phase 4a inspection APIs (#39)
- Stable engine API (all Phase 4a items)

dutiona:
## Idea: Tiered retrieval depth in MCP responses

**Source:** NeuroStack analysis (research note in autonomous-agent-project, 2026-03-14)

NeuroStack implements an auto-escalating tiered retrieval pattern for its MCP search tool that's worth stealing for our MCP server:

- **Depth 1 — sparse:** Return structured fact summaries only (~15 tokens/fact). Maps to our `explain_fact()` one-liner or a compact query result.
- **Depth 2 — medium:** Return fact + provenance summary (~75 tokens). Enough for the agent to decide if it needs more.
- **Depth 3 — full:** Return full provenance chain, event history, related facts (~300+ tokens).

Their `depth` parameter defaults to `"auto"` which starts sparse and escalates only if coverage is insufficient (< 2 unique sources or confidence < 0.4).

### Why this matters for us

Our MCP server will serve agents operating under context window pressure. Returning full provenance chains for every query wastes tokens when the agent just needs a quick "do I know anything about X?" check. The tiered pattern lets agents ask cheaply first, then drill down.

### Suggested approach

Add an optional `depth: "sparse" | "standard" | "full"` parameter to the `query` MCP tool (default `"standard"`):

- **sparse**: fact content + importance score + scope. No provenance, no edges, no event history.
[...]

PR #519, opened 2026-06-14
S0.5 (#45): secrets/PII redaction gate for session import

dutiona:
## S0.5 (#45) — Secrets/PII redaction gate

A verified, **dependency-free** redaction gate that sits in front of any bulk import of session `.jsonl` into ME/KB — a HARD prerequisite for the B3 backfill, reviewed by the owner (#51) before backfill. **Build-only: not wired into `bootstrap_session` yet.**

### What
`src/bootstrap/redact.rs` — `redact_text` / `redact_entries`. Over whitespace-delimited tokens:
- **provider-token prefixes** (gitleaks-style): AWS `AKIA…`, GitHub `ghp_`/`github_pat_`, OpenAI `sk-`, Google `AIza`, Slack `xox[bp]-`, `-----BEGIN…PRIVATE KEY`
- **email** PII
- **high Shannon entropy** (len ≥ 20, secret-like charset, ≥ 4.0 bits/char) — catches arbitrary keys the prefixes miss

Findings → `[REDACTED:<rule>]` in place (non-secret text + whitespace preserved exactly); `redact_entries` returns a `RedactionReport` whose `coverage_lines()` is the **auditable coverage report** for #51. Idempotent.

### Scope (v1)
prefix + email + entropy. #51 may upgrade to full regex/gitleaks rules (`key=value` assignments, JWT structure) and tune the entropy threshold against the real corpus — see `docs/audits/S0.5-redaction-gate.md`. Deliberately not wired into the import path until ratified.

### Gates
9 unit tests (`cargo test --lib bootstrap::redact`); clippy `-D warnings` + nightly `cargo fmt --check` clean. No engine behavior changed.

Addresses dutiona/autonomous-agent-project#45.

### DIFF
--- a/docs/audits/S0.5-redaction-gate.md
+++ b/docs/audits/S0.5-redaction-gate.md
@@ -11,45 +11,111 @@ and is a HARD prerequisite — non-negotiable, reviewed by the owner (#51) befor
 ## What it does
 
 `redact_text(text) -> (redacted, findings)` and `redact_entries(entries) -> (redacted, RedactionReport)`.
-Detection is **dependency-free**. Tokens are split on whitespace **and** JSON/assignment
-punctuation (`" ' { } [ ] ( ) : , ; =`) so a secret embedded in `"token":"ghp_…"` or
-`key=secret` is isolated — `.`/`@`/`/`/`-`/`_`/`+` are kept (they live inside emails, JWTs,
-base64, provider tokens). Each token is classified:
+Detection is **dependency-free**. Rather than splitting on a blanket delimiter set, the gate
+runs a **multi-detector span scanner**: each detector finds redaction spans over the raw text,
+non-overlapping spans are collected in priority order (earlier detectors win over the same
+character range), and the string is rebuilt replacing each span with `[REDACTED:<rule>]`.
+All indices are char-based to avoid panics on multibyte text. Non-secret characters are
+preserved byte-exact.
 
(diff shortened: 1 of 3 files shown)


dutiona:
The token-based redaction these comments reviewed has been **replaced** by a dependency-free 5-detector span scanner (`2dbcce7`), so:

- **Punctuation handling (Gemini, security-high) & "secrets embedded in JSON" (Codex, P1):** addressed by the rewrite. Detectors now find provider-prefixes at a token boundary (incl. after `"`/`:`/`/`), URL-credential passwords, emails, context-gated hex, and high-entropy tokens — over the raw text, so `{"OPENAI_API_KEY":"sk-..."}` and URL/dotted/angle-bracket-embedded secrets are caught. Covered by `redacts_secrets_embedded_in_json` + the new fix tests. Resolving.
- **`shannon_entropy` HashMap→array & `looks_like_secret_charset` single-pass (Gemini, perf):** valid micro-optimizations, but these run on short (≤~70-char) token candidates, so the gain is negligible; the `[usize;128]` entropy variant would also need a non-ASCII fallback (the fn is `pub` and `multibyte_tokens_no_panic` exercises non-ASCII). Keeping the security rewrite focused; not changing now. Resolving as acknowledged.

38 unit tests, `cargo +nightly fmt --check` + `cargo clippy --all-targets` clean.

Issue #506, opened 2026-06-01
[super-qa] Auto-fix: low (114)


dutiona:
> Mechanically auto-fixable low findings (114). Apply via `/super-qa` Phase 3.

| ID | Location | Title | Suggested fix |
| --- | --- | --- | --- |
| `bootstrap/correctness-silent-rollback-discard` | `src/bootstrap/mod.rs:111-112` | ROLLBACK TO and RELEASE errors silently discarded in savepoint error p | Log rollback failures at warn level: ```rust if le |
| `bootstrap/correctness-test-assertion-too-weak` | `src/bootstrap/extract.rs:224` | content_truncation test assertion allows 3 extra bytes, contradicting  | Replace: ```rust assert!(facts[0].content.len() <= |
| `cli/correctness-truncate-str-underflow` | `memory-engine-cli/src/output.rs:17-23` | truncate_str produces output longer than max_chars when max_chars < 3 | Cap the suffix to `max_chars` and use an early-exi |
| `embed/correctness-embed-missing-array-guard` | `memory-engine-embed/src/http.rs:145-158` | `embed()` format detection does not guard that `data` is an array, so  | Change line 145 to use the same guard as embed_bat |
| `embed/correctness-u64-usize-narrowing` | `memory-engine-embed/src/http.rs:60-62` | Silent narrowing cast `u64 as usize` can corrupt index validation on 3 | Replace the `as usize` cast with a checked convers |
| `pool/correctness-doc-claims-query-only-pragma-not-set` | `src/pool/connection_pool.rs:143-144` | Doc comment incorrectly claims PRAGMA query_only=ON is set for read-on | Replace the stale claim in the doc comment:  ```ru |
| `store/correctness-lineage-sql-interpolation` | `src/store/lineage.rs:44-46` | Fact-ID validation query uses string interpolation instead of paramete | Replace with the json_each pattern: ```rust let id |
| `cli/design-matchtype-non-exhaustive-wildcard` | `memory-engine-cli/src/commands/query.rs:124-130` | Non-exhaustive `MatchType` match silently renders `Archive` variant as | MatchType::Fts => "FTS".into(), MatchType::Vector  |
| `conflict/design-arbiter-importance-score-hazard` | `src/conflict/temporal.rs:65, src/traits.rs:60-67` | ConflictArbiter doc does not warn that importance_score on the new Fac | // traits.rs: ConflictArbiter::arbitrate doc block |
[...]

PR #527, opened 2026-06-15
fix(core): snap embed_batch error truncation to a UTF-8 char boundary

### LINKED ISSUE #526: fix(core): embed_batch error truncation panics on multibyte UTF-8 body
## Summary

`HttpEmbeddingProvider::embed_batch` builds an error message for an
unrecognized response body by slicing at a fixed **byte** index:

`memory-engine-embed/src/http.rs:222-223`
```rust
let truncated = if body_str.len() > 1000 {
    format!("{}... (truncated)", &body_str[..1000])
```

`body_str` is `serde_json::to_string(&body)`, and serde_json passes non-ASCII
through as raw UTF-8 (it does not `\u`-escape it). If a multibyte codepoint
straddles byte 1000, `&body_str[..1000]` is **not** on a char boundary and
**panics**: `byte index 1000 is not a char boundary`.

## Trigger

The embedding server returns a response in none of the recognized formats
(no `data`, `embeddings`, or `embedding` key) AND the serialized body exceeds
1000 bytes with a multibyte codepoint near byte 1000 — e.g. a non-English
error page, or a body containing emoji / CJK text.

## Impact

A panic in the embedding path instead of a clean `MemoryError::Internal`.
Low severity (error path, needs a specific body shape), but a panic in a
library is never acceptable.

## Fix

Snap the cut down to the nearest UTF-8 char boundary before slicing
(`str::is_char_boundary`), extracted into a small unit-tested helper. PR to
follow with a regression test (4-byte codepoint straddling the cut).

## Provenance

Found off-script during the #506 wave-1 auto-fix work — a parallel agent
proposed this fix, but it is outside #506's scope, so it is filed separately
per the collateral-issue convention in CLAUDE.md.

dutiona:
## Fix the multibyte-UTF-8 truncation panic in `embed_batch`

Closes #526.

`HttpEmbeddingProvider::embed_batch` built its "unrecognized response" error
message by slicing the serialized body at a fixed **byte** index:

```rust
let truncated = format!("{}... (truncated)", &body_str[..1000]);
```

`serde_json::to_string` passes non-ASCII through as raw UTF-8, so when a
multibyte codepoint straddles byte 1000 the slice is not on a char boundary
and **panics** (`byte index 1000 is not a char boundary`).

### Change

- Extract `truncate_on_char_boundary(s, max)` — snaps the cut **down** to the
  nearest UTF-8 boundary (byte 0 is always a boundary, so it terminates).
- Route the `embed_batch` error path through it.
- Regression test: a 4-byte codepoint placed at bytes 998–1001 so byte 1000
  lands mid-codepoint (panics on the old code, snaps to 998 on the new) + a
  short-string passthrough test.

### Verification

- `cargo test -p memory-engine-embed` — 7 passed, incl. the 2 new tests
- `cargo clippy --workspace --all-targets` — 0 deny
- `cargo fmt` — clean

Found off-script during #506 wave-1 work; filed and fixed separately per the
collateral-issue convention.

### DIFF
--- a/memory-engine-embed/src/http.rs
+++ b/memory-engine-embed/src/http.rs
@@ -219,11 +219,7 @@ impl EmbeddingProvider for HttpEmbeddingProvider {
             Self::parse_ollama_batch(arr, texts.len())?
         } else {
             let body_str = serde_json::to_string(&body).unwrap_or_default();
-            let truncated = if body_str.len() > 1000 {
-                format!("{}... (truncated)", &body_str[..1000])
-            } else {
-                body_str
-            };
+            let truncated = truncate_on_char_boundary(&body_str, 1000);
             return Err(MemoryError::Internal(format!(
                 "cannot extract embeddings from batch response: {truncated}"
             )));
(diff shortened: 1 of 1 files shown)


dutiona:
Superseded by #525 (fix/correctness-504), which already carries the identical char-boundary truncation fix. I ported this PR's regression test onto #525 (commit 53bac7b) so the fix there is now covered. Closing as duplicate.

PR #540, opened 2026-06-15
fix(storage): reject non-file backup paths in restore_sqlite (#191)

### LINKED ISSUE #191: fix: restore_sqlite should use is_file() instead of exists()
## Bug

In `src/engine/restore.rs`, `restore_sqlite` checks `backup_path.exists()` before copying. If a directory path is passed, `exists()` returns `true` but `std::fs::copy` will fail with a confusing OS-level error.

## Fix

Replace `backup_path.exists()` with `backup_path.is_file()` for a more precise check and a clearer error message when a directory is passed.

## Found by

Gemini Code Assist bot during PR #186 review. Pre-existing in the original `engine.rs`.

## Severity

Low — edge case, unlikely in practice (callers pass paths from `dump_state` output).

dutiona:
## #191 — `restore_sqlite` should reject non-file backup paths

`restore_sqlite` guarded the backup **source** with `!backup_path.exists()`. A directory path passes that guard and then fails deep inside `std::fs::copy` with an opaque OS-level error. Switched to `is_file()` for a precise up-front rejection.

### Scope

- `src/engine/restore.rs` — `restore_sqlite` source guard `exists()` → `is_file()` + clearer "not a regular file" message. The two `config.path.exists()` **target** checks are a different, correct semantic and were intentionally left untouched.
- `tests/restore_roundtrip_test.rs` — `restore_sqlite_fails_if_backup_is_directory` regression test (asserts a precise `NotFound` "not a regular file" error, _not_ a copy error, and no orphan target file).

### Gates

- `cargo build --workspace` ✓
- `cargo test -p memory-engine` ✓ (restore suite green, incl. existing `..._backup_missing`)
- clippy — project gate (`clippy::all = deny`) ✓; restore.rs adds **zero** findings
- fmt ✓ on changed files

> Pre-existing `clippy -D warnings` debt and rustfmt import-order drift on base `63ee3c6` are **out of scope** (filed separately) — this change contributes zero new findings, verified via `git stash` on the clean base.

Fixes #191

### DIFF
--- a/src/engine/restore.rs
+++ b/src/engine/restore.rs
@@ -106,7 +106,8 @@ impl MemoryEngine {
     /// # Errors
     ///
     /// Returns `MemoryError::Conflict` if the target path already exists.
-    /// Returns `MemoryError::NotFound` if the backup file doesn't exist.
+    /// Returns `MemoryError::NotFound` if the backup path is not an existing
+    /// regular file (missing, or a directory / other non-file).
     /// Returns `MemoryError::EmbeddingDimension` on dimension mismatch.
     pub fn restore_sqlite(backup_path: &Path, config: &EngineConfig) -> Result<Self> {
         if config.path.exists() {
@@ -114,9 +115,13 @@ impl MemoryEngine {
                 "target database path already exists".into(),
             ));
         }
-        if !backup_path.exists() {
+        // Use `is_file()` (not `exists()`): a directory passed as the backup
+        // source would pass an `exists()` guard and then fail deep inside
+        // `std::fs::copy` with a confusing OS-level error. `is_file()` rejects
+        // directories and other non-regular files up front with a clear message.
+        if !backup_path.is_file() {
             return Err(MemoryError::NotFound(format!(
-                "backup file: {}",
+                "backup file is not a regular file: {}",
                 backup_path.display()
             )));
         }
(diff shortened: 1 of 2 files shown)


dutiona:
## Adversarial review summary (replaces /super-review)

Per project workflow, this PR was reviewed by **2 clean-slate adversarial reviewers** (fresh context, no implementation bias), bug-fix lens pack:

| Lens                                     | Verdict    | Key finding                                                                                                                                                                                                                                                                          |
| ---------------------------------------- | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| **correctness** (edge cases, invariants) | ✅ APPROVE | `is_file()` correctly tightens rejection (dirs, FIFO/socket/device) while still following symlinks-to-files; TOCTOU window unchanged from `exists()`; missing/empty/relative paths still reject via `NotFound`.                                                                      |
| **refute** (break the fix / test)        | ✅ APPROVE | Empirically reverted the guard → `restore_sqlite_fails_if_backup_is_directory` **fails** at its first assertion (directory slips through `exists()`, `fs::copy` returns `Io(InvalidInput, "neither a regular file nor a symlink…")`). Test is genuinely discriminating, not vacuous. |

**Result: 2/2 approve, 0 blockers.** Non-blocking nit (both reviewers): `NotFound` for an existing-but-directory source is semantically "wrong-type" rather than "not-found"; the message disambiguates. Future-polish, not merge-blocking.

PR #580, opened 2026-06-16
fix(mcp): collision-safe default dump path (flaky test_dump_state_json)

### LINKED ISSUE #546: test(mcp): test_dump_state_json flaky — temp-file race on default dump path
## Flaky test surfaced during #116 (PR #544) work

`memory-engine-mcp` `test_dump_state_json` intermittently fails under parallel `cargo test` execution: the default dump path is derived from `temp_dir` + a **millisecond** timestamp, so two concurrent dumps collide on the same path.

## Likely root
Default dump-path generation has insufficient entropy (ms-resolution timestamp, no pid/nanos/counter). This is either:
- a **test** issue (the test should use a unique `tempfile::tempdir()` per run), or
- a **product** concern (the default dump path is not collision-safe under concurrency).

## Repro
Run the MCP test suite with parallelism (default). Failure is intermittent.

## Suggested fix
Give the default dump path unique entropy (pid + nanos or an atomic counter), and/or pin the test to a unique temp dir.

_Collateral finding from epic #241 Phase B; filed separately per the one-concern-per-issue contract._

dutiona:
## Summary

Hardens the fix for the intermittently flaky `memory-engine-mcp` test `test_dump_state_json` under parallel `cargo test`, and — per review — replaces a non-discriminating regression test with one that actually backstops the collision guard.

## What actually fixes the race

The within-process collision is eliminated by the **process-global atomic counter** `NEXT_DUMP_SEQ` (added in #538): any two default dumps in the *same* process (the exact flaky case — `cargo test` threads share one process) get distinct `seq` values regardless of clock resolution. The **pid** disambiguates across processes. These two together fully cover collision-safety.

The earlier millisecond→nanosecond timestamp bump in this PR is **cosmetic** (precision + time-ordering), not the causal fix; the counter already covers the same-process case the timestamp could not. The PR description has been corrected accordingly — the prior framing overstated the ms timestamp as the collision cause.

## Product change

- Factor naming into a **pure** `default_dump_name(timestamp, pid, seq, ext)` and have `default_dump_path(base_dir, ext)` call it with the live clock/pid/atomic. Filename remains `memory-dump-<nanos>-<pid>-<seq>.<ext>`.

## Test change (addresses review BLOCKERs)

The prior unit test `default_dump_paths_are_unique_within_process` was **non-discriminating**: with the real nanosecond clock, 1024 sequential calls never share a tick, so the timestamp alone satisfied the uniqueness assertion. Mutation testing confirmed it — deleting `seq` from the filename left the test passing — so it validated the clock, not the counter, and would silently miss a regression that strips the guard (the case that bites on coarse-clock CI/Windows hosts).

- New `default_dump_names_are_distinguished_by_seq_alone`: calls the pure helper 1024x with a **frozen** timestamp and pid, so distinctness can come *only* from `seq`. Asserts all 1024 names are unique and every `seq` segment `0..1024` is present. **Verified discriminating**: under the mutation that drops `seq`, the test fails (1 unique of 1024).
- Retained `default_dump_paths_are_unique_within_process` as an end-to-end smoke test of the live `default_dump_path` (well-formed, base-rooted, unique).

## Verification

- `cargo fmt --check` clean
- `cargo build --workspace` ok
- `cargo clippy --workspace --all-targets` exit 0
- `cargo test --workspace` ok (707 core + all CLI/MCP/embed suites pass)
- Mutation check: removing `seq` from `default_dump_name` makes the new test fail (1/1024); restored.

Branch rebased onto `origin/main` to pick up the #556 typestate-builder migration (the CLI integration test had moved from `MemoryEngine::open` to `MemoryEngine::builder`, which was breaking `--workspace` clippy/tests on the stale branch).

Fixes #546

### DIFF
--- a/memory-engine-mcp/src/tools/mod.rs
+++ b/memory-engine-mcp/src/tools/mod.rs
@@ -1050,6 +1050,20 @@ fn handle_forget(
 /// concurrent dumps (e.g. parallel tests) never collide on the timestamp.
 static NEXT_DUMP_SEQ: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
 
+/// Assemble a default dump filename from its already-resolved components.
+///
+/// Pure (no clock, no process state, no atomic): given a fixed `timestamp`,
+/// `pid`, and `seq`, it always yields the same `memory-dump-<ts>-<pid>-<seq>.<ext>`
+/// name. The atomic counter (`seq`) is the load-bearing uniqueness guard for
+/// same-process dumps; the timestamp only keeps names time-ordered and the pid
+/// disambiguates across processes. Factored out so a test can hold `timestamp`
+/// and `pid` constant and prove that `seq` alone makes the names distinct — if
+/// the counter were dropped the names would collide, which the timestamp would
+/// otherwise mask on a host with a fine-grained clock.
+fn default_dump_name(timestamp: &str, pid: u32, seq: u64, ext: &str) -> String {
+    format!("memory-dump-{timestamp}-{pid}-{seq}.{ext}")
+}
+
 /// Build a collision-safe default dump path inside `base_dir`.
 ///
 /// The filename combines a nanosecond timestamp, the process id, and a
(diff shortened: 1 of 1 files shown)


dutiona:
## Adversarial review summary (replaces /super-review)

3 reviewers — initially **2 blockers → remediated**:

| Lens                     | Verdict                | Key point                                                                                                                                                                                                |
| ------------------------ | ---------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **does-it-fix-the-race** | 🔴→✅ block→remediated | Original test was **non-discriminating**: the nanosecond clock alone made 1024 paths unique, so the atomic seq counter (the real race guard) was never load-bearing — deleting it left the test passing. |
| **test-quality**         | 🔴→✅ block→remediated | Same root cause.                                                                                                                                                                                         |
| **correctness**          | ✅ APPROVE             | The fix (atomic seq + ns + pid) is correct; no change to dump semantics.                                                                                                                                 |

**Remediation:** split a pure `default_dump_name(timestamp, pid, seq, ext)` out of `default_dump_path`; added `default_dump_names_are_distinguished_by_seq_alone` (1024× with a **frozen** timestamp+pid, so uniqueness can only come from `seq`) — **mutation-verified** (dropping `seq` → 1 unique of 1024, fails). Kept the original as an end-to-end smoke test, and corrected the PR body (the ms→ns bump is cosmetic; the atomic counter is the guard).

Gates: build --workspace + core/mcp tests (917) + clippy + fmt all green. **Fixes #546.**

Issue #573, opened 2026-06-16
refactor(qa): address super-qa #124 low design findings (L11/L12/L14/L16/L18/L21)


dutiona:
## Design / refactor "low" findings from super-qa #124

Grouped follow-up for the **design/refactor "low" findings** from the super-qa sweep tracked in #124 (epic #241). Re-confirmed present on `main` during the #124 reconciliation.

| ID  | Finding                                                   | Location (current)                             | Note                                                       |
| --- | --------------------------------------------------------- | ---------------------------------------------- | ---------------------------------------------------------- |
| L11 | `importance` vs `importance_score` naming confusion       | `src/types.rs:~162`                            | overlaps #571 (importance not enforced)                    |
| L12 | `NewFact` has many fields, no builder                     | `src/types.rs:~349`                            | builder ergonomics (cf. the `MemoryEngineBuilder` pattern) |
| L14 | `schema.rs` is large (now ~2751 LOC, migrations inline)   | `src/store/schema.rs`                          | split migrations into per-version modules                  |
| L16 | Read pool blocks forever (no timeout)                     | `src/pool/connection_pool.rs` (`Condvar` wait) | add an acquire timeout                                     |
| L18 | `AsyncMemoryEngine` spawn_blocking boilerplate (~674 LOC) | `src/async_engine.rs`                          | macro / codegen to reduce repetition                       |
| L21 | `hybrid_search` loads facts individually vs batch         | `src/search/hybrid.rs`                         | batch the fact fetch                                       |

_Note: L15 (pseudo-`Fact` construction for `SummaryGenerator`) is already tracked as #495. L1/L2 (benign `as f64` casts) and L20 (`pub(crate) pool` test seam) were assessed won't-fix in the #124 triage._

_Spun out of #124 · part of epic #241._

dutiona:
**Scope split** (concurrent-session conflict avoidance):

**In progress now** (low-conflict, quiet files):
- **L11** importance/importance_score naming clarity (types.rs)
- **L12** NewFact builder (types.rs)
- **L16** read-pool acquire timeout (connection_pool.rs)
- **L21** hybrid_search batch fact-fetch (hybrid.rs)

**Deferred until main settles** (collide with concurrent schema/async work):
- **L14** split schema.rs (2751 LOC) into per-version migration modules — directly conflicts with the in-flight schema work (#555 v11)
- **L18** AsyncMemoryEngine spawn_blocking boilerplate

[...]

Issue #539, opened 2026-06-15
chore(build): clear clippy -D warnings debt + rustfmt import drift on main


dutiona:
## Pre-existing build-hygiene debt on `main` (base `63ee3c6`)

Surfaced during epic #241 Phase A. The strict gate `cargo clippy --workspace --all-targets -- -D warnings` is **red on untouched `main`**, and `cargo fmt --check` flags import-order drift. Neither appears to be gated by CI (repo has only `add-to-project.yml`), so this accumulated since the #515/#517 hygiene pass.

### clippy `-D warnings` (44 default / 18 with `--features ann`) — all pre-existing

- **26×** "docs for function returning `Result` missing `# Errors` section" — mostly `src/async_engine.rs:91..229`
- **10×** `pub(crate)` module inside a private module (`clippy::redundant_pub_crate`) — `src/store/mod.rs:14-16` et al.
- **3×** missing backticks in docs
- **1×** each: `map_or_else`, `too_many_lines` (104/100), `const fn`, pass-by-value, pass-by-ref

> The **project gate** (`clippy::all = deny`, `pedantic`/`nursery = warn`) is **green** — these are pedantic/nursery warnings only promoted to errors under `-D warnings`. **Decision needed:** keep them as warnings, or clear them and pin the stricter gate.

### rustfmt drift (2 files)

- `src/conflict/temporal.rs:1`, `src/store/facts.rs:1` — `use rusqlite::{params, Connection}` → `{Connection, params}`
- _PR #536 (#112 fix) incidentally fixes these two files; after it merges this part may be nil._

### Recommendation

A single repo-wide `chore(build)` PR to clear the clippy `# Errors` / `redundant_pub_crate` debt, then optionally add a CI clippy/fmt workflow so it cannot regress silently. Mirrors the #515/#517 hygiene work.

_Collateral finding from epic #241 Phase A; filed separately per the one-concern-per-issue contract._

dutiona:
Deferring (status:parked). Reassessed on current main: now **154** clippy findings (up from 44), **all `pedantic`/`nursery`** — the project gate is `clippy::all = deny`, so these are warnings, **not gate-blocking**. fmt drift is now **0** (cleaned up). A 154-site sweep across files the concurrent sessions (#555/#560/#572) are actively editing would be a merge-conflict minefield for non-required debt. Will run as one clean sweep once `main` quiesces.

PR #587, opened 2026-06-16
perf(retrieval): batch fact fetch in hybrid_search

dutiona:
## What

`hybrid_search` materialized the post-RRF ranked facts **one id at a time** — N separate `SELECT ... WHERE id = ?` round-trips after fusion. This replaces the per-id loop with a single batched fetch.

## How

- New `FactStore::get_many(&[i64]) -> HashMap<i64, Fact>` resolves all ranked ids in one query: `WHERE id IN (SELECT value FROM json_each(?1))`.
  - `json_each` binds the id set as a single JSON parameter, so SQLite's bound-variable limit is never hit regardless of result size — no chunking needed (ranked sets are bounded by `effective_target` anyway).
  - Missing ids are simply absent from the map (no error), preserving the prior warn-and-skip-on-missing behavior.
- `hybrid_search` collects the ranked ids, calls `get_many` once, then walks `ranked` in order, `remove`-ing each fact from the map.

## Behavior preservation

This is behavior-preserving:
- **Same facts, same ranked order** — the iteration still walks `ranked` and re-indexes the map by id.
- **Temporal post-filter** (`t_valid`/`t_invalid` vs `cutoff`) is unchanged and still applied in the ranked-order loop.
- **`effective_limit` cap** and **`fact_type`/`scope`/`t_expired`** SQL-level filters are untouched.
- **Missing-id skip** matches the old `store.get` warn-and-continue path.
- Ranked ids are unique (FTS/vector lists carry distinct ids; RRF dedupes via `HashMap`), so `remove` is equivalent to the old per-id `get`.

The existing `hybrid_search` / `execute_query` tests pass unchanged (the oracle), plus a new `get_many_round_trips_and_skips_missing` unit test locks in the batch getter (round-trip, empty-slice no-op, absent missing id).

## Gates

- `cargo build --workspace` ✅
- `cargo test -p memory-engine -p memory-engine-mcp -p memory-engine-embed` ✅ (exit 0)
- `cargo clippy -p memory-engine -p memory-engine-mcp --all-targets` ✅ (exit 0, delta-clean — zero new warnings vs the pre-existing #539 set)
- `cargo fmt --check` ✅

Refs #573 (L21)

### DIFF
--- a/src/search/hybrid.rs
+++ b/src/search/hybrid.rs
@@ -211,12 +211,23 @@ pub fn hybrid_search(
     let effective_limit = effective_target;
     let ranked_count = ranked.len();
     let store = FactStore::new(conn, embed_dim);
+
+    // Batch-materialize all ranked ids in a single round-trip, then re-order
+    // to the ranked order below. Fetching every ranked id (not just the first
+    // `effective_limit`) preserves the previous semantics: the temporal
+    // post-filter can drop candidates, so later ranked ids may be needed to
+    // fill the limit. The map keys back to ranked order; ids the store failed
+    // to load (e.g. raced deletion) are simply absent and skipped, matching
+    // the prior per-id `store.get` warn-and-continue behavior.
+    let ranked_ids: Vec<i64> = ranked.iter().map(|&(id, _)| id).collect();
+    let mut facts_by_id = store.get_many(&ranked_ids)?;
+
     let mut results = Vec::new();
     for (id, score) in ranked {
         if results.len() >= effective_limit {
             break;
         }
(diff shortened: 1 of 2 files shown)


dutiona:
Adversarial review (Wave-1 workflow): correctness/api-design/maintainability — approve + 2 minor (no blockers). `FactStore::get_many` via `json_each` (single round-trip, no bound-var limit), behavior-preserving ranked order. Rebased onto current main (test-module conflict in facts.rs resolved — kept both #570's reinforce tests + the new get_many test); 814 core tests green, clippy + fmt clean. Refs #573 (L21).

Issue #608, opened 2026-06-16
Plan: skip DreamCycle if caller wrote facts (#209)


dutiona:
## Plan

# Implementation Plan — Skip DreamCycle if caller already wrote facts (#209)

> **Ephemeral artifact.** Per housekeeping `7bddc48`, `docs/plans/` is kept empty on the public repo.
> This file is a working artifact for the super-plan review loop only — it is **archived as a GitHub
> issue** and **never committed**. The #209 PR rebases onto post-`ac15b15` main (empty `docs/plans/`)
> and never stages anything under `docs/plans/`.

## Problem

`#554` swaps the harness's interim Ollama consolidator for ME's `dream_cycle`. The harness fires
**both** fact-writes (`add_fact` / `flush_insights`) and the cycle on the **same SessionStart
trigger** → a write/consolidate race. `#209` is the hard-gate: before running, the cycle detects
whether the _caller_ wrote facts since a cursor and, if so, **defers** (advances the cursor, skips
this run). A later quiet invocation clears the backlog.

## Fixed design decision (user-approved — research gate satisfied)

- **Cursor = fact-id high-water mark.** A persisted config value `last_caller_write_fact_id` holding
  the max `facts` rowid of _caller-written_ facts seen at the last decision. On guarded entry: if a
  caller-written fact exists with `id > cursor`, **advance the cursor to the new max and skip**;
  otherwise **run**. `facts.id` is `INTEGER PRIMARY KEY AUTOINCREMENT` (`schema/mod.rs:457`) → ids are
  monotonic and never reused (and facts are soft-deleted, never hard-deleted), so the HWM is sound by
  construction.
- **Cursor advances ONLY on skip**, never on a real run. A real run dream-marks everything it
  processed, so those facts drop out of the "caller-written" signal — the marker, not the cursor,
  protects a run. (Steady state: each new caller-write batch ⇒ exactly one skip, then a run.)
- **Exclude the cycle's OWN derived facts** from the signal. **This is the enabling fix — see
  Invariant M.**

## Invariant M (the enabling fix — Task 1)

**"Every fact a cycle creates OR leaves as the live survivor of a delta is dream-marked in the same
transaction."**

Today `apply_cycle_report` (`apply.rs:182`) stamps `mark_dream_cycled` over **only**
`report.metadata.processed_ids`. But a cycle produces _active, unmarked_ facts four ways:

[...]

Issue #554, opened 2026-06-15
epic: pluggable, opt-in consolidation backend (#554 reframe)


dutiona:
## Sequencing plan

Cross-issue sequencing plan: [`docs/design/plans/2026-06-16-epic-554-ollama-swap-sequencing.md`](https://github.com/dutiona/memory-engine/blob/main/docs/design/plans/2026-06-16-epic-554-ollama-swap-sequencing.md) (PR #576).

**Scope decision — Minimal + #209.** Binding prerequisite is #567's exit subset (#49 default DreamCycle impl + #225 MCP endpoints), per the blocked-by comment above. #209 (skip-if-caller-wrote-facts) is promoted to a **hard gate** before the harness re-wire — the swap introduces a write/consolidate race on the shared SessionStart trigger. #57/#231/#207/#208 may close #567 *after* #554.

Critical path: `#49 → #225 → #209 → prove-equivalent → my-dotfiles#132 → #554`. Dependency states verified against GitHub 2026-06-16.

dutiona:
**Blocked by #567** (epic: path to Phase 5a fully featured). The harness re-wire in this epic's final leaf (my-dotfiles#132) needs ME `dream_cycle`/`get_recent_insights` to actually exist — that is #567's exit (#49 default impl + #225 MCP endpoints). #554 cannot complete until #567 lands.

dutiona:
## Reframe (2026-06-17): pluggable opt-in backend + efficiency×quality benchmark

This epic is **reframed** from *"decommission the interim Ollama consolidator, replace with ME dream_cycle"* to:

**Keep the LLM consolidation pipe — make the consolidation backend pluggable + opt-in**, and **benchmark** the trade. A consumer chooses the **dream_cycle (deterministic)** backend or an **LLM** backend; both stay supported.

### Design (decided + multi-model reviewed)
- The LLM backend is an **`LlmDreamCycle` impl of the existing `DreamCycle` trait**, emitting the same `CycleReport` deltas — "which backend" = "which `&dyn DreamCycle`", reusing the whole #209/#225 apply spine (no core signature change). It holds its own `EmbeddingProvider`; the HTTP `DeltaProposer` lives in `memory-engine-embed`.
[...]

Issue #225, opened 2026-04-08
feat(mcp): cognitive pipeline endpoints (dream_cycle, get_recent_insights)


dutiona:
## Parent

Umbrella: #221

## Context

Subset of #221 that depends on Phase 5a cognitive infrastructure. These endpoints expose DreamCycle and InsightStream capabilities as MCP tools.

### Consolidation

- **`dream_cycle()`** — SessionStart hook (daily) triggers consolidation pipeline: dedup/pruning, pattern detection, promotion, rescoring. Returns `CycleReport`.

### Context Retrieval

- **`get_recent_insights(project_path, limit?)`** — Returns high-value model-logged insights (insight stream).

## Prerequisites

- [ ] #49 — DreamCycle trait + pipeline implementation
- [ ] #48 — InsightStream trait (or `FactType::Insight`)
- [ ] #55 — Provenance infrastructure (prerequisite of #49)
- [ ] #63 — Outcome tracking (prerequisite of #49)

## Acceptance Criteria

- [ ] `dream_cycle` runs consolidation (dedup/pruning) on stored activities and returns `CycleReport`
- [ ] `get_recent_insights` returns model-logged insights filtered by project and limited
- [ ] Both endpoints exposed as MCP tools
- [ ] Integration tests

## Dependencies

Blocked on Phase 5a: #48, #49 (which itself needs #55, #63).

## Refs

- Umbrella: #221
- NYX12 architecture: dutiona/knowledge-base#102

dutiona:
## Parent (epics)

- Umbrella: #221 (MCP endpoint exposure)
- Epic: #554 (replace Ollama interim consolidator) — this issue delivers the
  `dream_cycle`/`get_recent_insights` endpoints that epic swaps Ollama for.

Issue #207, opened 2026-04-02
feat(consolidate): distributed lock for DreamCycle


dutiona:
## Context

Claude Code's `autoDream` uses a distributed lock (mtime + PID + 1h staleness) to ensure only one consolidation process runs at a time across multiple concurrent sessions (note 29 §3.3).

## What

Implement a single-writer guarantee for DreamCycle consolidation:
- Lock file with PID as body, mtime as acquisition timestamp
- Stale detection: if lock mtime ≥ 1h AND PID is dead → reclaim
- Rollback on failure: restore prior mtime (or unlink if no prior lock existed)
- Race condition handling: both processes write PID; last write wins, loser re-reads and bails

## Engineering details (from CC source)

- CC's `consolidationLock.ts` uses file mtime + PID. For Rust: `flock()` is simpler and race-free for single-machine, but CC's approach works across restarts without a daemon.
- **Critical: rollback on failure.** Without it, a crash blocks future consolidation for 1h. CC's `rollbackConsolidationLock(priorMtime)` restores the prior state or unlinks.
- Consider `advisory_lock` in SQLite as an alternative for our single-DB architecture.

## Source

Steal-list item M2 from note 29.

dutiona:
**[Audit amendment — 2026-07-06]** Scope interaction to resolve BEFORE designing the lock: ADR-0016 §(b) adds a `MaterializationScheduler` that runs as a dream-cycle sub-step and mutates `DerivedStructureRegistry` rows (`dirty` → `rebuilding` → `fresh`) with `spawn_blocking` builds. Two processes racing that registry produce exactly the torn `rebuilding` states the ADR's crash-guards only half-handle.

Acceptance criterion to add: the lock's scope statement explicitly includes or excludes the MaterializationScheduler — and if excluded, names what protects the registry instead. If the answer is unclear at design time, STOP and escalate; do not default silently.

PR #609, opened 2026-06-16
feat(cognitive): skip DreamCycle if caller already wrote facts

### LINKED ISSUE #209: feat(consolidate): skip DreamCycle if caller already wrote facts
## Context

Claude Code's `extractMemories` checks `hasMemoryWritesSince()` — if the main agent already wrote to memory, the forked extraction agent skips entirely. This prevents redundant writes and race conditions (note 29 §3.2).

## What

Before starting a DreamCycle run, check if any `add_fact()` calls happened since the last cursor position. If yes, advance the cursor and skip consolidation — the caller already handled persistence.

## Engineering details

- CC scans tool call results for FileEdit/FileWrite to auto-memory paths.
- For us: simpler — query `SELECT COUNT(*) FROM events WHERE id > ?cursor AND event_type = 'fact_added'`.
- This also prevents the coalescing problem: if consolidation is in-flight when new facts arrive, skip rather than queue.

## Source

Steal-list item M4 from note 29.

dutiona:
Closes #209. Plan archived as #608. Epic #554 (Ollama→ME swap).

## What

Adds `MemoryEngine::run_dream_cycle_guarded` — the write/consolidate-race **hard-gate** for the #554 harness swap, where fact-writes (`add_fact`/`flush_insights`) and the cycle fire on the same SessionStart trigger.

When the caller wrote facts since a persisted **fact-id cursor** (`last_caller_write_fact_id`), the cycle **defers**: advances the cursor and returns `CycleOutcome::Skipped(SkipReason::CallerWroteFacts{..})`. A later quiet invocation runs and clears the backlog (defer, not drop). No new writes → `CycleOutcome::Ran(report)`.

## The enabling fix — Invariant M (own commit, `fix(cognitive)`)

`apply_cycle_report` marked only the cycle's *inputs* (`processed_ids`), leaving its *outputs* active+unmarked: the `AddFact` synthetic, the promoted wisdom fact (id was discarded — `let _ = promoted`), and a `Supersede` survivor. Those re-entered the next cycle's `list_undreamt_in_period` input (a **latent double-processing bug**) and, with #209's cursor, read as fresh caller writes (self-skip-forever).

Fix: mark `processed_ids ∪ new_fact_ids ∪ promoted_fact_ids ∪ supersede_new_ids`. **This PR also closes that pre-existing double-processing bug.**

## Design notes

- **Cursor = fact-id high-water mark** (`facts.id` is `AUTOINCREMENT` → monotonic, never reused; guard test added). Advances only on skip; a real run relies on the dream-marker (invariant M) to drop processed facts from the signal.
- `CycleOutcome::{Ran, Skipped}` — a skip is *unrepresentable as a report*, so it can't be wrongly applied. `run_dream_cycle` unchanged (preserves the #225 API).
- `DreamCycle::run` now documents the `processed_ids` contract; a malformed report (`facts_selected>0`, empty `processed_ids`) → `CycleError::MalformedReport`.
- **Deferral, not mutual exclusion** — concurrent guarded calls can both run (idempotent via marker+watermark); true locking is #207.
- MCP `memory_dream_cycle` uses the guarded path: skip → `{did_run:false, skipped}`, run → `{did_run:true, report, did_apply, applied?}`. **`did_run` is a new response key** (`mcp-server.md` updated).

## Test plan

- [x] `cargo test --workspace` green (1122 passed)
- [x] `cargo test --all-features` green
- [x] `cargo clippy --workspace --all-targets` exit 0
- [x] invariant-M test isolates the Supersede-survivor escape
- [x] guarded skip/run/cold-start/empty-store/malformed-report covered
- [x] MCP skip + run JSON shapes covered

Reviewed via /super-plan (3 lens drafts → synthesis → 2 adversarial subagents + clean-slate + gemini). Plan + full audit: #608.


dutiona:
## Post-implementation review (2 adversarial subagents) — resolved in `51c0592`

**Actioned:**
- **Contract clarity** — `DreamCycle::run` rustdoc now requires facts the cycle *adjusts*/*tags* (not just selects) to be in `processed_ids` (a consumer omitting a touched fact would livelock the guarded cycle).
- **Test gaps** — `MalformedReport` now tested in the **steady-state skip-then-run** path (not just cold start); a real run is asserted to **leave the cursor unchanged** + a 3rd quiet run still runs; direct `did_run` assertions on the empty-store and dry-run MCP paths; `CycleError::MalformedReport` added to the error-display suite.
- `#[must_use]` on `run_dream_cycle_guarded`.

**Considered, not actioned (with rationale):**
- *MalformedReport guard in `apply_cycle_report`*: deliberate plan decision — that method has a client-supplied-report contract where `facts_selected` is advisory; the guard belongs in the guarded cycle path only.
- *Promote-path cursor masking*: already caught by the `is_marked(promoted)` assertion (promoted facts are `is_pinned`, independently cursor-excluded).

Gate after fixes: workspace 32 suites green / 0 failed; clippy `--workspace --all-targets` exit 0; `fmt --check` clean.

Issue #627, opened 2026-06-16
Plan: pluggable consolidation backend + efficiency×quality benchmark (#554 reframe)


dutiona:
## Plan

# Implementation Plan — Pluggable, opt-in consolidation backend + benchmark (#554 reframe)

> **Ephemeral artifact.** Per memory-engine housekeeping (`7bddc48`), `docs/plans/` stays empty on the
> public repo. This file is a super-plan working artifact — **archived as a GitHub issue**, never
> committed. PRs rebase onto clean `main`; never stage `docs/plans/`.
>
> **v2 — revised after round-1 review** (clean-slate + architecture + benchmark-validity adversarial
> subagents). The review invalidated the original delta-translation, the embedding-ownership claim, and
> much of the judged-quality benchmark premise. See the `## Resolution` logs in the review artifacts.

## Reframe (user-decided)

Epic **#554** was "decommission the interim Ollama consolidator." It is now: **keep the LLM
consolidation pipe, make the consolidation backend pluggable + opt-in** (a consumer chooses the
deterministic `dream_cycle` backend _or_ an LLM backend), **and publish efficiency numbers** comparing
backends to justify the default. The LLM path is **not** removed; it becomes one selectable backend.

## Central design decision (RESOLVED + corrected)

ME has two non-interchangeable pipelines (`run_dream_cycle` → `CycleReport` deltas, zero LLM, apply via
`apply_cycle_report`; vs `consolidate()` → in-place LLM `SummaryGenerator` rewrite → `ConsolidationStats`).
Different ontologies → no fair A/B from an enum over them.

**Decision: the LLM backend is an `LlmDreamCycle` impl of the existing `DreamCycle` trait**, emitting the
same `CycleReport` deltas. "Which backend" = "which `&dyn DreamCycle`" — and `run_dream_cycle_guarded`
already takes `&dyn DreamCycle` (`cognitive.rs:247`), so **no core signature change**.

**Corrections from review (these were wrong in v1):**

- **Embedding ownership** — `CycleDelta::AddFact(NewFact)` carries a _pre-computed_ `embedding`;
  `apply_cycle_report` validates the dim and does NOT embed (`apply.rs:296-303`); `DreamContext` exposes
  no `embed()`. So the engine does **not** embed for the backend. **`LlmDreamCycle` holds its own
  `&dyn EmbeddingProvider`** and embeds summary text itself (the consumer already injects one; this keeps
[...]

PR #641, opened 2026-06-19
feat(cognitive): pluggable consolidation backend — Track A (#554)

dutiona:
## Summary

Track A of the **#554 reframe** ([plan: #627](https://github.com/dutiona/memory-engine/issues/627)): make the consolidation backend **pluggable and opt-in**. The deterministic `dream_cycle` and an LLM backend are now interchangeable `&dyn DreamCycle`s — no core signature change (both flow through the existing `run_dream_cycle_guarded` → `apply_cycle_report` path).

This PR is **Track A only** — Track B (the memarch-bench efficiency×quality study) and the my-dotfiles#132 harness wiring are separate, later work. It is partial progress on the epic; issues #554 and #627 stay open after merge (no closing link intended).

## What landed (six atomic commits)

| Commit | Layer |
| --- | --- |
| `CycleDelta::Synthesize` | The atomic merge primitive: insert synthetic → per-source expire + `supersedes` edge + one lineage row, all in one transaction. Cannot be `AddFact + Supersede` (the synthetic's id isn't knowable until insert) nor `AddFact + Quarantine` (orphans lineage). Extends `validate_report` + invariant M. |
| `DeltaProposer` + `ConsolidationProposal` | The consumer seam (`traits.rs` / `types.rs`): `propose(window, prior_wisdom) -> merge groups (ids + summary)`. Plain DTOs so a malformed proposal crossing the HTTP boundary reaches the clamp, not a parse failure. |
| `LlmDreamCycle` | The selectable LLM backend (`DreamCycle` impl). Holds its own `EmbeddingProvider` (engine stays LLM-free). Three rails vs. an untrusted proposer: **window clamp**, **`processed_ids` = whole window** (#209-safe), **apply-time validation**. Merge inherits max source importance. |
| `HttpDeltaProposer` | HTTP proposer in `memory-engine-embed` (Ollama `/api/generate`, JSON mode, temperature 0). Pure parse helpers + interior-atomic token capture (`eval_count`/`prompt_eval_count`). |
| CLI `consolidate --backend {dream-cycle\|llm}` | The subprocess seam memarch-bench drives. Opens writable, runs guarded, applies, prints machine-readable JSON (deltas, wall time, token counts). |
| docs | dream-cycle.md pluggable-backends + Synthesize row, cli-inspector.md subcommand, CLAUDE.md status, `LlmDreamCycle` injection rustdoc. |

## Testing

TDD throughout. New coverage: `Synthesize` apply (happy/edges/lineage/invariant-M + empty/missing/expired-source/dim rejections), `LlmDreamCycle` with a fake proposer (window-clamp, forgetful/hostile/out-of-window, embed-dim, cycle_id), `HttpDeltaProposer` pure-helper units + wiremock round-trip, and a full CLI **wiremock end-to-end** (`consolidate --backend llm` synthesizes one merge and reports captured tokens).

Gate green: `cargo test --workspace` (1151 passed), `cargo clippy --workspace --all-targets --all-features -- -D warnings` (clean), `cargo fmt --all --check` (clean).

Part of #554. Implements Track A of #627.


src/engine/cycle/llm_impl.rs
```diff
@@ -0,0 +1,396 @@
...
+        let now = Utc::now();
+        let mut deltas: Vec<CycleDelta> = Vec::new();
+        for group in &proposal.merges {
+            // Clamp every proposed id to the fed window (dropping out-of-window /
+            // hallucinated ids) and de-duplicate, preserving first-seen order. The
+            // engine's validate_report checks existence/active but NOT window
+            // membership, so this clamp is what stops an LLM acting on facts it was
+            // never shown. A group left with no in-window source is skipped (emitting
+            // an empty Synthesize would fail validation).
+            let mut seen = HashSet::new();
```
dutiona: [Antigravity Review] [BLOCKER] LlmDreamCycle missing global cross-group source deduplication causes livelock.
`seen` is instantiated inside the `for group` loop, deduplicating only within a single merge group. If the LLM proposes the same `source_id` across distinct merge groups, multiple `Synthesize` deltas will contain the same source. `apply_cycle_report` tracks in-report expiries, so the second `Synthesize` delta will fail `ensure_active` with `AlreadyExpired`. This aborts the entire transaction, `processed_ids` are never marked, and the deterministic LLM gets stuck in a permanent livelock analyzing the exact same window. Move `let mut seen = HashSet::new();` outside the loop.

memory-engine-cli/src/commands/consolidate.rs
```diff
@@ -0,0 +1,151 @@
...
+            )?;
+            let embedder = HttpEmbeddingProvider::new(
+                embed_url.to_owned(),
+                embed_model.to_owned(),
+                None,
+                dim,
+                args.timeout_secs,
+            )?;
+            let cycle = LlmDreamCycle::new(&proposer, &embedder);
+            let outcome = engine.run_dream_cycle_guarded(&cycle)?;
```
dutiona: [Antigravity Review] [HIGH] CLI `consolidate` command missing #209 double-run drain loop.
If a caller write happens to move the #209 cursor, `run_dream_cycle_guarded` returns `Skipped(CallerWroteFacts)` and the CLI gracefully exits. For a manual command intended to force consolidation, this means the user (or benchmark) gets no work done. The CLI should implement a `while let CycleOutcome::Skipped(SkipReason::CallerWroteFacts { .. })` loop to retry and drain the backlog.

memory-engine-cli/src/commands/consolidate.rs
```diff
@@ -0,0 +1,151 @@
...
+            (outcome, Some(proposer.stats()))
+        }
+    };
+
+    // Apply the report if the cycle ran; a skip is a deferral, not a failure.
+    let (outcome_label, applied, skip_reason) = match &outcome {
+        CycleOutcome::Ran(report) => {
+            let result = engine.apply_cycle_report(report)?;
```
dutiona: [Antigravity Review] [MEDIUM] JSON output contract broken on `apply_cycle_report` failure.
If `engine.apply_cycle_report(report)?` fails (e.g. due to a validation error like dimension mismatch), the `?` operator propagates the error and the CLI terminates. The JSON report is never printed to `stdout`, breaking the machine-readable contract for the Track B benchmark. Catch the error and serialize it into the JSON output.

memory-engine-embed/src/proposer.rs
```diff
@@ -0,0 +1,309 @@
...
+    ) -> Result<(ConsolidationProposal, u64, u64), MemoryError> {
+        let response = body
+            .get("response")
+            .and_then(serde_json::Value::as_str)
+            .ok_or_else(|| {
+                MemoryError::Internal(
+                    "proposer response missing 'response' string field".to_owned(),
+                )
+            })?;
+        let proposal: ConsolidationProposal = serde_json::from_str(response).map_err(|e| {
```
dutiona: [Antigravity Review] [MEDIUM] JSON parsing fails on markdown code blocks.
`serde_json::from_str` strictly expects raw JSON. Even with Ollama`s `format: "json"`, some models occasionally wrap their JSON in markdown fences (e.g. ```json). This will fail parsing, burning LLM tokens. Defensively strip markdown fences before deserializing.

dutiona (commented):
[Antigravity Review] Code review completed for Track A pluggable consolidation backend.
I've reviewed correctness of the Synthesize primitive, LlmDreamCycle safety rails, HttpDeltaProposer handling, CLI integration, and test gaps. 
Found a [BLOCKER] livelock bug in LlmDreamCycle due to lack of global deduplication across merge groups, a [HIGH] missing #209 double-run loop in the CLI, and a couple [MEDIUM] JSON handling/output bugs. See inline comments and test gaps documented in the report.

src/engine/cycle/llm_impl.rs
```diff
@@ -0,0 +1,396 @@
...
+                t_created: now,
+                t_expired: None,
+                t_valid: None,
+                t_invalid: None,
+                source_event_id: None,
+                importance,
+                access_count: 0,
+                last_accessed: now,
+                metadata: serde_json::json!({}),
+                scope_id: 1, // root scope (v1)
```
dutiona: [Codex Review] `scope_id` is hard-coded to root here, but the cycle window is selected across all scopes. If the proposer merges facts from a child scope, the synthesized replacement becomes globally visible at scope 1, which breaks scope isolation and widens retrieval unexpectedly. This should inherit a shared source scope or reject cross-scope groups before emitting `Synthesize`.

src/engine/cycle/llm_impl.rs
dutiona: [Codex Review] This only de-duplicates `source_ids` within a single merge group. If the proposer reuses the same fact in two different groups, the second `Synthesize` hits `AlreadyExpired` during report validation and the entire cycle aborts. Because this backend treats the proposer as untrusted input, it needs a cross-group consumed-id set here so one overlapping id does not take down the whole run.

src/engine/cycle/apply.rs
```diff
@@ -182,19 +189,72 @@ impl MemoryEngine {
...
+                        supersede_edges.push((synth_id, *src, edge_id));
+                    }
+                    // 3. One lineage row maps the synthetic to all its sources, so the
+                    //    knowledge-base can trace the merge (parity with promotion's
+                    //    provenance chain). `representative_ids` is capped at the first
+                    //    few per its quick-review contract; lineage holds the full set.
+                    let provenance = PromotionProvenance {
+                        source_count: u32::try_from(sources.len()).unwrap_or(u32::MAX),
+                        session_count: 0,
+                        date_range_start: new_fact.t_created,
```
dutiona: [Codex Review] The lineage row is correct, but this provenance envelope is not describing the merged sources: `date_range_start` and `date_range_end` are both set to the synthetic fact creation time rather than the min/max timestamps of the source facts. Any inspector or KB consumer reading lineage provenance will be told that the merge covers only the cycle instant, which defeats the point of storing source range metadata. This should be derived from the source facts the same way `cluster_provenance` does.

memory-engine-embed/src/proposer.rs
dutiona: [Codex Review] The accounting path starts by deserializing the inner `response` payload, and only reads `eval_count` / `prompt_eval_count` after that succeeds. A 200 response with malformed merge JSON therefore consumes a real LLM call but records zero calls and zero tokens, which will undercount the Track B benchmark exactly on the failure paths. Please extract and accumulate the usage fields even when proposal parsing fails.

dutiona (commented):
[Codex Review] I found 4 substantive issues in the pluggable consolidation backend: (1) synthesized facts are forced into root scope even though the cycle window spans all scopes, which breaks scope isolation; (2) overlapping merge groups from the proposer are not sanitized across groups, so one repeated source id aborts the entire cycle; (3) proposer usage accounting is only updated after merge JSON parses successfully, so failed 200 responses undercount calls and tokens; and (4) `Synthesize` provenance stores the synthetic creation time instead of the source fact date range. Workspace verification was otherwise clean: `cargo test --workspace` passed.

src/engine/cycle/llm_impl.rs
dutiona: Fixed in c32bbc3: hoisted a global `consumed` set so each source is claimed once across all merge groups (claimed only on emit). Added test `run_dedups_sources_across_merge_groups`.

dutiona: Fixed in c32bbc3: global cross-group `consumed` set (claimed on emit). Same fix as the Antigravity thread.

src/engine/cycle/llm_impl.rs
dutiona: Fixed in c32bbc3: the synthetic now inherits its sources' shared scope; a cross-scope group is skipped (no merge across an isolation boundary). Test `run_inherits_source_scope_and_skips_cross_scope_groups`.

memory-engine-cli/src/commands/consolidate.rs
dutiona: Fixed in c32bbc3: added a bounded #209 drain loop (`run_with_drain`) so one `consolidate` invocation drains a caller-write deferral and runs. Test `dream_cycle_backend_drains_and_runs_in_one_invocation`.

memory-engine-cli/src/commands/consolidate.rs
dutiona: Fixed in c32bbc3: any consolidate-step failure (cycle run OR apply) now emits outcome:"failed" + error in the JSON AND exits non-zero (fail-loud). Test `consolidate_failure_emits_json_and_exits_nonzero`.

memory-engine-embed/src/proposer.rs
dutiona: Fixed in c32bbc3: `parse_proposal` strips a wrapping markdown code fence before deserializing. Test `parse_proposal_strips_markdown_code_fence`.

dutiona: Fixed in c32bbc3: split into `extract_usage`/`parse_proposal`; `propose` records llm_calls+tokens from the envelope BEFORE the inner parse, so a malformed-JSON 200 is still counted. Test `propose_records_call_and_tokens_even_when_merge_json_is_malformed`.

src/engine/cycle/apply.rs
dutiona: Fixed in c32bbc3: the Synthesize lineage provenance date range now spans the sources' real min/max t_created, not the synthetic instant. Test `synthesize_lineage_provenance_spans_sources_not_synthetic_instant`.

dutiona: [Codex Review] LGTM. `Synthesize` provenance now derives its date range from the merged sources rather than the synthetic timestamp, with a targeted regression test.

src/engine/cycle/llm_impl.rs
dutiona: [Codex Review] LGTM. `LlmDreamCycle` now claims source ids across the whole report, so overlapping proposer groups no longer abort apply with `AlreadyExpired`.

src/engine/cycle/llm_impl.rs
dutiona: [Codex Review] LGTM. The synthetic now inherits a shared source scope and cross-scope groups are skipped, which closes the scope-leak reported here.

memory-engine-embed/src/proposer.rs
dutiona: [Codex Review] LGTM. Usage is now extracted and recorded before inner proposal parsing, and the new integration test covers the malformed-JSON 200 path.

src/engine/cycle/apply.rs
```diff
@@ -636,6 +714,255 @@ mod tests {
...
+            embedding: vec![0.1, 0.2, 0.3, 0.4],
+            fact_type: FactType::Semantic,
+            t_created: "2026-06-16T00:30:00Z".parse().unwrap(),
+            t_expired: None,
+            t_valid: None,
+            t_invalid: None,
+            source_event_id: None,
+            importance: 0.5,
+            access_count: 0,
+            last_accessed: "2026-06-16T00:30:00Z".parse().unwrap(),
```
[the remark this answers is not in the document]
dutiona: Not applicable here (respectfully): these are test-helper literals parsed by chrono into `DateTime<Utc>`, which normalizes `Z` and `+00:00` to the same instant — the stored format is set by the insert path's serialization, not the literal. The #641 provenance fix compares `DateTime` values with `.min()`/`.max()` (not SQL lexicographical TEXT sort), so the Z-vs-offset distinction does not affect it. Leaving the literals as-is.

memory-engine-embed/src/proposer.rs
```diff
@@ -0,0 +1,309 @@
...
+            "embedding": [0.0, 0.0, 0.0, 0.0],
+            "fact_type": "Semantic",
+            "t_created": "2026-06-16T00:00:00Z",
+            "t_expired": null,
+            "t_valid": null,
+            "t_invalid": null,
+            "source_event_id": null,
+            "importance": 0.5,
+            "access_count": 0,
+            "last_accessed": "2026-06-16T00:00:00Z",
```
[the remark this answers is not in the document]
dutiona: Not applicable here (respectfully): these are test-helper literals parsed by chrono into `DateTime<Utc>`, which normalizes `Z` and `+00:00` to the same instant — the stored format is set by the insert path's serialization, not the literal. The #641 provenance fix compares `DateTime` values with `.min()`/`.max()` (not SQL lexicographical TEXT sort), so the Z-vs-offset distinction does not affect it. Leaving the literals as-is.

Issue #619, opened 2026-06-16
feat(cli): wire query embedding to embed_query + embedding config


dutiona:
Epic: #610 · Wave 1
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.6)

CLI query subcommands -> embed_query; ingest/bootstrap stay on embed/embed_batch. Same embedding config keys as MCP.

Issue #618, opened 2026-06-16
feat(mcp): wire query path to embed_query + embedding config


dutiona:
Epic: #610 · Wave 1
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.6)

MCP query path: embed() -> embed_query(); add_fact stays on embed. Config: model, provider, query_instruction, dim, mrl_dim.

Issue #617, opened 2026-06-16
feat(retrieval): HttpEmbeddingProvider — query_instruction + MRL truncate/renormalize + fingerprint


dutiona:
Epic: #610 · Wave 1
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.5)

Extend HttpEmbeddingProvider (already OpenAI/TEI-compatible): query_instruction (Qwen query-only prefix), mrl_dim (truncate + L2 renormalize), and fingerprint() reporting {model, provider:tei, dim, matryoshka_base_dim, element_type:float32}.

Issue #612, opened 2026-06-16
feat(core): EmbeddingFingerprint type + EmbeddingProvider::fingerprint()


dutiona:
Epic: #610 · Wave 1
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.1, §Cross-layer policy)

Add `EmbeddingFingerprint { model, provider, dim, matryoshka_base_dim, element_type }` (canonical tuple per ADR 0015) and `EmbeddingProvider::fingerprint(&self)`.

PR #644, opened 2026-06-19
feat(core): asymmetric embedding — embed_query/embed_query_batch (default to embed)

### LINKED ISSUE #616: feat(core): asymmetric embedding — embed_query/embed_query_batch (default to embed)
Epic: #610 · Wave 1
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.4)

Add embed_query/embed_query_batch defaulting to embed/embed_batch (document semantics). Symmetric providers unaffected. Core never embeds queries; consumers do.

dutiona:
## What

Adds two **default** methods to `EmbeddingProvider` (`area:core`, §Design.4 of the embedding-identity spec):

```rust
fn embed_query(&self, text: &str) -> Result<Vec<f32>> { self.embed(text) }
fn embed_query_batch(&self, texts: &[&str]) -> Result<Vec<Vec<f32>>> { self.embed_batch(texts) }
```

## Why

`Qwen3-Embedding` (the target model for epic #610) is **asymmetric**: queries take an instruction prefix, documents do not. `embed(text)` alone cannot express that. These methods give the consumer layer (MCP/CLI, wired in #618/#619) a single place to express the asymmetry.

Both **default to their document counterparts**, so the ~30 existing providers (symmetric test doubles + `HttpEmbeddingProvider`) compile and behave unchanged — no call-site churn, unlike the required `fingerprint()` in #612.

## Design notes

- **The core never calls `embed_query`.** `MemoryQuery` carries a pre-computed vector; query embedding lives only at the consumer layer. The methods sit on the provider trait purely so that layer has one asymmetry seam.
- **`embed_query_batch` delegates to `embed_batch`, not a loop over `embed_query`** — preserves single-round-trip batching for HTTP providers. ⚠️ **Latent footgun, documented in the contract:** a provider overriding only `embed_query` (not `embed_query_batch`) silently gets unprefixed (document-semantic) batch queries. The real asymmetric provider (#617) overrides both, so it never bites in practice — but the doc comment now warns implementers explicitly. (Flagging for reviewer judgment: an alternative default looping `embed_query` would trade batch-efficiency for prefix-consistency.)

## Tests

3 new trait tests: default delegation, batch routes-through-`embed_batch`-once (call-counter), and asymmetric override leaves documents prefix-free (via `&dyn` vtable).

## Verification

- `cargo build --workspace` ✅
- `cargo test --workspace` ✅ 1181 passed
- `cargo test --all-features` ✅ 975 passed
- `cargo fmt --check` ✅ clean

Wave 1 of epic #610. Unblocks #617 / #618 / #619.

Fixes #616

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -37,6 +37,52 @@ pub trait EmbeddingProvider: Send + Sync {
         texts.iter().map(|t| self.embed(t)).collect()
     }
 
+    /// Compute an embedding vector for a **query** string.
+    ///
+    /// Asymmetric models (e.g. `Qwen3-Embedding` via TEI) prepend a query-only
+    /// instruction prefix here, while [`embed`](Self::embed) (documents) stays
+    /// prefix-free. The default delegates to [`embed`](Self::embed), so symmetric
+    /// providers (`Mock`, `Hash`, `Passthrough`) need no change and queries embed
+    /// identically to documents.
+    ///
+    /// The **core never calls this** — [`MemoryQuery`](crate::MemoryQuery) carries a
+    /// pre-computed vector, so query embedding happens only at the consumer layer
+    /// (MCP/CLI). The method lives on the provider trait so that consumer layer has a
+    /// single place to express the asymmetry.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if embedding computation fails.
+    fn embed_query(&self, text: &str) -> Result<Vec<f32>> {
+        self.embed(text)
+    }
+
+    /// Compute embedding vectors for multiple **query** strings in a single call.
+    ///
+    /// The default delegates to [`embed_batch`](Self::embed_batch) (document
+    /// semantics) — **not** a private loop over [`embed_query`](Self::embed_query) —
+    /// so a provider that overrides only `embed_batch` for a single-round-trip API
+    /// still gets one batched call for queries. Asymmetric providers override this to
+    /// apply the query prefix per element.
+    ///
+    /// # Contract
+    ///
+    /// The returned `Vec` **must** have the same length as `texts`. Each element
+    /// corresponds positionally to the input query at that index.
+    ///
+    /// **If you override [`embed_query`](Self::embed_query) to apply asymmetric query
+    /// handling (e.g. an instruction prefix), you MUST also override this method.**
+    /// The default routes through [`embed_batch`](Self::embed_batch) — *document*
+    /// semantics — so leaving it at default while overriding the single-query path
+    /// silently produces unprefixed batch-query embeddings (wrong vector space).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any embedding computation fails.
+    fn embed_query_batch(&self, texts: &[&str]) -> Result<Vec<Vec<f32>>> {
+        self.embed_batch(texts)
+    }
+
     /// Declare this provider's [`EmbeddingFingerprint`] — the identity of the vector
     /// space it produces (`model`, `provider`, `dim`, ...).
     ///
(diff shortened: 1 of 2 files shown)


src/traits.rs
```diff
@@ -37,6 +37,52 @@ pub trait EmbeddingProvider: Send + Sync {
...
+    /// **If you override [`embed_query`](Self::embed_query) to apply asymmetric query
+    /// handling (e.g. an instruction prefix), you MUST also override this method.**
+    /// The default routes through [`embed_batch`](Self::embed_batch) — *document*
+    /// semantics — so leaving it at default while overriding the single-query path
+    /// silently produces unprefixed batch-query embeddings (wrong vector space).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any embedding computation fails.
+    fn embed_query_batch(&self, texts: &[&str]) -> Result<Vec<Vec<f32>>> {
```
dutiona, line 80: [Codex Review] [HIGH] `embed_query_batch` should default to looping `embed_query`, not delegating to `embed_batch`. As written, any asymmetric provider that overrides only `embed_query` silently emits document-space vectors for batched queries. That reintroduces the exact silent retrieval corruption this epic is trying to remove; batch efficiency should require an explicit override, just like `embed_batch` already does for documents.

dutiona (commented):
[Codex Review] I found 1 high-severity issue in `src/traits.rs`: `embed_query_batch` currently defaults to `embed_batch`, which silently drops query-only semantics whenever a provider overrides `embed_query` but forgets the batched variant. That makes the additive API easy to compile but easy to use incorrectly, and it reintroduces the silent wrong-vector-space failure mode this epic is trying to eliminate. I recommend making the safe default a loop over `embed_query` and requiring asymmetric providers to opt into a batched override for efficiency.

src/traits.rs
```diff
@@ -37,6 +37,52 @@ pub trait EmbeddingProvider: Send + Sync {
...
+        self.embed_batch(texts)
```
dutiona: [Antigravity Review] **[BLOCKER] Design Flaw: Silent Correctness Footgun**

The default for `embed_query_batch` delegates to `embed_batch` (document semantics) rather than looping over `embed_query`. As your own documentation notes, if a provider overrides `embed_query` (to apply an asymmetric prefix) but forgets to override `embed_query_batch`, they will silently produce unprefixed batch queries. 

Given that the entire epic (#610) exists to eliminate "silently wrong retrieval, the worst RAG failure mode", prioritizing batch performance over correctness by default is an anti-pattern here. 

**Recommendation:** Change the default to a loop over `self.embed_query(t)`. This guarantees correctness by default. Symmetric or asymmetric providers that want single-round-trip batch performance for queries can explicitly override this method to delegate to their batch implementation.

src/traits.rs
```diff
@@ -802,6 +848,86 @@ mod tests {
...
+                crate::types::EmbeddingFingerprint::new("mock", "test", 1)
+            }
+        }
+        let p = Symmetric;
+        assert_eq!(p.embed_query("hello").unwrap(), p.embed("hello").unwrap());
+    }
+
+    #[test]
+    #[allow(clippy::cast_precision_loss)] // test data, precision irrelevant
+    fn embed_query_batch_default_routes_through_embed_batch() {
```
dutiona: [Antigravity Review] **[HIGH] Test enforces unsafe default**

This test currently enforces the performance-over-correctness design. If the default for `embed_query_batch` is updated to loop over `embed_query` (to prevent the correctness footgun), this test must be updated.

It should instead verify that the default implementation routes through `embed_query` rather than `embed_batch`.

src/traits.rs
dutiona: Fixed in 7dcc6c7 — default now loops `embed_query` (`texts.iter().map(|t| self.embed_query(t)).collect()`), exactly as recommended. Test updated to assert the embed_query routing; spec §Design.4 corrected. Thanks.

src/traits.rs
dutiona: Fixed in 7dcc6c7 — adopted the loop-`embed_query` default (correct by construction, mirrors `embed_batch`→`embed`). Spec §Design.4 updated to match.

src/traits.rs
dutiona: Done in 7dcc6c7 — test renamed to `embed_query_batch_default_loops_embed_query` and now asserts each input routes through the overridden `embed_query` (prefix preserved).

src/traits.rs
```diff
@@ -37,6 +37,52 @@ pub trait EmbeddingProvider: Send + Sync {
...
+    /// The default routes through [`embed_batch`](Self::embed_batch) — *document*
+    /// semantics — so leaving it at default while overriding the single-query path
+    /// silently produces unprefixed batch-query embeddings (wrong vector space).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any embedding computation fails.
+    fn embed_query_batch(&self, texts: &[&str]) -> Result<Vec<Vec<f32>>> {
+        self.embed_batch(texts)
+    }
```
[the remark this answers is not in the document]
dutiona: Applied your suggestion verbatim in 7dcc6c7. Doc + test follow-ups in the same commit.

src/traits.rs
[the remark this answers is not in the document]
dutiona: Fixed in 7dcc6c7 — batch default now loops `embed_query`, preserving query semantics; native-batch providers override explicitly.

src/traits.rs
```diff
@@ -37,6 +37,52 @@ pub trait EmbeddingProvider: Send + Sync {
...
+    /// # Contract
+    ///
+    /// The returned `Vec` **must** have the same length as `texts`. Each element
+    /// corresponds positionally to the input query at that index.
+    ///
+    /// **If you override [`embed_query`](Self::embed_query) to apply asymmetric query
+    /// handling (e.g. an instruction prefix), you MUST also override this method.**
+    /// The default routes through [`embed_batch`](Self::embed_batch) — *document*
+    /// semantics — so leaving it at default while overriding the single-query path
+    /// silently produces unprefixed batch-query embeddings (wrong vector space).
```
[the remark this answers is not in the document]
dutiona: Done in 7dcc6c7 — doc comment rewritten: default loops `embed_query` for correctness-by-default; native-batch providers override for a single round-trip.

src/traits.rs
```diff
@@ -802,6 +848,86 @@ mod tests {
...
+            batch_calls: std::sync::atomic::AtomicUsize::new(0),
+        };
+        let out = p.embed_query_batch(&["a", "bb", "ccc"]).unwrap();
+        assert_eq!(out, vec![vec![1.0], vec![2.0], vec![3.0]]);
+        assert_eq!(
+            p.batch_calls.load(std::sync::atomic::Ordering::Relaxed),
+            1,
+            "embed_query_batch must route through embed_batch exactly once"
+        );
+    }
```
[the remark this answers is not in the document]
dutiona: Done in 7dcc6c7 — test now asserts the default routes through the overridden `embed_query` (prefix preserved).

Issue #614, opened 2026-06-16
feat(core): embedding model mismatch detection + MemoryError::EmbeddingModelMismatch


dutiona:
Epic: #610 · Wave 1 (guardrail)
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.2)

Compare provider.fingerprint() vs stored embedding_meta at the embedding boundary (ingest + query). Hard error on mismatch (never panic, never silent). Optional eager check at MCP startup.

Issue #610, opened 2026-06-16
epic(embedding): model-identity policy + TEI/Qwen migration


dutiona:
Umbrella for the embedding model-identity policy + migration from all-MiniLM (Ollama, 384-dim) to Qwen3-Embedding-0.6B served via TEI.

Design spec: `docs/design/embedding-identity-and-tei-qwen-migration.md`
Cross-layer parity policy: `docs/design/adr/0015-cross-layer-embedding-identity-policy.md` (mirrored in dutiona/knowledge-base).

**Waves**
- **Wave 0** — premise-validation spike (gate before Wave 1).
- **Wave 1** — identity fingerprint -> mismatch detection -> pre-computed policy -> embed_query -> TEI/Qwen provider -> MCP/CLI wiring -> shared ADR -> gated smoke test. Guardrail sub-issues land before the swap.
- **Wave 2 (future)** — multi-embedding registry -> background reconstruction -> HNSW rebuild + similarity invalidation -> real-time concurrency-safety research.

KB parity (mismatch detection + ADR mirror + field audit) tracked separately in dutiona/knowledge-base.

dutiona:
Cross-layer KB parity track filed in dutiona/knowledge-base: #468 (mismatch detection), #469 (ADR mirror), #470 (field audit). Per shared ADR 0015.

dutiona:
**Wave 2 (#816)** relocates this epic's entry point: `types` + `engine` + `embed` → `me-types` + `embed`. Map + design + plan: https://github.com/dutiona/memory-engine/issues/925 . (Directional hint from the §13.2 routing table — confirm/refine on resume.)

PR #645, opened 2026-06-19
enhancement(storage): defer embedding_meta stamping until a vector is written

### LINKED ISSUE #643: enhancement(storage): defer embedding_meta stamping until a vector is actually written
**Discovered in:** PR #642 (#613). Flagged by both the one-shot Codex super-review and the `chatgpt-codex-connector[bot]` auto-review (P2).

## Problem
`record_embedding_identity` is called at the **start** of an embed-and-persist operation, before it is known whether a vector will actually be committed:
- `src/engine/bootstrap.rs` (all three bootstrap wrappers) — records before the inner import, which may import zero files.
- `src/consolidation/mod.rs:~92` — records at the top of the transaction, before clustering may produce zero summaries.

A **no-op** bootstrap/consolidate run on an empty store therefore permanently records the configured embedder's identity even though no vector was written. Because `record_if_absent` is write-once, a later real first write with a *different* embedder would inherit the earlier (stale) identity.

## Why it was accepted for #613
Within a consistent single-embedder deployment this is correct (the recorded identity is the configured embedder's). The cross-embedder "no-op-then-mismatched-ingest" staleness is exactly what **#614 enforcement** catches (a differing later fingerprint → hard reject). #613 ships no enforcement, so record-early vs record-late yield the same first-writer-wins identity. `agy` independently judged the current behavior fine.

## Proposed refinement (for the #614 era)
Record identity as close to the actual vector write as possible:
- Transactional paths (consolidate, `bootstrap_session` savepoint, `add_facts_batch`) can record inside the same txn *after* confirming a vector is written, with no loss of atomicity.
- **Constraint:** the `bootstrap_memory_directory` path is autocommit-per-file, so it MUST keep meta-first ordering (recording before the first file) for crash-safety — it cannot defer without reintroducing the orphan-vector window. Resolve this tension explicitly.

Epic #610, Wave 1 follow-up.

dutiona:
## Summary

Records the embedding identity (`embedding_meta`, ADR 0015 §2) **as close to the actual vector write as possible**, instead of at the *start* of an embed-and-persist operation. Closes #643 (Epic #610, Wave 1 follow-up; flagged by PR #642's Codex super-review and `chatgpt-codex-connector[bot]`).

## Problem

`record_if_absent` is **write-once**. Two paths stamped the configured embedder's identity *before* it was known whether a vector would commit:

- `consolidate` — at the top of the transaction, before clustering may produce zero summaries.
- `bootstrap_session` / `bootstrap_directory` — before the inner import, which may extract zero facts.

A **no-op** run (dedup-only consolidation, empty/no-op bootstrap) therefore permanently fixed a *stale* identity. Under #614 enforcement, a later real first write with a **different** embedder would inherit it and be hard-rejected — silent corruption.

## Fix — record where the vector lands

| Path | Change |
| --- | --- |
| `consolidate` | Stamp moved **after** `cluster_fusion`/`global_integration`, gated on `clusters_created > 0 \|\| global_summaries > 0` (the only summary-vector writers; `local_dedup` merely expires). Still inside the same `tx` → atomicity unchanged. |
| `bootstrap_session` / `bootstrap_directory` | Stamp moved into `bootstrap_within_savepoint`, gated on `facts_created > 0`. One chokepoint fixes both wrappers and **preserves per-session independence** (each session commits its identity atomically with its facts). |
| `bootstrap_memory_directory` | **Kept meta-first** (the constraint #643 calls out explicitly). It is autocommit-per-file; deferring would reopen the orphan-vector crash window. The harmless early stamp is retained *only* here, as the crash-safe choice. |
| `add_fact` / `add_facts_batch` | **Unchanged** — both already record inside their txn/savepoint and cannot no-op (empty batch returns early; a single add always inserts), so a vector is always written. |

The tension between deferred-in-transaction and meta-first-in-autocommit is resolved explicitly, per the issue's request.

## Tests (TDD — RED first)

- `dedup_only_consolidation_does_not_stamp_identity` — vector-less consolidation leaves the store unstamped; a later first writer with a different embedder wins.
- `summary_writing_consolidation_stamps_identity` — positive guard against over-deferral.
- `noop_bootstrap_does_not_stamp_identity` — fact-less bootstrap leaves the store unstamped.
- `noop_bootstrap_then_real_write_records_real_embedder` — the literal #614-era harm: `model-b` (real write) wins over the no-op bootstrap's `mock`.
- `bootstrap_creating_facts_stamps_identity` — positive guard.

## Verification

- `cargo test --workspace` — 1183 passed, 4 ignored
- `cargo test --all-features --lib` — 863 passed
- `cargo clippy --workspace --all-targets` — clean
- `cargo fmt --check` — clean

### DIFF
--- a/src/consolidation/mod.rs
+++ b/src/consolidation/mod.rs
@@ -84,13 +84,6 @@ pub fn consolidate(
 
     let tx = conn.unchecked_transaction()?;
 
-    // Record the embedding identity on first write (#613, ADR 0015 §2), inside the
-    // transaction so it commits atomically with any summary vectors. Consolidation
-    // embeds summary text via `embedder` into the same vector space, so it is a
-    // legitimate first-write path; `record_if_absent` is a no-op once an identity
-    // exists (the usual case — facts were ingested before consolidation runs).
-    crate::store::embedding_meta::record_if_absent(&tx, &embedder.fingerprint(), embed_dim)?;
-
     let (duplicates_removed, expired_ids) =
         local_dedup(&tx, embed_dim, config.dedup_threshold, last, now)?;
 
@@ -101,6 +94,19 @@ pub fn consolidate(
     let clusters_created =
         cluster_fusion(&tx, generator, embedder, embed_dim, config.min_cluster_size)?;
     let global_summaries = global_integration(&tx, generator, embedder, embed_dim)?;
+
+    // Record the embedding identity on first write (#613, ADR 0015 §2) — but only
+    // once a summary vector has actually been written (#643). `local_dedup` only
+    // expires rows; `cluster_fusion`/`global_integration` are the sole vector
+    // writers here, so gating on their counts defers the stamp past a vector-less
+    // run (dedup-only, or a no-op on a sparse store). Done inside `tx`, so the
+    // identity still commits atomically with the summaries it describes. A no-op
+    // run leaves the store unstamped, so a later real first write with a different
+    // embedder establishes the true identity instead of inheriting a stale one
+    // (the #614-enforcement landmine).
+    if clusters_created > 0 || global_summaries > 0 {
+        crate::store::embedding_meta::record_if_absent(&tx, &embedder.fingerprint(), embed_dim)?;
+    }
 
     // Only advance the watermark if dedup actually ran. When skipped, facts
     // ingested during the over-cap period must be retried on the next run.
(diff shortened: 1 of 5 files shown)


src/consolidation/mod.rs
```diff
@@ -102,6 +95,19 @@ pub fn consolidate(
...
+    // Record the embedding identity on first write (#613, ADR 0015 §2) — but only
+    // once a summary vector has actually been written (#643). `local_dedup` only
+    // expires rows; `cluster_fusion`/`global_integration` are the sole vector
+    // writers here, so gating on their counts defers the stamp past a vector-less
+    // run (dedup-only, or a no-op on a sparse store). Done inside `tx`, so the
+    // identity still commits atomically with the summaries it describes. A no-op
+    // run leaves the store unstamped, so a later real first write with a different
+    // embedder establishes the true identity instead of inheriting a stale one
+    // (the #614-enforcement landmine).
+    if clusters_created > 0 || global_summaries > 0 {
```
dutiona, line 107: [Antigravity Review]
[LOW] Excellent deferral gate here. Using `clusters_created > 0 || global_summaries > 0` precisely captures only the vector-writing paths (cluster fusion and global integration) while gracefully skipping dedup-only runs, effectively closing the #614-enforcement landmine.

src/engine/bootstrap.rs
```diff
@@ -139,10 +140,14 @@ impl MemoryEngine {
...
-        // any file's fact, so no vector is ever committed without its identity.
+        // Stamp the embedding identity on first write (#613) — and, unlike the
+        // savepoint paths above, this one MUST stay meta-first (#643). It is
+        // autocommit-per-file (no wrapping savepoint), so each file commits its
+        // vector independently; deferring the stamp until after a vector is written
+        // would reopen the orphan-vector crash window (a vector committed before its
+        // identity). Recording before the first file keeps a crash benign: identity
+        // declared, possibly no facts — the same harmless no-op-stamp #643 removes
+        // from the deferrable paths, retained here because it is the crash-safe choice.
         self.record_embedding_identity(&conn, embedder)?;
```
dutiona, line 151: [Antigravity Review]
[LOW] Retaining the meta-first approach for `bootstrap_memory_directory` is the correct decision. Since it relies on autocommit-per-file without an overarching savepoint, deferring the stamp would create a dangerous orphan-vector crash window. Trading a harmless no-op stamp for explicit crash safety is perfectly justified.

dutiona (commented):
[Antigravity Review] LGTM

dutiona (commented):
[Codex Review] LGTM. I found no predicate or atomicity regressions in the deferred stamping paths. GitHub blocked approval because the PR author cannot approve their own pull request.

PR #646, opened 2026-06-19
feat(retrieval): HttpEmbeddingProvider — query_instruction + MRL truncate/renormalize + fingerprint

### LINKED ISSUE #617: feat(retrieval): HttpEmbeddingProvider — query_instruction + MRL truncate/renormalize + fingerprint

dutiona:
## What

Extends `HttpEmbeddingProvider` for **TEI/Qwen asymmetric retrieval** (§Design.5 of the embedding-identity spec), via chainable builders that keep `new()` backward-compatible:

```rust
HttpEmbeddingProvider::new(endpoint, "Qwen/Qwen3-Embedding-0.6B", "tei", None, 1024, 30)?
    .with_query_instruction(HttpEmbeddingProvider::DEFAULT_QUERY_INSTRUCTION)
    .with_mrl_dim(256)
```

- **`with_query_instruction(prefix)`** — `embed_query` / `embed_query_batch` prepend it; the document `embed` / `embed_batch` path stays prefix-free. `DEFAULT_QUERY_INSTRUCTION` exposes Qwen's default for the consumer config layer (#618).
- **`with_mrl_dim(target)`** — Matryoshka truncation: each returned vector is sliced to `target` and **L2-renormalized**. `expected_dim` stays the **native** dim (validated against the raw response); the stored/reported dim is the truncated `target`. A zero/non-finite truncated-prefix norm is **hard-rejected**, not renormalized into NaN.
- **`fingerprint()`** → `with_matryoshka(model, provider, mrl_dim, native_dim)` when truncating, else the plain native-dim identity.

## Why backward-compatible builders (not new params)

Adding 2 positional params to `new()` would break all ~22 `HttpEmbeddingProvider::new` call sites *and* trip clippy's `too_many_arguments` (8 args). Chainable `with_*` builders default to `None`, so symmetric providers and every existing call site are untouched — same minimal-blast-radius philosophy as #616's default trait methods.

## Design notes

- **`embed_query` is prefix-and-delegate** over `embed`, so MRL (living in `embed`/`embed_batch`) applies to **both** document and query paths with zero duplication.
- **`expected_dim` = native dim** (validates the raw server response) when MRL is active; the engine's stored `dim` = `mrl_dim`. This keeps `validate_embedding`'s existing semantics (validate the wire response) intact.

## Tests

- **8 unit** (http.rs): MRL truncate/renorm correctness, `mrl_dim > len` error, zero-norm-prefix error, asymmetric vs plain fingerprint, default-instruction shape.
- **5 wiremock end-to-end** (`tests/asymmetric_embedding.rs`): asserts the **prefix actually goes out on the wire** (`received_requests` body inspection) for query (single + batch), documents stay prefix-free, MRL truncates+renormalizes a real response, and the combined prefix+MRL path.

## Verification

- `cargo build --workspace` ✅
- `cargo test --workspace` ✅ 1201 passed
- `cargo fmt --check` ✅ clean

Wave 1 of epic #610. Unblocks #618 (MCP query path + config wiring).

Fixes #617

### DIFF
--- a/memory-engine-embed/src/http.rs
+++ b/memory-engine-embed/src/http.rs
@@ -49,7 +49,8 @@ use memory_engine::traits::EmbeddingProvider;
 /// )
 /// .expect("failed to build HTTP client")
 /// .with_query_instruction(HttpEmbeddingProvider::DEFAULT_QUERY_INSTRUCTION)
-/// .with_mrl_dim(256);
+/// .with_mrl_dim(256)
+/// .expect("mrl_dim must be within the native dimension");
 /// ```
 pub struct HttpEmbeddingProvider {
     client: reqwest::blocking::Client,
@@ -125,47 +126,66 @@ impl HttpEmbeddingProvider {
     /// server returns and is validated against the raw response; `mrl_dim` is the
     /// stored/reported dimension after truncation. Chainable; defaults to `None`.
     ///
-    /// Truncation correctness (`mrl_dim <= response length`, non-degenerate norm) is
-    /// enforced at embed time, not here — the authoritative length is the live server
-    /// response, not the `expected_dim` hint.
-    #[must_use]
-    pub const fn with_mrl_dim(mut self, mrl_dim: usize) -> Self {
+    /// # Errors
+    ///
+    /// Returns an error if `mrl_dim` is `0` or greater than `expected_dim`. Because the
+    /// raw response is validated to be exactly `expected_dim` long *before* truncation,
+    /// `mrl_dim > expected_dim` is a deterministic configuration contradiction that
+    /// could never truncate successfully — so it is rejected here at build time rather
+    /// than failing every `embed` call at runtime. The non-degenerate-norm check still
+    /// happens at embed time (it depends on the live response values, not the config).
+    pub fn with_mrl_dim(mut self, mrl_dim: usize) -> Result<Self, MemoryError> {
+        if mrl_dim == 0 || mrl_dim > self.expected_dim {
+            return Err(MemoryError::Internal(format!(
+                "invalid mrl_dim {mrl_dim}: must be in 1..={} (the native expected_dim)",
+                self.expected_dim
+            )));
+        }
         self.mrl_dim = Some(mrl_dim);
-        self
+        Ok(self)
     }
 
-    /// Truncate `emb` to its first `target` components and renormalize to unit L2 length
-    /// (Matryoshka representation learning).
+    /// Truncate `emb` in place to its first `target` components and renormalize to unit
+    /// L2 length (Matryoshka representation learning), reusing the input buffer.
+    ///
+    /// The L2 norm is accumulated in `f64` to avoid `f32` overflow-to-`inf` or
+    /// underflow-to-`0` false positives on extreme-magnitude prefixes before the
+    /// non-finite/zero guard runs.
     ///
     /// # Errors
     ///
     /// - `target > emb.len()` — the truncation target exceeds the vector the server returned.
     /// - the truncated prefix has a zero or non-finite L2 norm — renormalization would
     ///   produce NaN/Inf, so the vector is rejected rather than silently corrupted.
-    fn mrl_truncate(emb: &[f32], target: usize) -> Result<Vec<f32>, MemoryError> {
+    #[allow(clippy::cast_possible_truncation)] // deliberate f64-norm → f32-storage narrowing
+    fn mrl_truncate(mut emb: Vec<f32>, target: usize) -> Result<Vec<f32>, MemoryError> {
         if target > emb.len() {
             return Err(MemoryError::Internal(format!(
                 "mrl_dim {target} exceeds embedding dimension {}",
                 emb.len()
             )));
         }
-        let mut truncated: Vec<f32> = emb[..target].to_vec();
-        let norm = truncated.iter().map(|x| x * x).sum::<f32>().sqrt();
+        emb.truncate(target);
+        let norm = emb
+            .iter()
+            .map(|&x| f64::from(x) * f64::from(x))
+            .sum::<f64>()
+            .sqrt();
         if !norm.is_finite() || norm == 0.0 {
             return Err(MemoryError::Internal(
                 "mrl truncation: prefix has zero or non-finite L2 norm; cannot renormalize".into(),
             ));
         }
-        for x in &mut truncated {
-            *x /= norm;
+        for x in &mut emb {
+            *x = (f64::from(*x) / norm) as f32;
         }
-        Ok(truncated)
+        Ok(emb)
     }
 
     /// Apply MRL truncation if configured; otherwise pass the vector through unchanged.
     fn maybe_truncate(&self, emb: Vec<f32>) -> Result<Vec<f32>, MemoryError> {
         match self.mrl_dim {
-            Some(target) => Self::mrl_truncate(&emb, target),
+            Some(target) => Self::mrl_truncate(emb, target),
             None => Ok(emb),
         }
     }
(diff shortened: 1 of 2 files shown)


memory-engine-embed/src/http.rs
```diff
@@ -63,9 +96,80 @@ impl HttpEmbeddingProvider {
...
+    ///   produce NaN/Inf, so the vector is rejected rather than silently corrupted.
+    fn mrl_truncate(emb: &[f32], target: usize) -> Result<Vec<f32>, MemoryError> {
+        if target > emb.len() {
+            return Err(MemoryError::Internal(format!(
+                "mrl_dim {target} exceeds embedding dimension {}",
+                emb.len()
+            )));
+        }
+        let mut truncated: Vec<f32> = emb[..target].to_vec();
+        let norm = truncated.iter().map(|x| x * x).sum::<f32>().sqrt();
```
dutiona: [Antigravity Review] Accumulating a sum-of-squares in `f32` precision for vectors can lead to catastrophic absorption for high-dimensional embeddings, and risks underflowing to 0 if the vector components are very small (unnecessarily triggering the `norm == 0.0` rejection). Consider calculating the norm directly on `emb[..target]` using an `f64` accumulator before allocating the truncated vector. For example:
```rust
let norm = emb[..target].iter().map(|&x| f64::from(x) * f64::from(x)).sum::<f64>().sqrt() as f32;
```

memory-engine-embed/src/http.rs
```diff
@@ -63,9 +96,80 @@ impl HttpEmbeddingProvider {
...
+
+    /// Truncate `emb` to its first `target` components and renormalize to unit L2 length
+    /// (Matryoshka representation learning).
+    ///
+    /// # Errors
+    ///
+    /// - `target > emb.len()` — the truncation target exceeds the vector the server returned.
+    /// - the truncated prefix has a zero or non-finite L2 norm — renormalization would
+    ///   produce NaN/Inf, so the vector is rejected rather than silently corrupted.
+    fn mrl_truncate(emb: &[f32], target: usize) -> Result<Vec<f32>, MemoryError> {
```
dutiona: [Antigravity Review] Efficiency nit: `maybe_truncate` takes `emb: Vec<f32>` by value, but `mrl_truncate` takes a slice `&[f32]`, forcing a redundant allocation in `emb[..target].to_vec()`. You could change `mrl_truncate` to accept `mut emb: Vec<f32>`, call `emb.truncate(target);`, and do the renormalization in-place to save an allocation per embedding.

memory-engine-embed/tests/asymmetric_embedding.rs
```diff
@@ -0,0 +1,181 @@
...
+    .await
+    .unwrap();
+
+    // Prefix went out on the wire...
+    assert_eq!(
+        sole_request_input(&server).await,
+        json!("Instruct: find this")
+    );
+    // ...and the response was MRL-truncated + renormalized.
+    assert_eq!(out, vec![0.6, 0.8]);
```
dutiona: [Antigravity Review] Strict equality `assert_eq!` for floating-point calculations is brittle and can be flaky across platforms or toolchains. Consider using an epsilon check as you did earlier in `mrl_truncation_end_to_end` with `(out[i] - val).abs() < 1e-6`.

memory-engine-embed/src/http.rs
```diff
@@ -63,9 +96,80 @@ impl HttpEmbeddingProvider {
...
+    ///
+    /// `expected_dim` (the constructor argument) remains the **native** dimension the
+    /// server returns and is validated against the raw response; `mrl_dim` is the
+    /// stored/reported dimension after truncation. Chainable; defaults to `None`.
+    ///
+    /// Truncation correctness (`mrl_dim <= response length`, non-degenerate norm) is
+    /// enforced at embed time, not here — the authoritative length is the live server
+    /// response, not the `expected_dim` hint.
+    #[must_use]
+    pub const fn with_mrl_dim(mut self, mrl_dim: usize) -> Self {
```
dutiona: [Codex Review] [MEDIUM] `with_mrl_dim` accepts values larger than `expected_dim`, but `validate_embedding` later enforces `emb.len() == expected_dim` before truncation. That means a provider configured with `mrl_dim > expected_dim` can never succeed: it builds, fingerprints as `dim = mrl_dim`, and only fails on the first embed call. This is a deterministic configuration contradiction, so it should be rejected when the builder is configured (or at least covered by a regression test).

memory-engine-embed/src/http.rs
dutiona: [Codex Review] [LOW] `mrl_truncate` computes the L2 norm entirely in `f32`. For finite but extreme prefixes, `x * x` can overflow to `inf` or underflow toward `0.0` before the guard runs, so the provider rejects vectors that are still mathematically normalizable. Accumulating in `f64` (or using a stable norm routine) would avoid these false positives; a regression test with large or tiny finite values would lock it in.

dutiona (commented):
[Codex Review] Two findings. First, `with_mrl_dim` accepts impossible `mrl_dim > expected_dim` configurations and only fails on the first embed call, even though `validate_embedding` makes that contradiction deterministic. Second, `mrl_truncate` accumulates the L2 norm in `f32`, so finite extreme prefixes can overflow/underflow into false zero/non-finite rejections. Asymmetric query prefixing, raw-before-truncate validation ordering, and fingerprint dim/base_dim semantics otherwise look sound.

memory-engine-embed/src/http.rs
dutiona: Fixed in 1a66b6b — with_mrl_dim now returns Result and rejects mrl_dim==0 or mrl_dim>expected_dim at build time, with regression tests (reject 0, reject >native, accept ==native).

memory-engine-embed/src/http.rs
dutiona: Fixed in 1a66b6b — L2 norm now accumulated in f64 before narrowing back to f32, avoiding overflow/underflow false positives.

dutiona: Fixed in 1a66b6b — f64 accumulator for the norm.

memory-engine-embed/src/http.rs
dutiona: Fixed in 1a66b6b — mrl_truncate now takes Vec<f32> by value and truncates in place (emb.truncate), saving the per-embedding allocation.

memory-engine-embed/tests/asymmetric_embedding.rs
dutiona: Fixed in 1a66b6b — epsilon comparisons replace exact float assert_eq in the affected tests.

memory-engine-embed/src/http.rs
```diff
@@ -63,9 +96,80 @@ impl HttpEmbeddingProvider {
...
+        Ok(truncated)
+    }
+
+    /// Apply MRL truncation if configured; otherwise pass the vector through unchanged.
+    fn maybe_truncate(&self, emb: Vec<f32>) -> Result<Vec<f32>, MemoryError> {
+        match self.mrl_dim {
+            Some(target) => Self::mrl_truncate(&emb, target),
+            None => Ok(emb),
+        }
+    }
```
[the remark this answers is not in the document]
dutiona: Applied in 1a66b6b — in-place truncate as suggested; norm accumulated in f64 (stronger than the f32 in the snippet) per the other reviewers.

memory-engine-embed/src/http.rs
```diff
@@ -560,4 +708,88 @@ mod tests {
...
+    #[test]
+    fn mrl_truncate_zero_norm_prefix_errors() {
+        // The first two components are zero -> truncated prefix norm is 0 -> reject
+        // rather than divide-by-zero into NaN.
+        let err = HttpEmbeddingProvider::mrl_truncate(&[0.0, 0.0, 1.0], 2).unwrap_err();
+        assert!(
+            err.to_string().contains("zero or non-finite L2 norm"),
+            "error: {err}"
+        );
+    }
```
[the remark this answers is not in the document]
dutiona: Applied in 1a66b6b — tests pass Vec by value and use epsilon comparisons.

dutiona (commented):
[Codex Review] LGTM. The round-2 delta resolves both prior findings: impossible `mrl_dim` configurations are now rejected at build time with regression coverage, and MRL renormalization now accumulates the norm in `f64` while reusing the input buffer in place. The targeted `with_mrl_dim`, `mrl_truncate`, and asymmetric embedding tests all pass.

PR #647, opened 2026-06-19
feat(mcp): wire query path to embed_query + embedding identity/asymmetry config

### LINKED ISSUE #618: feat(mcp): wire query path to embed_query + embedding config

dutiona:
## What

§Design.6 of the embedding-identity spec — wires the MCP to the asymmetric provider (#617). Two parts:

**1. Query path → `embed_query`** ([tools/mod.rs](memory-engine-mcp/src/tools/mod.rs)): `handle_query` now calls `embed_query` instead of `embed`, so asymmetric models (Qwen) get their query instruction prefix. `memory_add_fact` is unchanged — it passes the provider to `engine.add_fact`, which calls the document `embed()` internally, so documents stay prefix-free.

**2. Config** ([config.rs](memory-engine-mcp/src/config.rs) + [main.rs](memory-engine-mcp/src/main.rs)): `EmbeddingSection` gains:
- `provider` (default `"ollama"`) — feeds the #614 fingerprint; **replaces the hardcoded `"ollama"` `TODO(#618)`** in `build_embedder`. Also a `--embed-provider` CLI override.
- `query_instruction` — the query-only prefix.
- `mrl_dim` — Matryoshka truncation target.

## The MRL dim model

The subtle part. With MRL truncation, two dimensions diverge:
- **`dimensions`** = the **native** dim the model emits → the provider's `expected_dim` (validates the raw response).
- **`mrl_dim`** = the **stored** dim after truncation → must equal the engine's `embed_dim` (the engine stores post-truncation vectors).

`build_embedder` enforces `mrl_dim == embed_dim` **loudly at startup** rather than letting every embed call fail at runtime. Without MRL, native == stored == `embed_dim` (unchanged behavior).

## Tests

- **2 wiremock e2e** ([embedding_integration.rs](memory-engine-mcp/tests/embedding_integration.rs)): a `memory_query` request carries the `Q-PREFIX: ` instruction (asserted via the captured request body); a `memory_add_fact` request is prefix-free — proving the asymmetry end-to-end through MCP dispatch.
- **2 config-parse** ([config_merge.rs](memory-engine-mcp/tests/config_merge.rs)): legacy config defaults `provider` to `"ollama"`; full TEI/Qwen asymmetric config parses.

## Verification

- `cargo build --workspace` ✅
- `cargo test --workspace` ✅ 1208 passed
- `cargo fmt --check` ✅ clean

Wave 1 of epic #610. Unblocks #614 (the MCP eager mismatch check consumes this config + provider identity).

Fixes #618

### DIFF
--- a/memory-engine-mcp/src/main.rs
+++ b/memory-engine-mcp/src/main.rs
@@ -32,7 +32,8 @@ struct Cli {
     #[arg(long, env = "MEMORY_MCP_EMBED_API_KEY")]
     embed_api_key: Option<String>,
 
-    /// Embedding serving backend (`ollama`, `tei`, `openai`) — feeds the fingerprint.
+    /// Embedding serving backend (e.g. `ollama`, `tei`, `openai`) — operator-declared,
+    /// feeds the fingerprint. Free-form; unrecognized values warn but are accepted.
     #[arg(long, env = "MEMORY_MCP_EMBED_PROVIDER")]
     embed_provider: Option<String>,
 
@@ -100,6 +101,10 @@ async fn main() -> Result<(), BoxError> {
 ///
 /// CLI flags override TOML values (endpoint, model, `api_key`) when provided,
 /// enabling operators to inject runtime secrets via env/CLI.
+/// Recognized embedding backends. `provider` is free-form (it feeds the fingerprint),
+/// but a value outside this set warns at startup to catch typos.
+const KNOWN_PROVIDERS: [&str; 3] = ["ollama", "tei", "openai"];
+
 fn build_embedder(
     cli: &Cli,
     mcp_config: &config::McpConfig,
@@ -135,14 +140,25 @@ fn build_embedder(
         return Ok(None);
     };
 
-    // `expected_dim` is the NATIVE dim the model emits (validated against the raw
-    // response). Without MRL it equals the engine's stored `embed_dim`. With MRL the
-    // provider truncates to `mrl_dim`, so the native dim comes from the config's
-    // `dimensions` field while the engine stores (and probes) the truncated `mrl_dim`.
-    let native_dim = match mrl_dim {
-        Some(_) => base.map_or(embed_dim, |b| b.dimensions),
-        None => embed_dim,
-    };
+    // `native_dim` is the dimension the model emits — what the provider validates the
+    // raw HTTP response against. With MRL the provider truncates to `mrl_dim`, so the
+    // native dim comes from the config's `dimensions`; without MRL native == stored ==
+    // `embed_dim`. Always read `dimensions` (falling back to `embed_dim` only on the
+    // CLI-only path with no `[embedding]` section) so a `dimensions != embed_dim`
+    // misconfig surfaces at startup instead of failing cryptically on the first embed.
+    let native_dim = base.map_or(embed_dim, |b| b.dimensions);
+
+    // `provider` is stamped verbatim into the persisted fingerprint (#614), like the
+    // equally free-form `model`. We don't hard-restrict it — custom OpenAI-compatible
+    // backends are valid — but warn on an unrecognized value to catch typos before they
+    // bake a bogus identity into a fresh store.
+    if !KNOWN_PROVIDERS.contains(&provider.as_str()) {
+        tracing::warn!(
+            provider = %provider,
+            "embedding provider is not one of {KNOWN_PROVIDERS:?}; it is stamped into the \
+             persisted embedding fingerprint as-is — check for a typo"
+        );
+    }
 
     let mut provider =
         embedding::HttpEmbeddingProvider::new(url, mdl, provider, api_key, native_dim, timeout)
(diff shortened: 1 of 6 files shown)


memory-engine-mcp/src/main.rs
```diff
@@ -116,25 +120,54 @@ fn build_embedder(
...
-                timeout,
+    let (Some(url), Some(mdl)) = (endpoint, model) else {
+        return Ok(None);
+    };
+
+    // `expected_dim` is the NATIVE dim the model emits (validated against the raw
+    // response). Without MRL it equals the engine's stored `embed_dim`. With MRL the
+    // provider truncates to `mrl_dim`, so the native dim comes from the config's
+    // `dimensions` field while the engine stores (and probes) the truncated `mrl_dim`.
+    let native_dim = match mrl_dim {
```
dutiona: [Antigravity Review]
If MRL is inactive (`mrl_dim` is `None`), this silently ignores the `dimensions` field in `[embedding]` and forces `native_dim` to `embed_dim`. 

This prevents us from failing loudly at startup if the user misconfigures `dimensions != embed_dim` when MRL is disabled. Instead, it will fail at runtime on the first embed call because the model will return a vector of size `dimensions`, but the provider was instantiated expecting `embed_dim`.

Consider always using `base.dimensions` as the native dimension and explicitly validating it against `embed_dim` at startup:

```rust
let native_dim = base.map_or(embed_dim, |b| b.dimensions);

// ... provider init ...

if let Some(target) = mrl_dim {
    if target != embed_dim {
        return Err(format!("embedding.mrl_dim ({target}) must equal the engine embed_dim ({embed_dim})...").into());
    }
} else {
    if native_dim != embed_dim {
        return Err(format!("embedding.dimensions ({native_dim}) must equal engine embed_dim ({embed_dim}) when MRL is disabled").into());
    }
}
```

memory-engine-mcp/src/main.rs
```diff
@@ -116,25 +120,54 @@ fn build_embedder(
         .embed_api_key
         .clone()
         .or_else(|| base.and_then(|b| b.api_key.clone()));
+    // `provider` feeds the fingerprint (#614); source it from CLI > TOML, defaulting to
+    // "ollama" (also EmbeddingSection's serde default) for the CLI-only / legacy path.
+    let provider = cli
```
dutiona, line 130: [Codex Review] `provider` now becomes part of the persisted embedding fingerprint, but this path still accepts any arbitrary string from CLI/TOML. A typo like `--embed-provider tel` or `provider = "teii"` will start cleanly, stamp a bogus fingerprint into a fresh DB, and later trigger a hard mismatch once the operator fixes the spelling. Since the CLI/help and config docs describe a closed set (`ollama|tei|openai`), this should be validated at parse/startup time (for example with `clap::ValueEnum` plus a serde enum/custom deserializer) instead of being treated as free-form text.

dutiona (commented):
[Codex Review] One medium-severity issue: `embed_provider` / `embedding.provider` is now part of the persisted embedding identity, but it is still unvalidated free-form text. A typo can stamp a bogus fingerprint into a fresh store and only surface later as a model-mismatch failure once the spelling is corrected. The query/document asymmetry wiring and the `mrl_dim == embed_dim` startup guard look correct otherwise; `cargo test -p memory-engine-mcp --test config_merge --test embedding_integration` passed locally.

memory-engine-mcp/src/main.rs
dutiona: Fixed in 3b754d9 — build_embedder now always reads `dimensions` as the native dim and validates `dimensions == embed_dim` at startup when MRL is disabled. The mrl_dim<=dimensions bound is enforced by HttpEmbeddingProvider::with_mrl_dim (mrl_dim <= expected_dim==native), and mrl_dim==embed_dim is checked too.

memory-engine-mcp/src/main.rs
```diff
@@ -116,25 +120,54 @@ fn build_embedder(
...
+        if target != embed_dim {
+            return Err(format!(
+                "embedding.mrl_dim ({target}) must equal the engine embed_dim ({embed_dim}): \
+                 the engine stores post-truncation vectors"
             )
-            .map_err(|e| format!("failed to create embedding provider: {e}"))?,
-        ))),
-        _ => Ok(None),
+            .into());
+        }
```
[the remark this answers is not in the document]
dutiona: Done in 3b754d9. All three checks covered: (1) mrl_dim<=dimensions via with_mrl_dim (validates target<=native expected_dim); (2) dimensions==embed_dim when mrl_dim unset (new else-branch); (3) mrl_dim==embed_dim (existing). Fail-loud-at-startup as suggested.

memory-engine-mcp/src/main.rs
dutiona: Addressed in 3b754d9 with a startup WARNING (not a hard reject) + softened the docs/CLI that implied a closed set. Rationale: `provider` is an operator-declared identity slug exactly like `model` — both free-form, both feed the fingerprint (types.rs documents provider as "e.g." values). Hard-restricting provider but not model would be inconsistent and would block valid OpenAI-compatible backends (vllm, infinity). The warning catches the typo case you flagged; #614 mismatch detection is the real safety net for a changed identity. Open to hard-reject if you think the closed set should be authoritative — let me know.

Issue #613, opened 2026-06-16
feat(storage): persist embedding_meta, replace embed_dim; schema_version bump


dutiona:
Epic: #610 · Wave 1
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.1)

Store the identity tuple {model,provider,dim,matryoshka_base_dim,element_type} in DB meta, written on first embedding write. Replace bare embed_dim. Schema bump, no data migration (no users).

dutiona:
# Plan — #613 Persist `embedding_meta`, replace `embed_dim`; schema bump 11→12

**Issue:** #613 (Epic #610, Wave 1). **Branch/worktree:** `feat-613-embedding-meta` at
`/home/mroynard/dev/memory-engine/.worktrees/feat-613-embedding-meta` (off `main` `fe8d139`).
**Reviewers:** advisor + clean-slate subagent (mandatory) → Codex + agy (full loop). **Gemini decommissioned — not used.**

> **Synthesis provenance.** Structure (typed `store::embedding_meta` boundary, engine seam, unified open-path,
> dump/snapshot richening, seam map) from the **architecture-first** draft; correctness spine (atomicity txn,
> dim-consistency guard, 5-path enumeration, malformed-meta hard error, migration-not-snapshot-guarded caveat)
> from the **risk-first** draft; scope framing + `add_facts_batch` savepoint caution from **mvp-first**.
> Drafts in `docs/plans/613-embedding-meta-persistence-drafts/` (gitignored).

---

## 0. Scope boundary (HARD)

**This PR persists the identity tuple and keeps dim-validation working. It does NOT:**

- enforce model mismatch (`MemoryError::EmbeddingModelMismatch`) — that is **#614**;
- validate caller-declared `model` on pre-computed embeddings — **#615**;
- wire real provider/model config values — **#618** (today's fingerprint comes from whatever
  `EmbeddingProvider::fingerprint()` returns; the hardcoded `"ollama"` provider stays until #618);
- build any multi-space registry — **Wave 2 #622**.

Where a persistence choice could arm a downstream landmine, it is designed against here (atomicity, choke-point
completeness, internal-consistency) — but no enforcement code is written. The deliverable is a **typed seam**
the above issues graft onto without reworking call sites.

## 1. Decisions (BLUF)

[...]

PR #651, opened 2026-06-20
feat(core): embedding model mismatch detection + MemoryError::EmbeddingModelMismatch

### LINKED ISSUE #614: feat(core): embedding model mismatch detection + MemoryError::EmbeddingModelMismatch

dutiona:
## What

§Design.2 of the embedding-identity spec — turns mismatch **enforcement** on. This is the Wave-1 guardrail the #612 (`EmbeddingFingerprint`) and #613 (`embedding_meta` persistence) work was built toward.

- **`MemoryError::EmbeddingModelMismatch { expected, actual }`** — boxed fingerprints (keeps the enum small). Carries the full identity tuple on both sides because *any* field (`model`/`provider`/`dim`/`matryoshka_base_dim`/`element_type`) could be the culprit.
- **Write-boundary enforcement** — `record_if_absent`'s present-branch now compares `candidate == stored` (full-tuple equality, shared `ensure_match` helper) and hard-rejects a difference instead of silently returning the stored tuple. Enforces at *every* embedding write (ingest, batch, consolidate, bootstrap) with **zero call-site change** — exactly as the module was pre-designed for.
- **`check_compatible`** — read-only counterpart for the eager path.
- **`MemoryEngine::verify_embedding_identity(&dyn EmbeddingProvider)`** — engine-level eager check. The query path embeds at the *consumer* and hands the engine a pre-computed vector, so the engine can't fingerprint-check per query; this one-shot check catches a misconfigured provider.
- **MCP startup wiring** — `main.rs` calls it after building the embedder and **refuses to serve on a mismatch** (§Design.2 "optional eager check"), so wrong-vector-space queries fail fast instead of returning silently-wrong results.

## Why it matters

Vector *dimension* alone is insufficient identity. Two different models at the same dim pass the length check and then return vectors from incompatible spaces — "silently wrong retrieval, the worst RAG failure mode" (the spec's words). This PR makes that a hard error, never a panic, never silent.

## Behavior change

`record_if_absent` was previously "write-once, stored wins" (#613). A differing later fingerprint is now **rejected**. The engine test `first_add_fact_records_embedding_meta` is updated accordingly (its second-write assertion flips from silently-unchanged to error-and-unchanged).

## Tests

- `embedding_meta`: `record_if_absent` match-returns-stored / differ-rejects / dim-mismatch; `check_compatible` fresh-ok / match-ok / differ-errs (incl. a matryoshka-only difference).
- `engine`: `first_add_fact_records_embedding_meta` (updated), `verify_embedding_identity_enforces_match` (fresh→ok, match→ok, differ→err).
- All write paths (consolidate/bootstrap) already use the seam — full suite confirms no regression.

> Note: the MCP *startup* wiring (4-line guarded call in `main.rs`) delegates to the engine-tested `verify_embedding_identity`; `main()`'s startup flow isn't unit-testable without refactoring, so it's covered by the engine method's tests + reasoning.

## Verification

- `cargo test --workspace` ✅ (838 lib + all integration suites)
- `cargo test --all-features` ✅ 987 passed
- `cargo clippy --workspace --all-targets` ✅ / `--all-features` ✅ clean
- `cargo fmt --check` ✅ exit 0

Closes the Wave-1 guardrail of epic #610.

Fixes #614

### DIFF
--- a/memory-engine-mcp/src/main.rs
+++ b/memory-engine-mcp/src/main.rs
@@ -81,6 +81,17 @@ async fn main() -> Result<(), BoxError> {
     // Use the resolved embed_dim (from DB probe or config), not the TOML value.
     let embedder = build_embedder(&cli, &mcp_config, embed_dim)?;
 
+    // 4b. Eager embedding-identity check (#614, §Design.2). The query path embeds at
+    // this layer and hands the engine a pre-computed vector, so the engine can't
+    // fingerprint-check per query — verify once at startup that the configured provider
+    // matches the store's recorded identity, and refuse to serve on a mismatch rather
+    // than silently returning wrong-vector-space query results.
+    if let Some(provider) = embedder.as_deref() {
+        engine
+            .verify_embedding_identity(provider)
+            .map_err(|e| format!("embedding identity check failed: {e}"))?;
+    }
+
     // 5. Initialize summary generator (optional — text only; embedding is done
     //    by the embedder, injected separately into consolidation per #116)
     let summary_gen = build_summary_generator(&cli, &mcp_config)?
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -442,6 +442,28 @@ impl MemoryEngine {
         Ok(())
     }
 
+    /// Verify a provider's identity matches the store's recorded embedding identity —
+    /// a fail-fast check for consumer startup (#614, §Design.2).
+    ///
+    /// The query path embeds at the consumer layer and hands the engine a *pre-computed*
+    /// vector, so the engine cannot fingerprint-check per query. This one-shot check
+    /// catches a misconfigured provider — one whose vector space differs from the one
+    /// the store was built with — **before** any query silently returns
+    /// wrong-vector-space results. A fresh store with no recorded identity is compatible
+    /// with any provider (nothing to disagree with).
+    ///
+    /// This is read-only and does not establish the identity (that happens lazily on the
+    /// first embedding *write*, via [`record_embedding_identity`](Self::record_embedding_identity)).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`MemoryError::EmbeddingModelMismatch`] if the provider's fingerprint
+    /// disagrees with the store's recorded identity.
+    pub fn verify_embedding_identity(&self, provider: &dyn EmbeddingProvider) -> Result<()> {
+        let candidate = provider.fingerprint();
+        self.with_read(|conn| crate::store::embedding_meta::check_compatible(conn, &candidate))
+    }
+
     /// Whether this engine was opened in read-only mode.
     #[must_use]
     pub const fn is_read_only(&self) -> bool {
--- a/src/engine/mod.rs
+++ b/src/engine/mod.rs
@@ -449,18 +449,27 @@ impl MemoryEngine {
     /// vector, so the engine cannot fingerprint-check per query. This one-shot check
     /// catches a misconfigured provider — one whose vector space differs from the one
     /// the store was built with — **before** any query silently returns
-    /// wrong-vector-space results. A fresh store with no recorded identity is compatible
-    /// with any provider (nothing to disagree with).
+    /// wrong-vector-space results. On a **fresh** store (no recorded identity) there is
+    /// no model to disagree with, but the provider's `dim` must still match this engine's
+    /// configured dimension — otherwise every subsequent write/query would fail on
+    /// `EmbeddingDimension`, so we fail fast here too (mirroring `record_if_absent`).
     ///
     /// This is read-only and does not establish the identity (that happens lazily on the
     /// first embedding *write*, via [`record_embedding_identity`](Self::record_embedding_identity)).
     ///
     /// # Errors
     ///
-    /// Returns [`MemoryError::EmbeddingModelMismatch`] if the provider's fingerprint
+    /// Returns [`MemoryError::EmbeddingDimension`] if the provider's dimension differs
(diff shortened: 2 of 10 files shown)


memory-engine-mcp/src/main.rs
```diff
@@ -81,6 +81,17 @@ async fn main() -> Result<(), BoxError> {
...
 
+    // 4b. Eager embedding-identity check (#614, §Design.2). The query path embeds at
+    // this layer and hands the engine a pre-computed vector, so the engine can't
+    // fingerprint-check per query — verify once at startup that the configured provider
+    // matches the store's recorded identity, and refuse to serve on a mismatch rather
+    // than silently returning wrong-vector-space query results.
+    if let Some(provider) = embedder.as_deref() {
+        engine
+            .verify_embedding_identity(provider)
+            .map_err(|e| format!("embedding identity check failed: {e}"))?;
```
dutiona, line 92: [Antigravity Review] This eager check placement is perfectly positioned. It not only covers standard startup but effectively covers the `restore_snapshot_into` bypass — because restoring reconstructs the JSON config row exactly, and this eager check safely validates the active provider against the restored identity before any traffic is served.

src/store/embedding_meta.rs
```diff
@@ -85,21 +87,25 @@ pub fn store(conn: &Connection, fp: &EmbeddingFingerprint) -> Result<()> {
...
+/// `candidate` differs from it; `MemoryError::EmbeddingDimension` if no identity is
+/// recorded and `candidate.dim != expected_dim` (nothing is persisted); or any
+/// [`load`] / [`store`] error.
 pub fn record_if_absent(
     conn: &Connection,
     candidate: &EmbeddingFingerprint,
     expected_dim: usize,
 ) -> Result<EmbeddingFingerprint> {
     if let Some(stored) = load(conn)? {
+        ensure_match(&stored, candidate)?;
```
dutiona, line 108: [Codex Review] This hard compare breaks MCPs precomputed `memory_add_fact` path. `memory-engine-mcp/src/tools/mod.rs` still routes caller-supplied embeddings through `engine.add_fact()` via `PassthroughEmbedder`, whose fingerprint is the sentinel `("precomputed", "passthrough", dim)`. Before this PR, `record_if_absent()` ignored that candidate once a real provider had already stamped the store; now it rejects it here, so precomputed adds start failing against any existing stamped store. There is also no regression test covering a precomputed add after a real-provider write.

dutiona (commented):
[Codex Review] I found one high-severity regression. `record_if_absent()` now rejects any differing fingerprint, but MCPs precomputed `memory_add_fact` flow still goes through `engine.add_fact()` with `PassthroughEmbedder` (`precomputed`/`passthrough`). That caller relied on the old “stored wins” behavior, so explicit-embedding adds now fail against any store already stamped by a real provider. Existing MCP coverage only exercises the fresh-store happy path, so this regression is currently untested.

src/store/embedding_meta.rs
dutiona: Fixed in b24b068 — the MCP precomputed path now uses the new add_fact_precomputed (require_present, no sentinel comparison) instead of routing PassthroughEmbedder through record_if_absent. Precomputed add to a stamped store works (dim-checked); on a fresh store it errors (cannot establish identity), consistent with promote/cycle. Tests updated.

src/engine/mod.rs
```diff
@@ -442,6 +442,28 @@ impl MemoryEngine {
...
+    /// first embedding *write*, via [`record_embedding_identity`](Self::record_embedding_identity)).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`MemoryError::EmbeddingModelMismatch`] if the provider's fingerprint
+    /// disagrees with the store's recorded identity.
+    pub fn verify_embedding_identity(&self, provider: &dyn EmbeddingProvider) -> Result<()> {
+        let candidate = provider.fingerprint();
+        self.with_read(|conn| crate::store::embedding_meta::check_compatible(conn, &candidate))
+    }
```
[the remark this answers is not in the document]
dutiona: Applied your suggestion — verify_embedding_identity now guards candidate.dim == embed_dim before check_compatible, failing fast on a wrong-dim provider even on a fresh store.

src/engine/mod.rs
```diff
@@ -442,6 +442,28 @@ impl MemoryEngine {
...
+    /// This is read-only and does not establish the identity (that happens lazily on the
+    /// first embedding *write*, via [`record_embedding_identity`](Self::record_embedding_identity)).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`MemoryError::EmbeddingModelMismatch`] if the provider's fingerprint
+    /// disagrees with the store's recorded identity.
+    pub fn verify_embedding_identity(&self, provider: &dyn EmbeddingProvider) -> Result<()> {
+        let candidate = provider.fingerprint();
+        self.with_read(|conn| crate::store::embedding_meta::check_compatible(conn, &candidate))
```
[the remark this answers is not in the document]
dutiona: Fixed — added the dimension guard to verify_embedding_identity mirroring record_if_absent; fresh-store wrong-dim now returns EmbeddingDimension at the eager check.

memory-engine-mcp/src/main.rs
dutiona: Thanks — eager check placement retained; also added a dim fast-fail per the bots.

Issue #493, opened 2026-06-01
[super-qa] cli: low/info findings (8)


dutiona:
> ⚠️ Agent analysis; not source-verified. Grouped low/info for **area:cli**.

Grouped low/info findings for area **cli** (8):

- **low** `memory-engine-cli/src/commands/dump.rs:53-58` — Bare unreachable!() tightly couples value_parser literal list to match arms (`cli/correctness-dump-unreachable-brittle`)
- **low** `memory-engine-cli/src/db.rs:44-48` — open_engine_writable doc lists an incomplete set of write-requiring commands (`cli/documentation-incomplete-writable-examples`)
- **low** `memory-engine-cli/src/commands/batch_ingest.rs:33-35` — Bearer API key accepted as a plain command-line flag (process-arg / shell-history exposure) (`cli/secret-api-key-in-process-args`)
- **low** `memory-engine-cli/src/output.rs:17-22, memory-engine-cli/src/commands/batch_ingest.rs:108-120` — Strong property-based testing candidates lack both example-based and property tests (`cli/testing-proptest-candidates-no-coverage`)
- **low** `memory-engine-cli/src/embedding.rs:20-28` — PassthroughEmbedder::embed_batch error path for 0 texts is untested (`cli/testing-passthrough-embedder-embed-batch-zero-texts-untested`)
- **low** `memory-engine-cli/src/output.rs:15-22, memory-engine-cli/src/db.rs:32-55` — Public functions with behavioral doc comments lack runnable doc-test examples (`cli/testing-no-doc-tests-on-public-api`)
- **info** `memory-engine-cli/src/commands/import.rs:42-56` — Doc comment claims partial read but serde_json::from_reader deserializes the whole file (`cli/correctness-import-doc-misleading-peek`)
- **info** `memory-engine-cli/src/commands/query.rs:43-47, memory-engine-cli/src/commands/add_fact.rs:72-76` — parse_datetime is duplicated across two modules with no inline unit test in either (`cli/testing-duplicate-parse-datetime-untested`)

Issue #251, opened 2026-06-01
[super-qa] Area: cli — code-quality findings


dutiona:
All `/super-qa` 2026-06-01 findings in **area:cli**. Non-autofixable blocker/critical/high/medium are individual sub-issues; low/info are grouped in one sub-issue.

dutiona:
## Sweep complete — all 19 sub-issues resolved ✅

Every `/super-qa` 2026-06-01 area:cli finding is fixed and merged. Closing the epic.

**Original 17 findings — 6 waved PRs:**
- W0: #427, #305 closed (already-fixed-on-main, caught by triage-vs-main).
- W1 #654: #407 — snapshot-header DoS cap.
- W2 #660: #397 / #408 / #371 / #431 / #432 — JSONL ingest hardening (review caught a silent-data-loss off-by-one).
- W3 #675: #270 (`CliFactType` consolidation) + #430 (query-filter tests).
- W4 #686: #271 / #303 / #304 / #428 / #429 — CLI test coverage.
- W5 #692: #433 (proptest parser fuzzing) + #493 (low/info rollup).

**Collateral found during the sweep:**
- #664 ✅ #704 — read-error loop (CWE-835), `continue`→`break`.
- #663 ✅ **#732** — oversized-content batch-poisoning, fixed via the engine partial-success path (`add_facts_batch_partial` → per-record `Vec<Result<id, err>>`; CLI skips only rejected records). Implemented after #631 (StorageBackend wiring) landed, exactly as the locked design specified. Verified on current main: engine isolation tests green, CLI flush path uses the partial API.

**Out of scope (tracked separately):** #678 (MCP duplicate fact-type parser) — closed standalone via #711. #667 chore — closed-redundant (#655 beat it).

PR #654, opened 2026-06-20
fix(cli): bound snapshot-header parse, stop at embed_dim (#407)

### LINKED ISSUE #407: [super-qa] medium: Unbounded full-document JSON parse of attacker-controlled snapshot to read one field
> ⚠️ Generated by `/super-qa` 2026-06-01. Severity & details from agent analysis (security lens ran via octo tier-2 fallback, xhigh not effort:max); **not independently source-verified**. Review before action.

**Unbounded full-document JSON parse of attacker-controlled snapshot to read one field**

- **ID:** `cli/dos-import-header-unbounded-parse`
- **Severity:** medium
- **Area:** cli · **Type:** security
- **Location:** `memory-engine-cli/src/commands/import.rs:49-55`
- **Auto-fixable:** no

## Description
`peek_embed_dim` wraps the snapshot file in a `BufReader` and calls `serde_json::from_reader::<SnapshotHeader>(reader)` to extract a single `embed_dim: usize`. serde_json does not stop after the known field — it tokenizes and validates the ENTIRE JSON document up to EOF, allocating intermediate `String`/buffer storage for arbitrarily large object keys and values. There is no `.take(N)` byte cap on the reader. The snapshot file is the import source, i.e. fully attacker-controlled input (the trust boundary). This runs only on the plain-JSON path (when `--embed-dim` is omitted), but in that case a multi-GB or pathological header drives unbounded memory allocation before `restore_json` (which the inspect-module audit notes has its own 4 GiB cap) is ever reached. This is a redundant, uncapped pre-parse introduced by the CLI itself, separate from the engine's restore path. serde_json's default 128-deep recursion limit bounds the nesting/stack-overflow vector but NOT total width/bytes. CWE-400, CWE-770.

## Suggested approach
Cap the bytes fed to the header parser, e.g.:

    use std::io::Read;
    const MAX_HEADER_BYTES: u64 = 1 << 20; // 1 MiB is ample for a JSON header object
    let file = std::fs::File::open(path)?;
    let reader = std::io::BufReader::new(file).take(MAX_HEADER_BYTES);
    let header: SnapshotHeader = serde_json::from_reader(reader)
        .map_err(|e| anyhow::anyhow!("failed to read snapshot header (>1MiB or malformed): {e}"))?;

Better still, use `serde_json::Deserializer::from_reader(reader).into_iter::<SnapshotHeader>().next()` over the capped reader so the parser stops at the first complete value instead of requiring clean EOF.

## References
CWE-400, CWE-770

dutiona:
## What & why

`memory-engine-cli import` auto-detects `embed_dim` on the plain-JSON path by peeking the snapshot header. The old code did:

```rust
let header: SnapshotHeader = serde_json::from_reader(reader)?; // reads to EOF
```

`serde_json::from_reader::<SnapshotHeader>` tokenizes the **entire** snapshot document to EOF just to read one leading field. The import file is fully attacker-controlled (the trust boundary), so a multi-GB or pathological snapshot drives **unbounded memory allocation** (CWE-400 / CWE-770) *before* the engine's own restore cap is ever reached. (#407)

## Fix

A streaming **sink visitor** that captures `embed_dim` and stops the instant it is seen, over a reader capped at **1 MiB** (`MAX_HEADER_BYTES`).

The snapshot writes `embed_dim` as its **third field** — after the tiny `schema_version` / `storage_epoch` integers (see `inspect::dump::stream_snapshot`) — so the scan touches only the first few dozen bytes **regardless of total file size**.

### Why not the naive `.take(1 MiB)` + `from_reader`?
Because `from_reader` requires a clean EOF: on a legitimate snapshot **larger than 1 MiB** the capped reader truncates mid-object and serde errors `EOF while parsing` — *breaking imports of large stores*. True early-stop fixes the DoS **without** that regression. serde_json drains every object to its `}`, so the visitor abandons the object after `embed_dim`; the resulting benign trailing-comma error is ignored *only* when the value was captured.

## Doc correctness (part of #493)
The doc comment claimed "reads only the leading `embed_dim` field" while the old code parsed the whole document. It now matches behavior — resolves the `cli/correctness-import-doc-misleading-peek` item of #493. The other low/info findings in #493 remain open.

## Tests (5 inline, TDD)
- `reads_embed_dim_from_valid_snapshot` — happy path
- `stops_before_invalid_trailing_data` — returns 384 despite a poisoned non-JSON tail (proves early-stop)
- `stops_before_huge_valid_tail` — returns 384 despite a **2 MiB valid** `facts` array past the cap (proves early-stop is about *volume*, the DoS fix)
- `rejects_header_larger_than_cap` — 2 MiB leading field → error (cap bites)
- `errors_when_embed_dim_absent` — well-formed object without `embed_dim` → clean error

## Verification
- `cargo fmt --check` ✓
- `cargo clippy --workspace --all-targets` — zero warnings ✓
- `cargo test --workspace` — all pass (core 846, cli 28+15+26+4, embed 44, mcp 124); `export_import_roundtrip` still green (no import-path regression) ✓

Closes #407.
Part of #493 (import-doc-misleading-peek only).

Wave 1 of the #251 area:cli super-qa sweep.

### DIFF
--- a/memory-engine-cli/src/commands/import.rs
+++ b/memory-engine-cli/src/commands/import.rs
@@ -42,13 +42,14 @@ pub fn run(db: &Path, args: &ImportArgs) -> anyhow::Result<()> {
 /// Maximum number of bytes scanned while looking for the snapshot's `embed_dim`.
 ///
 /// In a well-formed snapshot `embed_dim` is the third field — it follows the
-/// small `schema_version` and `storage_epoch` integers (see
-/// `inspect::dump::stream_snapshot`), so it always lies within the first few
-/// dozen bytes. This deliberately generous cap bounds the scan against an
+/// small `schema_version` (`u32`) and `storage_epoch` (`u16`) integers (see
+/// `inspect::dump::stream_snapshot`), so its value always ends within ~80 bytes
+/// of the start. 64 KiB is therefore ~800× the real worst case (ample headroom
+/// for any future header growth) while still hard-bounding the scan against an
 /// adversarial snapshot that omits `embed_dim` or buries it behind a huge
-/// leading field, turning a would-be unbounded allocation (CWE-400 / CWE-770)
+/// leading field — turning a would-be unbounded allocation (CWE-400 / CWE-770)
 /// into a bounded read.
-const MAX_HEADER_BYTES: u64 = 1 << 20; // 1 MiB
+const MAX_HEADER_BYTES: u64 = 64 << 10; // 64 KiB
 
 /// Peek at a JSON snapshot file's header to extract `embed_dim` without
 /// loading the whole snapshot.
@@ -77,9 +78,10 @@ fn peek_embed_dim_from_reader(reader: impl std::io::Read) -> anyhow::Result<usiz
 
     // The sink captures `embed_dim` into `found` and stops reading the moment it
     // is seen, so the bulk of the snapshot is never parsed. Because the object is
-    // abandoned mid-stream, `deserialize_map` then reports a trailing-comma /
-    // trailing-character error from its closing-brace check — that error is
-    // *expected* and is ignored once we already hold the value. The underlying
+    // abandoned mid-stream, `deserialize_map`'s closing-brace check then fails on
+    // whatever follows the value — a trailing comma, the next key, EOF, or pure
+    // garbage; ANY post-value error from the abandoned map traversal. That error
+    // is *expected* and is ignored once we already hold the value. The underlying
     // parse error is only surfaced when `embed_dim` was never reached: a missing
     // field, malformed leading bytes, or a field buried past the byte cap.
     let mut found: Option<usize> = None;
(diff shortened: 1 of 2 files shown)


dutiona:
## Review: 3 clean-slate subagent panel (distinct adversarial lenses)

In lieu of `/super-review`, three independent clean-slate reviewers each verified against source (and ran empirical probes). **All three returned SHIP**, no BLOCKER/HIGH.

| Lens | Verdict | Key result |
|---|---|---|
| **Security / DoS adversary** | SHIP | All 6 vectors closed empirically. Notably: `IgnoredAny` skip is *iterative* in serde_json (nesting tracked on the heap, not the call stack) → deep nesting **cannot** stack-overflow; mid-multibyte truncation **errors, never panics**; oversized values rejected without alloc. |
| **serde / Rust correctness** | SHIP | No masked errors — an invalid `embed_dim` value propagates (the `?` fires before `found` is written). Non-object top level → clean error. First key wins on duplicates. Mid-stream `Deserializer` abandon is sound (no UB under `forbid(unsafe_code)`). No simpler correct alternative exists. |
| **Regression / format premise** | SHIP | Premise **CONFIRMED against `inspect::dump::stream_snapshot:43-46`**: `embed_dim` is the hand-written 3rd field (~80 bytes max). `export_import_roundtrip` exercises the peek path (no `--embed-dim`) and passes. Doc now accurate. The 64 KiB peek cap and the 4 GiB `restore_json` cap compose correctly. |

### Folded into commit d6bdf41
- **[LOW]** tightened the cap 1 MiB → **64 KiB** (defense-in-depth; ~800× headroom over the ~80-byte real max, 16× tighter adversarial ceiling)
- **[MEDIUM, comment]** broadened the inline note: the ignored post-`embed_dim` error is *any* closing error from the abandoned map, not just a trailing comma
- **5 adversarial tests** locking the invariants the reviewers verified (no-stack-overflow, UTF-8-truncation-safety, non-object, invalid-value-not-masked, duplicate-key)
- **field-order invariant comment** in `stream_snapshot` (the cross-crate producer this peek depends on)

Re-gated after the changes: `fmt` ✓, `clippy --workspace --all-targets` zero warnings ✓, `cargo test --workspace` all 35 binaries green ✓.

PR #666, opened 2026-06-20
feat(mcp): require declared model for pre-computed embeddings; reject mismatch

### LINKED ISSUE #615: feat(mcp): require declared model for pre-computed embeddings; reject mismatch
Epic: #610 · Wave 1 (guardrail)
Spec: `docs/design/embedding-identity-and-tei-qwen-migration.md` (§Design.3)

Pre-computed embedding submissions (PassthroughEmbedder) must declare a model matching the store fingerprint; reject otherwise. Closes the same-dim foreign-vector hole.

dutiona:
## What

§Design.3 of the embedding-identity spec — closes the same-dimension foreign-vector hole the `PassthroughEmbedder` sentinel left open (the `TODO(#615)` #614 named). A pre-computed `embedding` in **`memory_add_fact`** and **`memory_query`** must now declare the identity of the model that produced it, giving the vector a *real* `EmbeddingFingerprint` to record/compare instead of a sentinel.

## Behavior

- **`memory_add_fact`** with a precomputed `embedding`: the declared identity is **recorded** on a fresh store (so a precomputed-only workflow can bootstrap — re-enabling what #614's `require_present` removed) or **compared** (full-tuple `Eq`) against the stored identity on a populated one → hard-reject mismatch.
- **`memory_query`** with a precomputed `embedding`: the declared identity is **verified** against the store before retrieval → reject a foreign-space query vector.
- A precomputed `embedding` **without** a declared `model`/`provider` is rejected.

## Design (confirmed with maintainer)

- **Full-tuple strict**: `model` + `provider` required; `dim` inferred from the vector length; `matryoshka_base_dim` + `element_type` optional (default None/`float32`). Exact `Eq` compare — symmetric with how a live provider's fingerprint is treated.
- **Record on fresh store**: a declared model is a real identity, so it bootstraps a fresh store.

## Changes

- **MCP** (`tools/mod.rs`): `parse_declared_fingerprint`; `memory_add_fact` → `add_fact_precomputed(req, emb, &declared)`; `memory_query` → `verify_embedding_fingerprint(&declared)`; tool schemas gain `model`/`provider`/`matryoshka_base_dim`/`element_type`.
- **Engine**: `add_fact_precomputed` now takes the declared fingerprint and routes through `record_if_absent` (was `require_present`, #614's interim). New `MemoryEngine::verify_embedding_fingerprint` (read-only declared-identity check); `verify_embedding_identity` delegates to it.
- **Removed `PassthroughEmbedder`** (memory-engine-mcp) — its sentinel is fully replaced by the declared identity.

## Tests

- Engine: `add_fact_precomputed` record-on-fresh / match / mismatch / dim-guard (on a fresh store).
- MCP: missing-model rejected; add + query model-mismatch rejected; existing precomputed tests declare a model (and drop the now-obsolete #614 `stamp_identity` helper).

## Verification

- `cargo test --workspace` ✅ 1237 passed
- `cargo test --all-features` ✅ 1002 passed
- `cargo fmt --check` ✅ exit 0

Wave 1 of epic #610.

Fixes #615


memory-engine-mcp/src/tools/mod.rs
```diff
@@ -524,6 +532,54 @@ fn parse_embedding(args: &Map<String, Value>) -> Result<Option<Vec<f32>>, ErrorD
...
+                        None,
+                    )
+                })?,
+        ),
+    };
+    let mut fp = match matryoshka_base_dim {
+        Some(base) => EmbeddingFingerprint::with_matryoshka(model, provider, dim, base),
+        None => EmbeddingFingerprint::new(model, provider, dim),
+    };
+    if let Some(element_type) = get_str(args, "element_type") {
```
dutiona: [Codex Review] `element_type` is silently ignored when it is present with the wrong JSON type because `get_str()` returns `None` for non-strings and this branch falls back to the constructor default `"float32"`. That weakens the new full-tuple enforcement: a request like `{ "embedding": [...], "model": "m", "provider": "p", "element_type": 123 }` will be accepted against a float32 store instead of being rejected as invalid params. Please parse `element_type` explicitly and add a regression test.

dutiona (commented):
[Codex Review] Found 1 issue in the declared-fingerprint parsing path: malformed `element_type` values are silently defaulted to `float32`, which weakens the new full-tuple identity check. See inline comment for details.

memory-engine-mcp/src/tools/mod.rs
```diff
@@ -715,6 +771,17 @@ fn handle_query(
...
 
+    // §Design.3 (#615): a pre-computed query embedding must declare the model that
+    // produced it; verify that declaration against the store's identity before using the
+    // vector, so a foreign-vector-space query can't silently mis-retrieve. Covers both
+    // use sites below (with-text hybrid/vector, and embedding-only).
+    if let Some(ref emb) = pre_emb {
+        let declared = parse_declared_fingerprint(args, emb.len())?;
+        engine
+            .verify_embedding_fingerprint(&declared)
+            .map_err(to_mcp_error)?;
```
dutiona, line 795: [Antigravity Review] Placing this verification here optimally covers both the with-text hybrid fallback and the embedding-only search below, closing any potential bypass. The `to_mcp_error` correctly handles the mismatched dimension or model.

memory-engine-mcp/src/tools/mod.rs
dutiona: Fixed in b2c9410 — element_type with a present-but-non-string value is now rejected as invalid params (explicit Value::String match), instead of silently defaulting to float32. Regression test added.

memory-engine-mcp/src/tools/mod.rs
```diff
@@ -524,6 +532,54 @@ fn parse_embedding(args: &Map<String, Value>) -> Result<Option<Vec<f32>>, ErrorD
...
+                })?,
+        ),
+    };
+    let mut fp = match matryoshka_base_dim {
+        Some(base) => EmbeddingFingerprint::with_matryoshka(model, provider, dim, base),
+        None => EmbeddingFingerprint::new(model, provider, dim),
+    };
+    if let Some(element_type) = get_str(args, "element_type") {
+        fp.element_type = element_type;
+    }
```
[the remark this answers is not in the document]
dutiona: Applied — element_type now returns a validation error on a non-string value rather than silently ignoring it.

memory-engine-mcp/src/tools/mod.rs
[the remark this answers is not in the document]
dutiona: Good catch — added a to_mcp_error arm mapping MemoryError::EmbeddingModelMismatch to invalid_params (it was hitting the internal_error wildcard). Covers both the add_fact and query precomputed mismatch paths; regression test added.

dutiona: Thanks — verification placement kept; also fixed element_type validation + the mismatch→invalid_params mapping per the other reviewers.

PR #704, opened 2026-06-20
fix(cli): break on persistent read error in JSONL ingest (CWE-835)

### LINKED ISSUE #664: fix(cli): persistent read error in JSONL ingest can loop indefinitely (CWE-835)
Discovered during the #408 review (PR #660, security lens).

## Problem
`ingest_from_reader` (memory-engine-cli/src/commands/batch_ingest.rs) handles a `read_line` I/O error by warning, `total_skipped += 1`, and `continue` — **without consuming any bytes**. A reader that returns a *sticky* (non-`Interrupted`) error on every `fill_buf` would spin forever: warn + skip++ indefinitely, never reaching EOF (CWE-835, uncontrolled loop).

This is **pre-existing parity** with the original `BufReader::lines()` (its `Iterator` re-polls after an error too), and it cannot be triggered by real `File`/stdin inputs (they surface a terminal error or EOF). Hence LOW.

## Fix options
- `break` on the first non-transient read error (consistent with the skip path, which already `break`s on a `skip_until` error), or
- bail after N consecutive read errors.

The loop makes observable progress (stderr + `total_skipped` grow), so it is not a silent freeze.

Found in PR #660 (#251 Wave 2). Not a regression from #408.

dutiona:
Collateral follow-up from the #251 sweep (filed during W2 review).

## Problem (CWE-835)
`ingest_from_reader`'s read-error arm did `continue`. A reader that returns a persistent (non-`Interrupted`) error spins forever: `read_line` errors → skip → `continue` → `read_line` errors again → … The `--file` argument makes this reachable from a broken pipe / failing fd.

## Fix
`continue` → `break`. A byte-stream read error is terminal, so stop and flush whatever was already batched — consistent with the `skip_until`-error path (which already `break`s).

## Test
A regression test uses an always-erroring reader **capped at a few reads** (it panics if polled more — so a regression to looping *fails* the test instead of *hanging* the suite). RED before the fix (panicked: "looped on a persistent read error"), GREEN after.

## Verification
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean ✓
- `cargo test --workspace --all-features` — all green ✓
- `cargo deny check` — advisories ok ✓ · `cargo fmt --check` ✓

Closes #664.

### DIFF
--- a/memory-engine-cli/src/commands/batch_ingest.rs
+++ b/memory-engine-cli/src/commands/batch_ingest.rs
@@ -201,10 +201,13 @@ pub fn ingest_from_reader(
             Ok(0) => break, // EOF
             Ok(n) => n,
             Err(e) => {
+                // A read error on the byte stream is terminal — retrying would
+                // spin on a persistent error (CWE-835). Stop and flush whatever
+                // was already batched, consistent with the skip_until-error path.
                 line_no += 1;
-                eprintln!("warning: line {line_no}: read error: {e}");
+                eprintln!("warning: line {line_no}: read error, stopping: {e}");
                 total_skipped += 1;
-                continue;
+                break;
             }
         };
         line_no += 1;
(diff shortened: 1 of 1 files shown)


dutiona:
## Review: 2-lens workflow panel — both **SHIP**, no BLOCKER/HIGH

- **Behavior correctness:** `break` is right. std's `read_line`/`skip_until` already retry `ErrorKind::Interrupted` internally, so only a *terminal* (non-Interrupted) error reaches this arm — re-polling would re-spin (the CWE-835 defect). Bounded retry unwarranted (no recovery/reset at this layer). **No data loss:** break falls through to the final `flush_batch`, so records read before the error are still ingested. Symmetric with the skip_until-error path; the one `total_skipped` asymmetry is correct (avoids double-counting).
- **Test soundness:** the `assert reads<3` tripwire is the real regression-catch (RED panics on the old `continue`, GREEN on `break`; hang-proof). `Error::other` → `ErrorKind::Other` (not Interrupted), so the arm is genuinely reached. No false-pass (verified read_line calls the inner read exactly once per invocation).

Folded the two optional NITs: added `read_error_after_a_valid_line_still_flushes_it` (locks the no-data-loss-on-break claim) + tightened the bail-message assertion.

PR #711, opened 2026-06-21
refactor(core): single-source FactType string parser; reconcile CLI/MCP casing

### LINKED ISSUE #678: refactor(mcp): deduplicate fact-type parser; reconcile casing with CLI CliFactType
Found during #270 / PR #675 (3-lens review, API-completeness lens).

The MCP crate has its own fact-type converter — `fn parse_fact_type` (memory-engine-mcp/src/tools/mod.rs:618-623) — mapping the same 3 `FactType` variants, **independent of** the CLI's new `CliFactType` (memory-engine-cli/src/commands/types.rs).

**Divergence:** the MCP parser uses **PascalCase** tokens (`Episodic`/`Semantic`/`Procedural`, see the JSON-schema enums at mod.rs:61/95/185), whereas the CLI's `CliFactType` uses **snake_case** (`episodic`/…). So the two consumers of the same core enum disagree on wire casing.

The CLI consolidation (#270) legitimately didn't touch the MCP crate. Options:
- lift a shared `FactType` wrapper into the core crate or a shared util (the core intentionally exposes no ValueEnum/Deserialize on FactType), or
- at minimum reconcile the token casing between the two surfaces.

Not blocking #675.

dutiona:
## Why

The MCP server and CLI each rolled their **own** string→`FactType` parser, and they **diverged on wire casing** — MCP matched PascalCase (`"Episodic"`, per the serde derive + JSON-schema enums) while the CLI matched snake_case (`"episodic"`, per `Display`). Two parsers for one core enum, disagreeing on input. That is the divergence flagged in #678.

Root cause: core `FactType` exposed **no canonical string parse**, so every consumer invented one — and core was itself self-inconsistent (`Display` → snake_case, serde derive → PascalCase; each consumer bound to a different one of the two).

## What (single source of truth lifted into core)

- **core** (`src/types.rs`): add a case-insensitive `impl FromStr for FactType` + `ParseFactTypeError` (thiserror, already a core dep). It accepts **both** casings present in the codebase and reconciles them to one enum; `Display` stays the canonical snake_case output. **Orthogonal to the serde derive** — serde stays PascalCase on purpose to keep `.pak` cold-storage archives readable.
- **mcp** (`tools/mod.rs`): `parse_fact_type` now delegates to `FactType::from_str`, deleting the duplicate match. The JSON schemas still advertise PascalCase; parsing merely **widens** to also accept snake_case.
- **cli** (`commands/types.rs`, `batch_ingest.rs`): JSONL ingest routes through core via a `deserialize_fact_type` serde adapter; `CliFactType` shrinks to a **clap-only `ValueEnum` shim** (clap needs an owned enum for `--help`/completion; core must stay framework-free). A **round-trip test** locks the clap tokens to core's canonical casing so they cannot drift. Also fixed a CLI doc comment that falsely claimed core exposes no `Deserialize`.

Behavior is preserved for existing inputs; both surfaces now **also** accept the other's casing (back-compatible widening). No new dependencies.

## Verification

- `cargo fmt --check` ✓
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean ✓
- `cargo test --workspace` — 1525 passed ✓ · `--all-features` — 1559 passed ✓
- `cargo deny check` — advisories / bans / licenses / sources ok ✓

Closes #678

### DIFF
--- a/memory-engine-cli/src/commands/types.rs
+++ b/memory-engine-cli/src/commands/types.rs
@@ -1,19 +1,29 @@
-//! CLI-local fact-type wrapper shared across subcommands.
+//! CLI-local fact-type plumbing.
 //!
-//! `memory_engine::FactType` deliberately exposes no `clap::ValueEnum`,
-//! `serde::Deserialize`, or `FromStr` impl (the core crate has no CLI/serde
-//! concerns), so the CLI needs a thin local wrapper. This is the single source of
-//! truth for it: `add-fact` (clap arg), `query` (clap arg), and `batch-ingest`
-//! (JSONL field) all parse into [`CliFactType`] and convert via `From`.
+//! The canonical string→[`FactType`] mapping lives in **core** as
+//! [`FactType::from_str`] (case-insensitive, accepts both `snake_case` and
+//! `PascalCase`) — it is the single source of truth shared with the MCP server
+//! (#678). This module holds only the two CLI-framework shims that cannot live in
+//! the framework-free core crate:
+//!
+//! * [`CliFactType`] — a `clap::ValueEnum` so `--fact-type` gets `[possible
+//!   values: …]` in `--help` and shell completion. Its tokens are locked to
+//!   core's canonical casing by a round-trip test below, so they cannot drift.
+//! * [`deserialize_fact_type`] — a serde adapter that parses the JSONL
+//!   `fact_type` field through core's `FromStr`, so batch-ingest shares the exact
+//!   same parser as every other surface.
+
+use std::str::FromStr;
 
 use memory_engine::types::FactType;
+use serde::Deserialize;
 
-/// Fact type as accepted on the CLI (`--fact-type`) and in JSONL (`fact_type`).
+/// Fact type as accepted on the CLI (`--fact-type` on `add-fact` / `query`).
 ///
-/// The clap value tokens and the serde field values are both the lower-cased
-/// variant names (`episodic`, `semantic`, `procedural`).
-#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum, serde::Deserialize)]
-#[serde(rename_all = "snake_case")]
+/// This exists solely to give clap a `ValueEnum` (core stays framework-free). The
+/// value tokens are the lower-cased variant names (`episodic`, `semantic`,
+/// `procedural`), matching core's canonical [`FactType`] `Display`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, clap::ValueEnum)]
 pub enum CliFactType {
     Episodic,
     Semantic,
@@ -30,6 +40,20 @@ impl From<CliFactType> for FactType {
     }
 }
 
+/// Serde adapter for the JSONL `fact_type` field, routing through core's
+/// canonical [`FactType::from_str`].
+///
+/// Use via `#[serde(deserialize_with = "crate::commands::types::deserialize_fact_type")]`.
+/// Because it delegates to core, the CLI's JSONL ingest accepts the same casings
+/// as every other surface (`snake_case` and `PascalCase`) and can never diverge.
+pub fn deserialize_fact_type<'de, D>(deserializer: D) -> Result<FactType, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    FactType::from_str(&s).map_err(serde::de::Error::custom)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -44,13 +68,19 @@ mod tests {
         );
     }
 
+    /// The JSONL adapter delegates to core, so it accepts the canonical
+    /// `snake_case` and rejects unknown tokens.
     #[test]
-    fn deserializes_snake_case() {
-        assert_eq!(
-            serde_json::from_str::<CliFactType>("\"episodic\"").unwrap(),
-            CliFactType::Episodic
-        );
-        assert!(serde_json::from_str::<CliFactType>("\"unknown\"").is_err());
+    fn deserialize_fact_type_parses_snake_case() {
+        #[derive(Deserialize)]
+        struct Wrapper {
+            #[serde(deserialize_with = "deserialize_fact_type")]
+            fact_type: FactType,
+        }
+
+        let w: Wrapper = serde_json::from_str(r#"{"fact_type":"episodic"}"#).unwrap();
+        assert_eq!(w.fact_type, FactType::Episodic);
+        assert!(serde_json::from_str::<Wrapper>(r#"{"fact_type":"unknown"}"#).is_err());
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


src/types.rs
```diff
@@ -66,6 +68,38 @@ impl fmt::Display for FactType {
...
+    /// ingest, the MCP server's tool parameters). It is intentionally lenient on
+    /// casing so it accepts both wire conventions present in the codebase:
+    /// `Display` emits `snake_case` (`"episodic"`), while serde-derive and the MCP
+    /// JSON-schema enums use `PascalCase` (`"Episodic"`). Parsing reconciles both
+    /// to one canonical enum; [`FactType::to_string`] remains the canonical output.
+    ///
+    /// Note: this is orthogonal to the serde `Deserialize` derive, which stays
+    /// `PascalCase` to preserve `.pak` cold-storage archive back-compat.
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_ascii_lowercase().as_str() {
```
dutiona: [Antigravity Review] `s.to_ascii_lowercase()` allocates a new `String` on every invocation. For a `FromStr` parser, it is more efficient to avoid allocation by using zero-allocation comparisons like `if s.eq_ignore_ascii_case("episodic") { ... }`.

dutiona (commented):
[Antigravity Review] Review of PR #711 complete.

**Findings:**
1. **Performance (Medium):** The `FromStr` impl in `types.rs` allocates a new `String` on every valid parse (`s.to_ascii_lowercase().as_str()`). I've left an inline comment suggesting to use zero-allocation checks like `eq_ignore_ascii_case`.
2. **Missed Parse Site (Medium):** There is still a hand-rolled parser for `FactType` at `src/store/facts.rs:51` (`fn str_to_fact_type`). This needs to be updated to delegate to the new `FromStr` canonical parser.
3. **Behavior/Compatibility (Clean):** Widening the casing acceptance surface is a safe change and won't cause regressions. The `.pak` archive back-compatibility via `serde` `Deserialize` remains strictly `PascalCase` because the `FromStr` trait does not interfere with the derived `Deserialize` behavior.
4. **Testing (Clean):** The tests added are comprehensive. The `clap_tokens_round_trip_through_core_from_str` test genuinely guards against drift between the CLI string tokens and core string mapping. It is not tautological.

Please address the allocations and the missed parse site.

memory-engine-mcp/src/tools/mod.rs
```diff
@@ -616,13 +616,14 @@ fn parse_event_type(s: &str) -> Result<EventType, ValidationError> {
     }
 }
 
+/// Parse an MCP `fact_type` tool parameter into the core [`FactType`].
+///
+/// Delegates to core's canonical [`FactType::from_str`] (the single source of
+/// truth shared with the CLI), so casing is reconciled across surfaces: the JSON
+/// schemas advertise `PascalCase` (`"Episodic"`), but `snake_case` is also accepted.
```
dutiona, line 623: [Codex Review] [MEDIUM] The handler now accepts `snake_case`, but the MCP tool schemas above still publish `enum: ["Episodic", "Semantic", "Procedural"]` for `memory_add_fact` / `memory_query`. Schema-validating clients can reject `episodic` before `parse_fact_type()` ever runs, so the advertised MCP contract is still PascalCase-only. If the goal is real cross-surface reconciliation, widen or relax the schema alongside this parser.

memory-engine-cli/src/commands/types.rs
```diff
@@ -1,19 +1,29 @@
...
+//! `PascalCase`) — it is the single source of truth shared with the MCP server
+//! (#678). This module holds only the two CLI-framework shims that cannot live in
+//! the framework-free core crate:
+//!
+//! * [`CliFactType`] — a `clap::ValueEnum` so `--fact-type` gets `[possible
+//!   values: …]` in `--help` and shell completion. Its tokens are locked to
+//!   core's canonical casing by a round-trip test below, so they cannot drift.
+//! * [`deserialize_fact_type`] — a serde adapter that parses the JSONL
+//!   `fact_type` field through core's `FromStr`, so batch-ingest shares the exact
+//!   same parser as every other surface.
```
dutiona, line 14: [Codex Review] [MEDIUM] This overstates what the PR actually enforces. `batch-ingest` now routes through `FactType::from_str`, but the `add-fact` clap path still parses through `CliFactType` and remains case-sensitive. I verified locally that `memory-engine-cli add-fact --fact-type Semantic ...` is still rejected by clap, while `query --fact-type Semantic` gets past argument parsing because that arg has `ignore_case = true`. So the CLI surface still diverges from the new canonical parser/test lock.

dutiona (commented):
[Codex Review] I found two medium-severity contract gaps in the casing reconciliation. First, the CLI still does not share one parser end-to-end: `batch-ingest` uses `FactType::from_str`, but `add-fact` still parses through clap `ValueEnum` and remains case-sensitive, so `--fact-type Semantic` is rejected there even though `query` accepts it with `ignore_case = true`. Second, the MCP handler now accepts `snake_case`, but the published tool schemas still enumerate only PascalCase tokens, so schema-validating clients can reject lowercase inputs before the new parser runs. I did not find evidence that `.pak` serde back-compat was broken; the remaining issue is that the widened parser is only partially exposed on the actual CLI/MCP surfaces.

src/types.rs
dutiona: Fixed in fa1f9f5: `FromStr` now matches via `eq_ignore_ascii_case` — zero allocation on the parse path (no temporary lowercased `String`).

memory-engine-cli/src/commands/types.rs
dutiona: Fixed in fa1f9f5: `add-fact --fact-type` now sets `ignore_case = true` (matching `query`), so both CLI clap surfaces are case-insensitive and consistent with core. The doc is now accurate — every CLI string→FactType path (clap args + JSONL) routes through / aligns with the canonical core parser.

memory-engine-mcp/src/tools/mod.rs
dutiona: Acknowledged, and intentional. The PascalCase JSON-schema enum is the advertised MCP contract and stays as-is; the parser widening to also accept snake_case is deliberate *server-side leniency* (so a non-schema-validating client or an internal caller passing snake_case still works), not a contract change. The doc comment on `parse_fact_type` states exactly this. Changing the advertised schema enum would be a separate, intentional contract decision — out of scope for this dedup.

dutiona (commented):
[Codex Review] No new findings in the latest commit (`fa1f9f52f54be47db38248eca479064a9267dabd`). The prior CLI inconsistency is resolved: `add-fact` now matches `query` on case-insensitive clap parsing. I also re-checked the store delegation and did not find a regression risk: persisted `facts.fact_type` values are still written as lowercase snake_case, so delegating `str_to_fact_type` to the canonical core parser preserves existing on-disk data while only widening reads for non-canonical/manual values. The `eq_ignore_ascii_case` rewrite preserves the accepted token set for these ASCII literals, and the new whitespace/empty-string rejection test covers the intended strictness.

PR #726, opened 2026-06-21
chore(build): deny clippy::unwrap_used in library code (#725)

### LINKED ISSUE #725: chore(build): enforce no-unwrap-in-library-code via clippy::unwrap_used restriction lint
## Summary

The core crate currently has **zero `.unwrap()` in production code** — all 2538 occurrences are in `#[cfg(test)]` modules or `///` doc-examples. Nothing *enforces* this, so a future production `.unwrap()` can land silently. This issue installs the `clippy::unwrap_used` **restriction** lint to lock in the already-achieved invariant.

## Why this matters (for an *embedded* library)

`memory-engine` runs in-process inside the consumer's address space (the autonomous agent). A panic from an `.unwrap()` in a path like `add_fact` / hybrid search / consolidation aborts **the consumer's process**, not a binary we control. This is the same defense-in-depth rationale behind the existing `unsafe_code = "forbid"`: make a whole class of footgun *impossible by tooling*, not merely *avoided by vigilance*. With the `#560` typed-`MemoryError` work landed, every production fallibility already has a structured home — an `.unwrap()` that throws that away is strictly a regression.

## Current state (audit, 2026-06-21)

| Bucket | `.unwrap()` count |
|---|---|
| Tests (`#[cfg(test)]`, `*tests.rs`, `tests/`) + doc-examples | 2538 |
| **Production (non-test, non-doc)** | **0** |

`clippy::unwrap_used`/`expect_used` live in clippy's `restriction` group, which is *not* pulled in by the crate's existing `pedantic` + `nursery` config — hence currently unlinted.

## Proposed change

1. Add to the workspace `[lints.clippy]` in `Cargo.toml`:
   - `unwrap_used = "deny"` — production `.unwrap()` becomes a hard error. (Start at `"warn"` for one PR cycle if a soft landing is preferred, then flip to `deny`.)
   - Keep `expect_used = "allow"` — `expect("documented invariant")` is the sanctioned escape hatch for genuinely-infallible-but-uncheckable cases (e.g. a regex match immediately after a successful `is_match`), since it forces an inline justification string.
2. Exempt test code so the 2538 test unwraps don't light up:
   - Crate root: `#![cfg_attr(test, allow(clippy::unwrap_used))]` — covers inline `mod tests`, `#[cfg(test)] mod foo;` modules (e.g. `engine/equivalence.rs`, `storage/sqlite/realization.rs`), and `*tests.rs`.
   - Integration tests under `tests/` are separate crates: add a top-of-file `#![allow(clippy::unwrap_used)]` to each, or accept that `cfg(test)` is set there too and rely on the same `cfg_attr` pattern per file.
3. Confirm doc-examples are unaffected (doctests are idiomatic with `.unwrap()` and are not linted by the `[lints]` table).

## Acceptance criteria

- [ ] `cargo clippy --workspace --all-targets --all-features` is clean with `unwrap_used = "deny"`.
- [ ] A deliberately-added production `.unwrap()` fails the clippy CI job (manually verified once).
- [ ] Test modules and doctests still compile and lint clean (no churn to the 2538 existing test unwraps).
- [ ] One-line note in `CLAUDE.md` Code Style / verification gate documenting the policy: production = `?`/`expect("why")`; tests = `unwrap` free.

## Out of scope

- The companion crates (`-cli`, `-mcp`, `-embed`) — file follow-ups if we want the same gate there; the CLI especially has legitimate top-level `unwrap`/`expect` in `main`-adjacent glue that may warrant `expect` rather than `deny`.

## Notes

Audit performed via a `#[cfg(test)]`-boundary + inner-attribute + `mod`-site gating scan; the two apparent "production" modules (`engine/equivalence.rs`, `storage/sqlite/realization.rs`) are `#[cfg(test)] mod …` declarations (mod.rs:57 / mod.rs:52), confirming true production count = 0.

dutiona:
Closes #725.

## What

Enforce the already-true invariant that the `memory-engine` core library contains **no `.unwrap()` in production code**, via `clippy::unwrap_used = "deny"`.

## Why

`memory-engine` is an embedded library — it runs in-process inside the consumer's address space. An `.unwrap()` panic in a production path (e.g. `add_fact`, hybrid search, consolidation) aborts the **consumer's** process, not a binary we control. With the typed-`MemoryError` work landed, production fallibility already has a structured home, so a panicking unwrap is strictly a regression. This is the same defense-in-depth philosophy as the existing `unsafe_code = "forbid"`: make a class of footgun *impossible by tooling*, not merely *avoided by vigilance*.

An audit (documented in #725) found the production unwrap count is already **0** — all 2538 `.unwrap()` occurrences are in tests or doc-examples. This PR locks that in.

## How

| File(s) | Change |
| --- | --- |
| `Cargo.toml` | `clippy::unwrap_used = "deny"` (package-scoped to core). `expect("…")` stays allowed as the sanctioned escape hatch. |
| `src/lib.rs` | `#![cfg_attr(test, allow(clippy::unwrap_used))]` — exempts the crate's own `#[cfg(test)]` unit tests, where a panic is the intended failure signal. |
| 16 integration-test / bench crate roots | justified, file-local `#![allow(clippy::unwrap_used)]` (only where `.unwrap()` is actually used). |
| `CLAUDE.md` | policy recorded as Key Design Decision #7. |

**Fail-safe by design:** a new test file that uses `.unwrap()` without the opt-out **red-builds**. The default is "deny"; opting out must be deliberate and justified — exactly the property we want.

## Verification

- `cargo clippy --workspace --all-targets --all-features` — clean (0 warnings/errors). This empirically confirms zero production unwraps *across all feature-gated paths* (`ann`/`archive`/`async`/`compress-*`), not just the default feature set.
- `cargo test --all-features` — 1137 unit + all integration binaries pass, 0 failures.
- `cargo test -p memory-engine-cli -p memory-engine-mcp -p memory-engine-embed` — pass.
- `cargo fmt --check` — clean.
- No dependency changes → `cargo deny` unaffected.

## Out of scope

The companion crates (`-cli`, `-mcp`, `-embed`) are not gated here — follow-ups can extend the policy, noting the CLI's legitimate top-level `unwrap`/`expect` in `main`-adjacent glue may warrant `expect` rather than `deny`.


dutiona:
## Adversarial review outcome — SHIP

One adversarial reviewer (rust-specialist) checked scope leakage, test-exemption completeness, lib.rs `cfg_attr` correctness, gate effectiveness, and examples. Verdict **SHIP**. Confirmed clean: deny is package-scoped (no `lints.workspace` in consumer crates), all test/bench roots exempted, `examples/` are unwrap-free, lib `#![cfg_attr(test, allow(...))]` placement valid.

### Known limitation (documented, not silently accepted)

**Clippy does not lint doctests** — the `[lints]` table doesn't reach rustdoc, even under `--all-targets`. So a production *doctest* `.unwrap()` escapes this deny. There is no clean way to enforce `unwrap_used` inside doctests via the lint system.

Mitigation applied: the most visible offender — the public `DefaultDreamCycle` usage example — was rewritten to model `?` instead of `.unwrap()` (commit 9e9ac03), so the canonical copy-paste example aligns with the policy. Remaining doctest unwraps are conventional setup boilerplate (`tempfile::tempdir().unwrap()`, `Runtime::new().unwrap()`) where `?` adds noise; left as-is. This blind spot is inherent to Clippy, tracked here for awareness rather than folded into scope.

Issue #499, opened 2026-06-01
[super-qa] retrieval: low/info findings (13)


dutiona:
> ⚠️ Agent analysis; not source-verified. Grouped low/info for **area:retrieval**.

Grouped low/info findings for area **retrieval** (13):

- **low** `src/graph/memory_graph.rs:78-87` — remove_node has no # Panics section; petgraph remove_node panics on out-of-bounds index (`graph/documentation-remove-node-missing-panics`)
- **low** `src/search/query.rs:17-31` — Doc example variable named `results` but holds a `QueryResponse`, not a `Vec` (`search/documentation-example-return-type-misleading`)
- **low** `src/search/ann.rs:170-217` — `from_snapshot` and `to_snapshot` missing `# Errors` sections (`search/documentation-from-snapshot-missing-errors`)
- **low** `src/graph/memory_graph.rs:67-76` — remove_edge_by_id performs an O(E) full edge scan (`graph/performance-remove-edge-by-id-linear-scan`)
- **low** `src/graph/memory_graph.rs:237-247` — from_snapshot trusts deserialized edge fields (source/target/edge_id/weight) without revalidating against the DB (`graph/data-integrity-trusts-snapshot-edges`)
- **low** `src/graph/memory_graph.rs:190-208, 338-393` — load_from_db test covers only 1 active + 1 expired edge; empty table, weight preservation, and parallel edges are untested (`graph/testing-load-from-db-edge-cases`)
- **low** `src/graph/memory_graph.rs:142-136` — Strong property-based test candidates exist for graph invariants but proptest is unused in the graph module (`graph/testing-proptest-graph-invariants`)
- **low** `src/search/ann.rs:145-218` — `HnswStrategy::to_snapshot` / `from_snapshot` round-trip semantics not tested (`search/testing-hnsw-snapshot-roundtrip`)
- **low** `src/search/query.rs:17` — Only doc-example is `rust,ignore` — no compile-checked runnable examples across the public search API (`search/testing-doctest-query-ignore`)
- **low** `src/search/fts.rs:30, src/search/hybrid.rs:136` — No fuzz targets for FTS5 query text — untrusted string input path (`search/testing-no-fuzz-targets`)
- **info** `src/graph/mod.rs:1` — Module doc describes MemoryGraph as 'knowledge graph' but struct doc calls it entity/relation graph; isolated-node behavior undocumented (`graph/documentation-module-doc-incomplete`)
[...]

Issue #257, opened 2026-06-01
[super-qa] Area: retrieval — code-quality findings


dutiona:
All `/super-qa` 2026-06-01 findings in **area:retrieval**. Non-autofixable blocker/critical/high/medium are individual sub-issues; low/info are grouped in one sub-issue.

dutiona:
## Epic complete — all 22 sub-issues resolved ✅

The `area:retrieval` super-qa sweep is done. The 22 findings were triaged/reconciled against current `main` (many had drifted since the 2026-06-01 audit), then addressed across **3 collision-safe waves of PRs**, each following the full `/address-issue` workflow (TDD implement → agy + adversarial-lens review → fix → finish-pr → squash-merge).

**Reconciliation outcome:** 2 already-fixed (closed with evidence), 3 duplicates collapsed (#284/#313→#261, #362→#288, #285⊃#295), the rest implemented.

**Wave 1** (6 PRs): `#833` remove_node + edge swap-remove fix (#261/#284/#313) · `#837` HnswStrategy::notify_insert → `IndexInconsistent` (#295/#285) · `#835` expired-probe e2e (#324) · `#831` cosine truncation tests (#479) · `#838` FTS fuzz target (→#499) · `#830` hybrid_search extraction (#363)

**Wave 2** (3 PRs): `#867` batch HNSW candidate fetch — killed the 2N N+1 (#288/#362) · `#865` search public-surface curation + SearchQuery builder (#364/#365) · `#866` snapshot edge referential-validation (#412)

**Wave 3** (3 PRs): `#875` shared `build_hnsw_inner` kernel + white-box snapshot roundtrip (#404) · `#873` hybrid unit tests (#325/#480/#478) · `#874` graph module coverage (#456/#457)

**`#499`** (13 low/info) distributed across the waves — see its closing comment for the per-finding map.

**Already-fixed closes:** #287 (N+1 killed by `get_many`), #385 (SearchConfig doc).

**Collateral filed:** #852 (single-source HashSet opt), #871 (cfg-gate the test-only sync oracle). #756/#751 (clippy/fuzz CI gaps) confirmed pre-tracked.

The adversarial review earned its keep — it caught a graph-corruption BLOCKER (#833's edge sibling), a write-path contract redesign (#837), an over-strict security regression (#866), and a verification-theater test (#875), none of which `cargo test` flagged.

PR #833, opened 2026-06-26
fix(retrieval): re-index displaced node in MemoryGraph::remove_node after petgraph swap-remove

### LINKED ISSUE #261: [super-qa] high: remove_node leaves node_map with a stale NodeIndex for the displaced node after petgraph's swap-remove
> ✅ **Verified against primary source.** VERIFIED real by source; DORMANT — only caller engine/archive.rs:86 is #[cfg(feature=archive)] (blocked by #1), so cannot fire until archive compiles

**remove_node leaves node_map with a stale NodeIndex for the displaced node after petgraph's swap-remove**

- **ID:** `graph/correctness-remove-node-stale-nodemap`
- **Severity:** high
- **Area:** retrieval · **Type:** bug
- **Location:** `src/graph/memory_graph.rs:82-87`
- **Auto-fixable:** no

## Description
petgraph::DiGraph::remove_node uses swap-remove semantics (documented in petgraph 0.7.1 src/graph_impl/mod.rs:639 and 668): removing the node at NodeIndex(k) moves the last node (at NodeIndex(N-1)) into position k. MemoryGraph::remove_node correctly removes fact_id from node_map, but never updates the map entry for the displaced fact whose NodeIndex just changed from N-1 to k.

Every subsequent operation that calls node_map.get(&displaced_fact_id) → stale_idx → self.graph[stale_idx] is broken:
- has_node(displaced_fact) returns true (node_map still has the key) — false sense of safety
- degree(displaced_fact) and neighbors(displaced_fact) call neighbors_directed(stale_idx, …) and return data for whatever fact now occupies stale_idx, or return 0/empty if stale_idx is out-of-range
- connected_component(displaced_fact) starts BFS from the wrong node and may silently corrupt importance scores and query results
- ensure_node(displaced_fact) returns the stale NodeIndex; any add_edge call wires the new edge to the wrong graph node, producing invisible write-side data corruption
- remove_edges_by_fact(displaced_fact) removes edges from the wrong node

The call site in engine/archive.rs:84-87 loops over fact_ids (≥100 facts, the ArchivePolicy default) calling remove_edges_by_fact then remove_node for each. Every iteration after the first operates on a corrupted node_map. The archive feature makes this reachable in production.

## Suggested approach
Option A (preferred, eliminates the foot-gun for all future callers): switch DiGraph to petgraph::stable_graph::StableDiGraph. StableDiGraph does not relocate nodes on removal; NodeIndex values remain valid until that node is explicitly removed, so node_map never becomes stale.

Option B (targeted fix without changing the graph type): after calling self.graph.remove_node(idx), check if a displaced node now occupies idx and update node_map accordingly:

```rust
pub fn remove_node(&mut self, fact_id: i64) {
    let Some(idx) = self.node_map.remove(&fact_id) else {
        return;
    };
    self.graph.remove_node(idx);
    // petgraph swap-remove: the node that was last (now at `idx`) needs
    // its node_map entry updated. self.graph[idx] holds its fact_id.
    if let Some(&displaced_fact_id) = self.graph.node_weight(idx) {
        self.node_map.insert(displaced_fact_id, idx);
    }
}
```

Note: node_weight returns None when idx is out-of-range (the removed node was the last one, no displacement), so the guard is safe.

## References
petgraph 0.7.1 src/graph_impl/mod.rs:639-690 (remove_node swap-remove), src/engine/archive.rs:84-87 (primary call site — loop over ≥100 fact_ids)

### LINKED ISSUE #284: [super-qa] high: remove_node doc omits NodeIndex invalidation side-effect; silent correctness hazard in archive loop

**remove_node doc omits NodeIndex invalidation side-effect; silent correctness hazard in archive loop**

- **ID:** `graph/documentation-remove-node-stale-index`
- **Severity:** high
- **Area:** retrieval · **Type:** docs
- **Location:** `src/graph/memory_graph.rs:78-87`
- **Auto-fixable:** no

## Description
MemoryGraph::remove_node (line 82) calls petgraph::DiGraph::remove_node, which uses swap_remove internally (petgraph 0.7.1 src/graph_impl/mod.rs:668): the last node in the internal Vec is moved into the vacated slot. The MemoryGraph::node_map (HashMap<i64, NodeIndex>) removes the entry for the deleted fact_id (line 83) but does NOT update the node_map entry for the node that was relocated by swap_remove. After the call, that node's map entry points to a stale (now out-of-bounds or aliased) NodeIndex. All subsequent operations that use node_map — neighbors(), degree(), connected_component(), remove_edges_by_fact() — will silently produce wrong results or access wrong nodes for the relocated node.

The only non-test caller is src/engine/archive.rs:83-88, which iterates a fact_ids Vec and calls remove_edges_by_fact + remove_node in a loop. Any archive batch with more than one fact reliably corrupts node_map for surviving nodes whose internal petgraph index was the 'last' before each swap_remove.

The doc comment (lines 78-81) is silent on this invariant break. It says 'Remove a node and all its edges from the graph. No-op if the fact id is not in the graph.' — both claims are locally true but together they imply the operation is safe to call iteratively, which it is not with the current DiGraph + node_map design.

No test covers sequential remove_node calls on a multi-node graph followed by queries on surviving nodes, so the bug is latent but not caught by the test suite.

## Suggested approach
Fix the implementation so node_map is kept consistent after each swap_remove. After `self.graph.remove_node(idx)`, if petgraph placed a node at `idx` (i.e., `idx` was not the last slot), read back its weight (the fact_id) and update node_map:

```rust
pub fn remove_node(&mut self, fact_id: i64) {
    let Some(idx) = self.node_map.remove(&fact_id) else {
        return;
    };
    self.graph.remove_node(idx);
    // petgraph::DiGraph uses swap_remove: the former last node is now at `idx`.
    // Update node_map so that node's entry points to its new index.
    if let Some(&relocated_fact_id) = self.graph.node_weight(idx) {
        self.node_map.insert(relocated_fact_id, idx);
    }
}
```

Alternatively, switch `DiGraph` to `petgraph::stable_graph::StableDiGraph`, which does not invalidate NodeIndex values on removal (uses tombstones instead of swap_remove). That eliminates the invariant entirely at the cost of slightly higher memory use.

Also add a regression test:
```rust
#[test]
fn remove_node_keeps_surviving_nodes_accessible() {
    let mut g = MemoryGraph::new();
    g.add_edge(1, 2, make_edge_data(1, "a"));
    g.add_edge(2, 3, make_edge_data(2, "b"));
    // remove node 1 — petgraph swap_remove may relocate node 3 to slot 0
    g.remove_node(1);
    // nodes 2 and 3 must still be reachable via node_map
    assert!(g.has_node(2));
    assert!(g.has_node(3));
    assert_eq!(g.neighbors(2), vec![3]);
    assert_eq!(g.degree(3), 1);
}
```

## References
petgraph 0.7.1 src/graph_impl/mod.rs:648-691 (swap_remove in DiGraph::remove_node), src/engine/archive.rs:83-88 (only non-test caller; iterates fact_ids in a loop)

### LINKED ISSUE #313: [super-qa] high: remove_node is untested and its petgraph swap_remove semantics silently corrupt node_map

**remove_node is untested and its petgraph swap_remove semantics silently corrupt node_map**

- **ID:** `graph/testing-remove-node-stale-index`
- **Severity:** high
- **Area:** retrieval · **Type:** test
- **Location:** `src/graph/memory_graph.rs:82-87`
- **Auto-fixable:** no

## Description
MemoryGraph::remove_node has zero unit tests. The function calls petgraph DiGraph::remove_node (petgraph-0.7.1/src/graph_impl/mod.rs:668), which uses swap_remove: the last node in the internal Vec is moved into the removed node's slot. MemoryGraph does NOT update node_map for the displaced node, so after any remove_node call every subsequent lookup on the swapped-in node resolves to a stale, now-invalid NodeIndex — yielding wrong neighbors/degree results or out-of-bounds panics. The production path in src/engine/archive.rs:84-87 calls remove_node in a loop over multiple fact_ids, making the corruption cumulative. A test with three nodes where the middle one is removed, then operations on the remaining two, would immediately fail under the current implementation.

## Suggested approach
Add a test that exposes the bug and then either (a) fix MemoryGraph::remove_node to update node_map for the displaced node, or (b) switch the backing graph from petgraph::graph::DiGraph to petgraph::stable_graph::StableDiGraph, which preserves NodeIndex through removals and makes the stale-map problem impossible.

Minimal reproducer test:
```rust
#[test]
fn remove_node_updates_node_map_for_swapped_node() {
    let mut g = MemoryGraph::new();
    g.add_edge(1, 2, make_edge_data(10, "a")); // nodes: 0→1, 1→2
    g.add_edge(2, 3, make_edge_data(20, "b")); // node 2→NodeIndex(2)
    g.remove_node(1); // petgraph swaps node at idx=2 (fact_id=3) into idx=0
    // WITHOUT the fix, node_map[3] still holds NodeIndex(2) which is now OOB
    assert!(g.has_node(2)); // must still be reachable
    assert!(g.has_node(3)); // must still be reachable
    assert_eq!(g.neighbors(2), vec![3]); // must not panic or return wrong data
}
```

Fix for remove_node:
```rust
pub fn remove_node(&mut self, fact_id: i64) {
    let Some(idx) = self.node_map.remove(&fact_id) else { return; };
    self.graph.remove_node(idx);
    // petgraph swaps the last node into 'idx'; update its node_map entry.
    if let Some(&swapped_fact_id) = self.graph.node_weight(idx) {
        self.node_map.insert(swapped_fact_id, idx);
    }
}
```

Alternative — switch backing type:
```rust
use petgraph::stable_graph::{StableDiGraph, NodeIndex};
// StableDiGraph never invalidates NodeIndex on removal; node_map stays valid.
```

## References
petgraph-0.7.1/src/graph_impl/mod.rs:648-691 (swap_remove implementation), src/engine/archive.rs:84-87 (production callsite — multi-node removal loop)

dutiona:
## What

`MemoryGraph::remove_node` now keeps `node_map` consistent after petgraph's swap-remove, and the previously-untested path gains a regression test + a documented invalidation contract.

## Why

petgraph 0.7 `DiGraph::remove_node` uses **swap-remove**: deleting the node at `NodeIndex(k)` relocates the former last node (at `NodeIndex(N-1)`) into slot `k`, invalidating that displaced node's cached `NodeIndex`. The old `remove_node` dropped the deleted fact from `node_map` but never updated the entry for the **displaced** fact — leaving a stale (out-of-range or aliased) `NodeIndex`.

Every later lookup on the displaced fact (`has_node` / `neighbors` / `degree` / `connected_component` / `remove_edges_by_fact` / `ensure_node`) then resolved against the wrong slot, silently returning empty/wrong data or wiring new edges to the wrong node. The only production caller is the archival loop (`engine/archive.rs`, `#[cfg(feature = "archive")]`), which removes many nodes in sequence — so the corruption is **cumulative**.

## Fix

After `graph.remove_node(idx)`, read back `node_weight(idx)` (the displaced fact id, or `None` when the removed node *was* the last slot — a no-op) and rewrite its `node_map` entry to `idx`. This is the minimal, local Option B from the findings; deliberately **not** switching to `StableDiGraph` (larger blast radius, conflicts with other in-flight PRs on this file).

## Tests

`remove_node_keeps_surviving_nodes_accessible` builds a 3-node chain, removes a **non-last** node to force a displacement, and asserts every surviving node still resolves to its correct neighbors/degree/component. Confirmed RED before the fix (the displaced node's `degree` resolved to 0), GREEN after. Full `cargo test -p memory-engine` (1217) green; `cargo fmt --check` clean; `cargo clippy -p memory-engine --all-targets` clean on the changed file (incl. `--features archive`).

Part of #257, #499.

Closes #261
Closes #284
Closes #313

## Follow-up: edge sibling of the swap-remove bug (adversarial-review BLOCKER)

Adversarial review found the **same petgraph swap-remove invalidation** in `remove_edges_by_fact` — now for **edges** (`Graph::remove_edge` swap-removes, relocating the current last edge into the freed slot). The old code collected a node's outgoing+incoming `EdgeIndex`es into `to_remove` and removed them in collection order, which:

1. **Invalidates a still-pending cached `EdgeIndex`** when a prior `remove_edge` swap-relocates the last edge into the freed slot — silently **leaking** the relocated edge (or deleting the wrong one).
2. **Double-deletes** on a **self-loop** (`source == target`): it is yielded by BOTH the Outgoing and Incoming iterators, so its `EdgeIndex` is pushed twice and the second `remove_edge` deletes an innocent relocated edge.

**Fix:** sort `to_remove` **descending** then `dedup()` before removing — highest-index-first guarantees each removed edge is always the current last edge (swap-remove relocates nothing still pending); `dedup` (valid only after the sort) collapses the self-loop duplicate.

Two TDD tests added, both verified RED against real petgraph 0.7 under the bug, GREEN after:
- `remove_edges_by_fact_with_self_loop`
- `remove_edges_by_fact_multi_edge_preserves_others`

This folds the adversarial-review BLOCKER into this PR as the **edge sibling** of the `remove_node` swap-remove fix above.

### DIFF
--- a/memory-engine/lib/memory-engine/src/graph/memory_graph.rs
+++ b/memory-engine/lib/memory-engine/src/graph/memory_graph.rs
@@ -122,6 +122,19 @@ impl MemoryGraph {
                 .edges_directed(idx, Direction::Incoming)
                 .map(|e| e.id()),
         );
+        // petgraph's `remove_edge` swap-removes: it relocates the current last
+        // edge into the freed slot, invalidating any still-pending cached
+        // `EdgeIndex` that pointed at that last slot. Removing highest-index
+        // first guarantees each edge we remove is always the current last edge,
+        // so swap-remove never relocates anything still in `to_remove`.
+        //
+        // The descending sort also groups duplicate indices adjacently so
+        // `dedup` can collapse them: a self-loop (source == target) is yielded
+        // by BOTH the Outgoing and Incoming iterators, pushing the same
+        // `EdgeIndex` twice. Without dedup the second `remove_edge` would delete
+        // an innocent relocated edge.
+        to_remove.sort_unstable_by(|a, b| b.cmp(a));
+        to_remove.dedup();
         for ei in to_remove {
             self.graph.remove_edge(ei);
         }
@@ -408,6 +421,79 @@ mod tests {
         let mut comp = g.connected_component(3);
         comp.sort_unstable();
         assert_eq!(comp, vec![2, 3]);
+    }
+
+    #[test]
+    fn remove_edges_by_fact_with_self_loop() {
+        let mut g = MemoryGraph::new();
+        // Self-loop on fact 1, plus a genuine edge 1→2 and an unrelated edge 3→4.
+        // The self-loop is yielded by BOTH the Outgoing and Incoming iterators,
+        // so its EdgeIndex is collected twice — without dedup the second remove
+        // would swap-delete an innocent edge.
+        g.add_edge(1, 1, make_edge_data(10, "self"));
+        g.add_edge(1, 2, make_edge_data(20, "a"));
+        g.add_edge(3, 4, make_edge_data(30, "unrelated"));
+        assert_eq!(g.edge_count(), 3);
+
+        // Must not panic (the double-collected self-loop would otherwise try to
+        // remove an already-relocated index).
+        g.remove_edges_by_fact(1);
+
+        // The self-loop (1→1) and the fact's edge (1→2) are gone.
+        assert_eq!(g.edge_count(), 1);
+        assert_eq!(g.degree(1), 0);
+        assert!(g.neighbors(1).is_empty());
+
+        // The unrelated edge 3→4 must survive intact, right endpoints and all.
+        assert_eq!(g.neighbors(3), vec![4]);
+        assert_eq!(g.degree(3), 1);
+        assert_eq!(g.degree(4), 1);
+    }
+
+    #[test]
+    fn remove_edges_by_fact_multi_edge_preserves_others() {
+        let mut g = MemoryGraph::new();
+        // This exact layout reproduces the swap-remove invalidation against real
+        // petgraph 0.7: removing fact 2's edges frees a non-last slot, so
+        // petgraph swap-relocates the current last edge into it, invalidating a
+        // still-pending cached EdgeIndex in `to_remove`. The buggy
+        // collection-order removal then either skips that stale (now
+        // out-of-bounds) index — LEAKING a fact-2 edge that should be gone — or
+        // deletes the wrong relocated edge. Verified empirically: under the bug,
+        // edge 13 (3→2) survives; the fix removes all of fact 2's edges and
+        // leaves only the genuine survivor 11 (1→3).
+        //
+        // Edges touching fact 2: 1→2 (in), 2→1 (out), 3→2 (in).
+        // The only edge NOT touching fact 2: 1→3 (must survive verbatim).
+        g.add_edge(1, 2, make_edge_data(10, "drop_in")); // EdgeIndex 0
+        g.add_edge(1, 3, make_edge_data(11, "survivor")); // EdgeIndex 1
+        g.add_edge(2, 1, make_edge_data(12, "drop_out")); // EdgeIndex 2
+        g.add_edge(3, 2, make_edge_data(13, "drop_in2")); // EdgeIndex 3
+        assert_eq!(g.edge_count(), 4);
+
+        g.remove_edges_by_fact(2);
+
+        // Exactly one edge survives — the only one not touching fact 2.
+        assert_eq!(g.edge_count(), 1);
+
+        // Fact 2 is fully disconnected: none of its three edges leaked.
+        assert_eq!(g.degree(2), 0);
+        assert!(g.neighbors(2).is_empty());
+
+        // The survivor 1→3 resolves to the correct endpoints/degree.
+        assert_eq!(g.neighbors(1), vec![3]);
+        assert_eq!(g.degree(1), 1);
+        assert_eq!(g.degree(3), 1);
+
+        // Cross-check via the snapshot: exactly the survivor, all data intact.
+        let snap = g.to_snapshot();
+        assert_eq!(snap.edges.len(), 1);
+        let e = &snap.edges[0];
+        assert_eq!(e.edge_id, 11);
+        assert_eq!(e.source, 1);
+        assert_eq!(e.target, 3);
+        assert_eq!(e.relation_type, "survivor");
+        assert!((e.weight - 1.0).abs() < f64::EPSILON);
     }
 
     #[test]
--- a/memory-engine/lib/memory-engine/src/graph/memory_graph.rs
+++ b/memory-engine/lib/memory-engine/src/graph/memory_graph.rs
@@ -424,6 +424,81 @@ mod tests {
     }
 
     #[test]
+    fn remove_node_in_loop_keeps_map_consistent() {
+        // The single-removal test above exercises `remove_node`'s re-index guard
+        // exactly once, so it cannot catch a regression that only COMPOUNDS over
+        // a sequence of removals — the documented loop-safe contract. This test
+        // removes two non-last nodes in sequence, and crucially the SECOND node
+        // removed is one that was itself DISPLACED (swap-relocated) by the first
+        // removal. A stale-map regression therefore compounds: the first removal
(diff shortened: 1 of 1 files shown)


dutiona:
**[Adversarial Review] — agy (Gemini 3.1 Pro) + lens subagents (correctness, test-adequacy)**

agy found a **BLOCKER the original fix missed**: `remove_edges_by_fact` had the *same* petgraph swap-remove invalidation (+ a self-loop double-delete) that the PR fixed for `remove_node` — folded in (descending-sort + dedup). Re-review confirmed both fixes correct for all enumerated cases (last/single/self-loop/sequential). test-adequacy gap closed with a compounding sequential-removal loop test. All regression tests proven to fail without their fix.

PR #835, opened 2026-06-26
test(retrieval): cover include_expired_probe diagnostics wiring end-to-end

### LINKED ISSUE #324: [super-qa] high: `include_expired_probe` contract has no end-to-end test

**`include_expired_probe` contract has no end-to-end test**

- **ID:** `search/testing-expired-probe-e2e`
- **Severity:** high
- **Area:** retrieval · **Type:** test
- **Location:** `src/search/query.rs:200-203, src/engine/query.rs:253-262`
- **Auto-fixable:** no

## Description
The `include_expired_probe` flag is tested only as a builder-setter (`query.rs:381-383`). No test anywhere in `src/` or `tests/` asserts that `diagnostics.expired_matches` is actually populated with a non-None count when `include_expired_probe = true` and facts have been expired, or that it is `None` for a vector-only query (documented invariant at `query.rs:59`). The execution path through `engine/query.rs:253-262` that calls `fts_count_expired` and writes `diagnostics.expired_matches = Some(expired_count)` is never exercised by a test that inspects the returned value.

## Suggested approach
Add an integration test (in `tests/` or `src/engine/tests.rs`) that:
1. Adds two facts matching a keyword, expires one.
2. Calls `execute_query` with `MemoryQuery::new().text(keyword).include_expired_probe()`.
3. Asserts `response.diagnostics.expired_matches == Some(1)`.
4. Adds a second sub-test with embedding-only query and asserts `expired_matches == None`.

## References
—

dutiona:
## What

Adds three engine-level (full-stack) tests to `engine/tests.rs` that exercise the `include_expired_probe` diagnostics contract through `execute_query`:

1. **Text query + probe** — after expiring one of two facts sharing a keyword, `diagnostics.expired_matches == Some(1)` and only the active fact is returned.
2. **Text query, no probe** — `expired_matches == None` even though a matching expired fact exists (probe stays off unless opted in).
3. **Vector-only query + probe** — `expired_matches == None`; a query with no text has no FTS5 terms to probe (the documented limitation in `search/hybrid.rs`).

## Why

The flag was previously covered only as a builder setter (`search/query.rs`). The execution path in `engine/query.rs` that calls `lexical_count_expired` and writes `QueryDiagnostics::expired_matches` was never asserted end-to-end, so the contract — `Some` only when the probe is opted in **and** the query carries text, `None` otherwise — could silently regress.

Facts are inserted via the engine's normal `add_fact` write path so they land in the FTS5 index the probe's `MATCH` query scans.

## Verification

- `cargo test -p memory-engine` — 1219 passed.
- `cargo clippy -p memory-engine --all-targets` — no new warnings (the 3 remaining warnings pre-exist on `main` in unrelated `storage/` files).
- `cargo fmt` — clean.
- Confirmed the positive test fails (`None` vs `Some(1)`) when the wiring is temporarily broken, proving it genuinely exercises the path.

Closes #324.
Part of #257.

### DIFF
--- a/memory-engine/lib/memory-engine/src/engine/tests.rs
+++ b/memory-engine/lib/memory-engine/src/engine/tests.rs
@@ -5529,7 +5529,7 @@ mod expired_probe_e2e {
         let expired_id = add_indexed_fact(&engine, "quasar discovered last year").await;
         engine
             .storage()
-            .expire_fact(expired_id, Utc::now())
+            .expire_fact(expired_id, Utc::now() - chrono::Duration::hours(1))
             .await
             .unwrap();
 
@@ -5557,7 +5557,7 @@ mod expired_probe_e2e {
         let expired_id = add_indexed_fact(&engine, "pulsar timing glitch").await;
         engine
             .storage()
-            .expire_fact(expired_id, Utc::now())
+            .expire_fact(expired_id, Utc::now() - chrono::Duration::hours(1))
             .await
             .unwrap();
 
@@ -5582,7 +5582,7 @@ mod expired_probe_e2e {
         let expired_id = add_indexed_fact(&engine, "nebula collapses inward").await;
         engine
             .storage()
-            .expire_fact(expired_id, Utc::now())
+            .expire_fact(expired_id, Utc::now() - chrono::Duration::hours(1))
             .await
             .unwrap();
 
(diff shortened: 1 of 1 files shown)


dutiona:
**[Adversarial Review] — agy (Gemini 3.1 Pro, pinned diff) + lens subagent (test-quality)**

Caught a stray debug `println!` left in `query.rs` by a concurrent-session stash collision (discarded; verified clean). The two `None`-expecting tests were tautological (would pass even if the wiring were deleted) — strengthened with a `Some(0)` case that proves the probe both ran AND is FTS-term-restricted. Expiry timestamps moved to a guaranteed-past value to remove a clock-resolution race.

Issue #258, opened 2026-06-01
[super-qa] Area: storage — code-quality findings


dutiona:
All `/super-qa` 2026-06-01 findings in **area:storage**. Non-autofixable blocker/critical/high/medium are individual sub-issues; low/info are grouped in one sub-issue.

dutiona:
## Epic complete — `[super-qa] Area: storage` fully cleared ✅

All **65 sub-issues closed**. Full `/address-issue` pipeline (triage-vs-current-main → plan → implement → adversarial review [agy + diverse-lens subagents + workflows] → `/finish-pr` → rebase-before-merge → squash-merge), run in collision-ordered waves.

| Wave | Scope | Outcome |
|------|-------|---------|
| 0 | 10 dead findings | closed with evidence (already-fixed / FP) |
| 1 | 22 findings | 11 PRs merged |
| 2 | 26 findings | 10 PRs merged |
| 3 | #847 / #879 / #405 | #912 / #913 / #906 (+ Docs-CI hotfix #917) |
| 4 | #280 schema god-split | #918 (`store/schema/` submodules, byte-identical move) |
| 5 | #114 `Edge.relation_type` | #921 (`RelationType` open enum: canonicalizing `From`, semantic `Eq`/`Hash`, byte-identical wire) |
| 6 | #485 + #120 test-harness | #922 (shared `test_utils`, FK-pragma fix + mutation-proof guard, `MockEmbedder` consolidation) |
| 7 | #500 (29 low/info) | #924 (docs + perf/flake fixes + coverage) + dispositioned (6 fixed / 4 won't-fix / 6 deferred → #923) |

**Collateral filed during the sweep:** #849, #859, #860, #898, #901, #911, #919, #923.

Engine-wide invariants upheld throughout: `unsafe` forbidden, no `unwrap` in lib code, `clippy --workspace --all-targets --all-features -D warnings`, `cargo doc` intra-doc-link gate, `cargo test --workspace --all-features`. Closing.

PR #840, opened 2026-06-26
refactor(storage): inspect dump — tmp_path returns Result + structured snapshot serializer

### LINKED ISSUE #338: [super-qa] medium: tmp_path() silently degrades when path.file_name() is None

**tmp_path() silently degrades when path.file_name() is None**

- **ID:** `inspect/correctness-tmp-path-silent-fallback`
- **Severity:** medium
- **Area:** storage · **Type:** bug
- **Location:** `src/inspect/dump.rs:20-24`
- **Auto-fixable:** no

## Description
tmp_path uses `path.file_name().unwrap_or_default().to_os_string()` (line 21). For paths where file_name() returns None (e.g. '/', '..', a path ending with '/'), the default empty OsStr is used, name becomes ".tmp", and `path.with_file_name(".tmp")` produces a sibling at the parent directory rather than a temp file next to the intended target. dump_json/dump_json_gzip/dump_json_zstd then write to the wrong location; the subsequent fs::rename either overwrites an unintended file or fails with a misleading I/O error. The atomic-write-then-rename safety invariant is silently lost.

## Suggested approach
Change tmp_path to return Result<PathBuf>: `fn tmp_path(path: &Path) -> Result<PathBuf> { let name = path.file_name().ok_or_else(|| MemoryError::Internal(format!("dump path has no file name: {}", path.display())))?; let mut name = name.to_os_string(); name.push(".tmp"); Ok(path.with_file_name(name)) }` and propagate the error in each caller.

## References
CWE-22

### LINKED ISSUE #352: [super-qa] medium: JSON object header hand-assembled with raw `write!` instead of a structured serializer — schema contract held

**JSON object header hand-assembled with raw `write!` instead of a structured serializer — schema contract held only by a test**

- **ID:** `inspect/design-manual-json-header-in-stream-snapshot`
- **Severity:** medium
- **Area:** storage · **Type:** refactor
- **Location:** `src/inspect/dump.rs:43-73`
- **Auto-fixable:** no

## Description
`stream_snapshot` at line 43 manually constructs the JSON object header:
```rust
write!(writer, r#"{{"schema_version":{schema_version},"storage_epoch":{storage_epoch},"embed_dim":{embed_dim}"#)?;
```
and then calls `stream_for_each` for each collection. The field names and order are duplicated relative to `EngineSnapshot`'s serde-derived representation. If `EngineSnapshot` fields are renamed or reordered (e.g., a new top-level scalar is added for Phase 5), the streaming path silently diverges. The only thing preventing a mismatch is the `streaming_output_matches_snapshot_schema` test at dump.rs:278 — a test, not a compile-time constraint.

A fully type-safe streaming approach would use `serde_json::Serializer` with `serialize_map` for the object root, serializing scalar fields normally and collections via `stream_for_each`. Alternatively, a `SnapshotHeader { schema_version, storage_epoch, embed_dim, config }` struct carrying only scalars could be fully serde-serialized, with collections streamed afterwards — but this requires changing the JSON structure.

Current state: raw `write!` macro assembles JSON field names.
Target state: use serde's `Serializer`/`serialize_map` API so field names track the struct definition.
Refactoring pattern: Replace Hand-Rolled Serialization with Library Serializer.
Motivation: correctness, schema evolution safety.
Estimated scope: `src/inspect/dump.rs` only; restore.rs and the snapshot format are unchanged.
Prerequisites: familiarity with `serde_json::ser::Serializer`.
Risk: medium — the streaming guarantee (O(1) peak memory) must be preserved; the serde map API supports this but requires care.

## Suggested approach
// Sketch using serde_json::Serializer + serialize_map:
use serde::ser::SerializeMap;
let mut ser = serde_json::Serializer::new(&mut *writer);
let mut map = ser.serialize_map(None)?;
map.serialize_entry("schema_version", &schema_version)?;
map.serialize_entry("storage_epoch", &storage_epoch)?;
map.serialize_entry("embed_dim", &embed_dim)?;
// stream facts array: collect into a helper type that implements Serialize via streaming

## References
—

dutiona:
Resolves two `/super-qa` findings (epic #258) in `memory-engine/lib/memory-engine/src/inspect/dump.rs`. Both are correctness/schema hardening of the snapshot-dump path. Scope: this one file only.

## #338 — `tmp_path` silently degraded when `path.file_name()` is `None`
The bare `file_name().unwrap_or_default()` fell back to an empty `OsStr`, so a path with no final component (the root `/`, `..`, `.`) produced a `.tmp` sibling in the **parent** directory instead of next to the intended target. The follow-up `fs::rename` then overwrote an unrelated file or failed with a misleading I/O error — silently losing the atomic write-then-rename invariant.

**Change:** `tmp_path` now returns `Result<PathBuf>`, erroring with `MemoryError::Internal` when `file_name()` is `None`; the three `dump_json{,_gzip,_zstd}` callers propagate via `?`.

Note vs. the original finding: Rust's `Path` normalizes a trailing slash away (`/var/data/` reports `data`), so the genuine `None` cases are `/`, `..`, `.` — verified empirically and reflected in the doc + test.

## #352 — hand-rolled `write!` JSON header replaced with a structured serializer
The snapshot object header was assembled with raw `write!` (braces and field names typed by hand), duplicating `EngineSnapshot`'s serde representation; only the runtime `streaming_output_matches_snapshot_schema` test prevented drift.

**Change:** the whole object is now serialized through serde's `SerializeMap` — every field (leading scalars **and** the streamed collections) goes through `map.serialize_entry`, so field names track the struct by construction. Collections still stream row-by-row via a new `SeqStreamer<T, F>` `Serialize` adapter that drives each store's `for_each` through `SerializeSeq`, preserving the O(1)-per-collection peak-memory guarantee. serde_json's compact output is byte-identical to the previous hand-rolled bytes.

### HARD INVARIANT preserved
`embed_dim` stays the **third leading scalar**, before any collection, so the CLI's byte-capped `peek_embed_dim_from_reader` finds it. The `serialize_entry` ordering is fixed to match `EngineSnapshot`, and a new `streaming_header_keeps_embed_dim_leading` test asserts both the field order and that `embed_dim` sits within the byte cap. The existing `streaming_output_matches_snapshot_schema` and the two `insta` snapshot tests still pass — confirming byte-level format stability.

## Test plan
Run from the worktree:
- `cargo fmt --all` + `cargo fmt --check` — clean
- `cargo build --workspace` — clean (1 pre-existing warning in `storage/sqlite/consolidation.rs`, untouched)
- `cargo test -p memory-engine inspect::dump` — 12 passed (incl. the 3 new/round-trip + 2 insta snapshot tests)
- `cargo test --workspace` — 1567 passed, 5 ignored (incl. the CLI `peek_embed_dim_from_reader` tests — byte compatibility confirmed)
- `cargo clippy --workspace --all-targets` — my file clean; 3 remaining warnings are all in untouched files outside this PR's scope (pre-existing on main)

Fixes #338, Fixes #352

### DIFF
--- a/memory-engine/lib/memory-engine/src/inspect/dump.rs
+++ b/memory-engine/lib/memory-engine/src/inspect/dump.rs
@@ -5,6 +5,8 @@ use std::path::Path;
 
 use rusqlite::Connection;
 use serde::Serialize;
+use serde::Serializer as _;
+use serde::ser::SerializeMap as _;
 
 use crate::error::{ConflictError, MemoryError, Result};
 use crate::inspect::types::{EmbeddingSpaceSnapshot, FactVectorSnapshot};
@@ -18,10 +20,22 @@ use crate::store::scopes::ScopeStore;
 use crate::store::summaries::SummaryStore;
 
 /// Build a sibling temporary path for atomic write-then-rename.
-fn tmp_path(path: &Path) -> std::path::PathBuf {
-    let mut name = path.file_name().unwrap_or_default().to_os_string();
+///
+/// # Errors
+///
+/// Returns [`MemoryError::Internal`] if `path` has no final component — i.e.
+/// `Path::file_name()` is `None`, which happens for the root `/`, `..`, and `.`
+/// (a trailing slash is normalized away, so `/var/data/` still reports `data`).
+/// Falling back to an empty file name in that case would silently produce a
+/// `.tmp` sibling in the *parent* directory rather than next to the intended
+/// target, breaking the atomic write-then-rename invariant.
+fn tmp_path(path: &Path) -> Result<std::path::PathBuf> {
+    let name = path.file_name().ok_or_else(|| {
+        MemoryError::Internal(format!("dump path has no file name: {}", path.display()))
+    })?;
+    let mut name = name.to_os_string();
     name.push(".tmp");
-    path.with_file_name(name)
+    Ok(path.with_file_name(name))
 }
 
 /// Create (or truncate) the dump's temporary write target, refusing to follow a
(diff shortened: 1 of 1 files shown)


dutiona:
**Review thread resolution (pre-merge):**
- *SeqStreamer Box alloc (gemini):* fixed — callback is now `&mut (dyn FnMut(T)->Result<()> + '_)` (`F: FnOnce(&mut dyn FnMut...)`), removing the per-collection heap allocation; closure lives on the serializer stack.
- *Store-error variant lost as Serialization (codex P2):* confirmed real and fixed — there was no recovery path, so `MemoryError::Database`/`Io` from a streamed iterator collapsed to `Serialization`. Added a `store_err` RefCell that stashes the original typed error before the `serde::ser::Error::custom` funnel; `stream_snapshot` recovers it on serde failure, preserving the `dump_json` error contract. Test strengthened to assert `matches!(err, MemoryError::Io(_))`.

Issue #847, opened 2026-06-26
fix(storage): archive() fails with FK violation when a boundary edge crosses the archive set


dutiona:
## Summary
`MemoryEngine::archive` fails with `FOREIGN KEY constraint failed` when any edge has **exactly one** endpoint in the archive candidate set (a 'boundary' edge: archived → surviving, or vice-versa).

## Root cause
`commit_archive_atomic` (`memory-engine/lib/memory-engine/src/storage/sqlite/cold_storage.rs:119-121`) deletes edges then facts:
```
EdgeStore::new(&tx).hard_delete_by_facts(&fact_ids)?;  // BOTH endpoints in set only
FactStore::new(&tx, dim).hard_delete_ids(&fact_ids)?;
```
`hard_delete_by_facts` (`store/edges.rs:307-`) deletes only edges where **both** endpoints are archived (by design — it must not touch edges to live facts). But then `hard_delete_ids` hard-deletes the archived facts, and a *surviving* boundary edge still references one of them via `edges.source_fact_id/target_fact_id REFERENCES facts(id)` → FK violation, aborting the whole archive transaction.

## Evidence (reproducible)
A test that inserted an edge `archived_fact → survivor_fact` and then ran `archive()` panicked with `Storage(Backend("FOREIGN KEY constraint failed"))` inside `commit_archive_atomic`. (The #332 PR's consistency test deliberately avoids boundary edges to stay in scope.)

## Suggested fix
Decide the intended semantics for boundary edges and make `commit_archive_atomic` consistent:
- Option A: **expire** (soft-delete) boundary edges before hard-deleting the facts (set `t_expired`), keeping the audit trail and satisfying the FK.
- Option B: hard-delete *all* edges incident to any archived fact (both internal and boundary), accepting that a surviving fact loses its link to the archived one.
A says 'a survivor remembers it was once linked'; B says 'the link is gone with the fact'. Likely A (matches the engine's soft-deletion philosophy). Add a test that archives with a boundary edge present and asserts archive succeeds.

Surfaced during #332; not folded in (out of that PR's scope).

Issue #846, opened 2026-06-26
fix(retrieval): MemoryGraph::remove_node corrupts surviving nodes' cached NodeIndex (petgraph index swap)


dutiona:
## Summary
`MemoryGraph::remove_node` (`memory-engine/lib/memory-engine/src/graph/memory_graph.rs:85-90`) calls petgraph's `Graph::remove_node(idx)`, which removes the node by **swapping the last node into the freed slot** — invalidating the swapped node's `NodeIndex`. `MemoryGraph` only removes the *target* fact's `node_map` entry; it does **not** update the swapped node's `node_map` entry to its new index. After a multi-node removal, any *surviving* node that got swapped holds a **stale `NodeIndex`** in `node_map`, so all its queries (`neighbors`, `degree`, `connected_component`) silently return wrong results.

## Evidence (reproducible)
While implementing #332 (PR for archive's post-commit graph update), a test that archived 20 expired facts while keeping 2 active survivors with a survivor→survivor edge found that, under the old node-by-node `remove_node` loop, `graph_neighbors(survivor_a)` returned `[]` instead of `[survivor_b]` — the survivor's edge was silently lost because its cached `NodeIndex` was invalidated by the swaps. `graph_has_node` still returned `true` (it only checks the `node_map` key), masking the corruption.

## Blast radius
Every caller of `remove_node` in a context where other nodes survive:
- `engine/archive.rs` (fixed in #332's PR by rebuilding the cache from the DB instead of mutating node-by-node — that PR sidesteps the bug but does not fix `remove_node` itself)
- `engine/conflict.rs`, `cycle/apply.rs`, `forgetting/policy.rs` (all call `remove_edges_by_fact`/`remove_node`)

## Suggested fix
After `self.graph.remove_node(idx)`, petgraph returns nothing but the swapped node is whatever was at the *last* index. Either (a) look up the node that now occupies `idx` and patch its `node_map` entry, or (b) prefer the codebase's established source-of-truth reconciliation — rebuild from `list_active_edges` (as `consolidate` and now `archive` do) at call sites that remove in bulk. A direct `remove_node` fix should re-sync `node_map` for the swapped node and add a unit test that removes a low-index node and asserts a high-index survivor's edges are intact.

Surfaced during #332; not folded in (out of that PR's scope).

dutiona:
[...]

PR #848, opened 2026-06-26
fix(storage): make post-commit in-memory graph update panic-safe (archive 2-phase consistency)

### LINKED ISSUE #332: [super-qa] medium: In-memory graph not updated when commit_archive succeeds but graph.write() is interrupted

**In-memory graph not updated when commit_archive succeeds but graph.write() is interrupted**

- **ID:** `archive/correctness-graph-state-after-commit`
- **Severity:** medium
- **Area:** storage · **Type:** bug
- **Location:** `src/engine/archive.rs:82-88`
- **Auto-fixable:** no

## Description
After a successful `commit_archive` (facts hard-deleted from SQLite), the code acquires `self.graph.write()` and removes the corresponding nodes and edges. If the process panics or is killed between the SQLite commit and the graph removal loop, the in-memory graph retains stale nodes pointing to facts that no longer exist in the database. On the next engine load the graph is rebuilt from the DB (facts are gone), so persistence survives a restart — but within a single session, any graph query that hits the stale nodes will encounter dangling references. While `parking_lot::RwLock` does not poison, a panic inside the `graph.write()` lock block would drop the lock guard and leave the graph partially pruned.

## Suggested approach
This is inherent to a two-phase commit (SQLite + in-memory). The safest mitigation is to make the graph update loop panic-safe by collecting the fact IDs into a local Vec before entering the lock and wrapping the loop in a catch_unwind guard, or by rebuilding the graph lazily on next access when a missing node is detected.

## References
—

dutiona:
## Summary

`MemoryEngine::archive` performs an inherent two-phase update: an atomic SQLite commit (hard-delete facts + edges) followed by an update to the in-memory graph cache. super-qa finding #332 flagged that a panic/kill between the committed transaction and the graph update leaves stale in-memory nodes for the rest of the session.

While addressing it, a stronger problem surfaced: the old node-by-node prune was also **silently corrupting surviving nodes**. petgraph's `Graph::remove_node` swaps the last node into the freed slot, invalidating an unrelated `NodeIndex` that `MemoryGraph::node_map` still caches. Removing archived nodes therefore left survivors with stale indices — `graph_neighbors`/`degree`/`connected_component` returned wrong results while `graph_has_node` still reported `true`, masking the corruption.

## What changed (`memory-engine/lib/memory-engine/src/engine/archive.rs`)

- Replaced the post-commit node-by-node graph prune with the **source-of-truth-driven reconciliation** the engine already uses after a bulk expiry in `consolidate`: read the committed active edge set via the storage port, then rebuild the cache wholesale with `MemoryGraph::from_active_edges`.
  - **#332 (two-phase consistency):** correct by construction and self-healing — a kill anywhere in the path leaves the committed DB as the source of truth, and the next `open` rebuilds the cache from the exact same state. Panic-safe: one port read + one assignment, with no lock held across the `.await` (the read happens before the write guard is taken).
  - **Correctness bonus:** sidesteps the petgraph index-swap corruption entirely (no in-place node removal).
- Updated the module/method docs to describe the rebuild-from-DB reconciliation.
- Added `archive_keeps_in_memory_graph_consistent`: archives 20 expired facts while keeping 2 active survivors with a survivor→survivor edge, then asserts archived nodes are gone, survivors and their edge remain intact (this assertion fails under the old prune via the swap bug), and the graph holds no dangling references to archived ids.

## Collateral (filed separately, not folded in — out of this PR's single-file scope)

- **#846** (`type:bug`, `area:retrieval`, high) — `MemoryGraph::remove_node` corrupts surviving nodes' cached `NodeIndex` via the petgraph swap. This PR makes `archive` immune by not using it; the underlying defect (also reachable from `conflict`/`forgetting`/`cycle`) still needs a direct fix.
- **#847** (`type:bug`, `area:storage`, medium, linked under epic #258) — `archive()` fails with a FK violation when a boundary edge (exactly one archived endpoint) exists, because `hard_delete_by_facts` only removes edges with both endpoints archived but then the fact hard-delete trips the surviving boundary edge's FK.

## Test plan (all run from the worktree)

- `cargo fmt --all` + `cargo fmt --all --check` → clean
- `cargo build --workspace` → 0 errors (1 pre-existing warning in `consolidation.rs`, untouched)
- `cargo test -p memory-engine` → 1216 passed
- `cargo test -p memory-engine --all-features` → 1280 passed, 70 ignored (PG testcontainers)
- `cargo test --workspace` → 1564 passed, 5 ignored
- `cargo clippy --workspace --all-targets` → 0 errors (3 pre-existing default-feature warnings in untouched files; CI runs `--all-features`)
- `cargo clippy -p memory-engine --all-features --all-targets` → no issues

Fixes #332

### DIFF
--- a/memory-engine/lib/memory-engine/src/engine/archive.rs
+++ b/memory-engine/lib/memory-engine/src/engine/archive.rs
@@ -1,8 +1,8 @@
 //! Archival compression orchestration for [`MemoryEngine`].
 //!
 //! Moves expired, non-pinned facts into `.pak` files (zstd + blake3),
-//! records a manifest row, hard-deletes from `SQLite`, and prunes the
-//! in-memory graph — all in a crash-safe sequence.
+//! records a manifest row, hard-deletes from `SQLite`, and rebuilds the
+//! in-memory graph cache from the committed DB — all in a crash-safe sequence.
 
 use std::path::PathBuf;
 
@@ -14,6 +14,7 @@ use crate::archive::types::{
     CURRENT_PAK_VERSION,
 };
 use crate::error::{ArchiveError, MemoryError, Result};
+use crate::graph::MemoryGraph;
 use crate::store::schema::CURRENT_SCHEMA_VERSION;
 use crate::types::{Edge, Fact};
 
@@ -24,8 +25,10 @@ impl MemoryEngine {
     ///
     /// Returns `None` if fewer than `policy.min_facts` candidates exist.
     /// Otherwise writes the `.pak`, inserts a manifest row, hard-deletes
-    /// facts and edges from `SQLite` (single transaction), and updates the
-    /// in-memory graph.
+    /// facts and edges from `SQLite` (single transaction), then rebuilds the
+    /// in-memory graph cache from the committed active edge set — the DB is the
+    /// source of truth, so the cache is reconciled to it rather than mutated in
+    /// place (#332).
     ///
     /// # Panics
     ///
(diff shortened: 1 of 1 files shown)


dutiona:
**Review thread resolution (pre-merge):**
- *Archive graph-rebuild lost-update window (codex P2):* resolved by reverting the approach. The off-lock `list_active_edges`+swap rebuild this flags has been replaced by an in-place prune under a single `graph.write()` (atomic, no `.await` in the critical section), so the lost-update window is eliminated. `MemoryGraph::remove_node` safely re-indexes `node_map` on petgraph swap-remove (#833). The same read-then-swap pattern in `consolidate` is tracked separately in #860. Thanks!

Issue #843, opened 2026-06-26
chore(build): clean up clippy 1.96 lint drift in storage lib


dutiona:
## Pre-existing lint drift in the storage lib (clippy 1.96 / rustc)

Surfaced during the #256 (area:mcp super-qa) sweep — **three independent review agents** working on PRs #832/#834/#836 each reported the same baseline warnings on `main` (`41f27e3`). Confirmed by direct inspection; these are in lib files none of those PRs touched, so they are pre-existing, not introduced.

| Warning | Location | Evidence |
|---|---|---|
| `unused variable: embedding` | `memory-engine/lib/memory-engine/src/storage/sqlite/consolidation.rs:648` | `let embedding = fact.embedding.clone();` — bound but never read |
| `clippy::needless_pass_by_value` | `memory-engine/lib/memory-engine/src/storage/sqlite/mod.rs:205` | `mut self` taken by value in a builder-style fn |
| dead method (reported) | `memory-engine/lib/memory-engine/src/storage/factory.rs:55` | flagged by reviewers; verify |

These don't fail CI today (CI pins an older stable than the local clippy 1.96 — see the worktree-tooling lint-drift note), but they will surface the moment CI's toolchain advances, and the unused-var is a real smell regardless of clippy version.

### Task
Clean up the three. For the unused `embedding`, determine whether it should be used (a dropped write?) or removed — don't just `_`-prefix without checking intent.

Issue #256, opened 2026-06-01
[super-qa] Area: mcp — code-quality findings


dutiona:
All `/super-qa` 2026-06-01 findings in **area:mcp**. Non-autofixable blocker/critical/high/medium are individual sub-issues; low/info are grouped in one sub-issue.

dutiona:
## Reconciliation sweep — triaged against current `main` (2026-06-26)

All 20 sub-issue findings (super-qa snapshot 2026-06-01) re-verified against current code. The crate moved to `memory-engine/bin/mcp/` (#814) and `tools/mod.rs` grew **1386 → 1996 lines**, so every cited line number was stale and was re-anchored by symbol.

### Dispositions (evidence-backed)

| # | Disposition | Evidence on current `main` |
|---|---|---|
| 266 | **open** | `handle_flush_insights` @1027 — no `insights` cap |
| 267 | **open** (canonical for bootstrap-cap) | `handle_bootstrap_session` @1500 — no `jsonl_data` cap |
| 277 | **open (worse)** | `tools/mod.rs` = 1996 lines |
| 294 | **open** | `parse_embedding` @532 — `from_value::<Vec<f32>>` before any dim/len check |
| 296 | **open** (canonical for dump-path) | `handle_dump_state` @1322; parent-only canonicalize @1344; original `p` used @1351 |
| 317 | **partial** | `#[cfg(test)]` block exists @1826 (10 tests) but parser cases missing |
| 318 | **open** | 6 tools → 0 dispatch tests (grep `<none>`) |
| 319 | **open** → folds into cap PR | DoS paths untested |
| 320 | **open** | dump test @1990 checks default-path only, not rejection |
| 321 | **partial** | `fuzz/` crate exists (6 targets) but none hit `tools::dispatch` |
| 339 | **open** | `get_usize` @511 = `try_from(v).ok()` (negative → silent `None`) |
| 353 | **partial false-positive** | `parse_fact_type` @629 already #678 core-`FromStr` (dual-casing); only `parse_event_type`@614 / `parse_outcome`@1567 remain |
| 354 | **dup of 296** | same code |
| 355 | **dup of 267** | same code |
| 381 | **✅ already-fixed → closed** | field read `main.rs:172`, validated `main.rs:211` |
| 414 | **dup of 296/354** | same code |
| 469 | **open** | `summary.rs` 3-branch parser, no wiremock tests |
[...]

PR #863, opened 2026-06-26
refactor(mcp): EventType/Outcome parsing via core FromStr SSOT

### LINKED ISSUE #353: [super-qa] medium: Hand-rolled string→enum parsers duplicate what serde already provides on EventType/FactType/Outcome

**Hand-rolled string→enum parsers duplicate what serde already provides on EventType/FactType/Outcome**

- **ID:** `mcp/design-stringly-typed-enums`
- **Severity:** medium
- **Area:** mcp · **Type:** refactor
- **Location:** `/home/mroynard/dev/memory-engine/memory-engine-mcp/src/tools/mod.rs:488-505`
- **Auto-fixable:** no

## Description
Current state: `parse_event_type`, `parse_fact_type`, and `parse_outcome` are manual `match s { ... => Ok(...), other => Err(...) }` functions. `EventType`, `FactType`, and `Outcome` already derive `serde::Deserialize`. Target state: replace each parser with `serde_json::from_value::<T>(Value::String(s.to_owned()))` or add a `FromStr` impl delegating to serde, then call `T::from_str(s).map_err(|_| ...)` at each call site. Refactoring pattern: Delete Dead Code / Delegate to Existing Abstraction. Motivation: maintainability — when a new variant is added to `EventType`, the MCP parser must be updated separately; the compiler won't warn about the mismatch. Estimated scope: ~45 LOC removed across `tools/mod.rs`. Prerequisites: verify serde variant names match the JSON strings sent by clients (they do — both use PascalCase without `rename`). Risk: LOW.

## Suggested approach
// Instead of parse_event_type(s):
serde_json::from_value::<EventType>(serde_json::Value::String(s.to_owned()))
    .map_err(|_| ValidationError::UnknownEventType(s.to_owned()))
// Identical pattern for FactType and Outcome.

## References
—

dutiona:
## Summary

Extends the #678 `FromStr`-in-core single-source-of-truth pattern (already applied to `FactType`) to `EventType` and `Outcome`. The MCP `parse_event_type` / `parse_outcome` were hand-rolled exact-match `match` statements duplicating the string→enum mapping; they now delegate to canonical core `FromStr` impls.

## Changes

**Core lib (`types/events.rs`)**
- Add case-insensitive `FromStr` for `EventType` and `Outcome`, mirroring `FactType::from_str`. They accept the PascalCase wire form the JSON schemas advertise **and** the snake_case/lowercase `Display` form, so `to_string` round-trips.
- Add a `Display` impl for `EventType` (it had none) so the round-trip holds.
- `EventType::from_str` is **complete** — it parses `OutcomeSignal` too; an incomplete parser that cannot round-trip `Display` is a latent bug.
- New error types `ParseEventTypeError` / `ParseOutcomeError` preserve the offending token, like `ParseFactTypeError`.
- These lenient consumer parsers stay **orthogonal** to the strict, exact-match DB serialization in `store::events` (which must round-trip bytes verbatim) and to the serde derives — same separation `FactType` documents.

**MCP (`tools/mod.rs`)**
- `parse_event_type` / `parse_outcome` now delegate via `s.parse()`, preserving their existing error types (`ValidationError::UnknownEventType`, the outcome `ErrorData`) and messages.
- `parse_event_type` keeps one MCP-specific narrowing: it **rejects `OutcomeSignal`** (a system-generated event, deliberately omitted from the ingest/replay JSON schemas) at the boundary, so the schema contract is unchanged. The core parser is complete; the boundary gate enforces the surface policy.

## Tests
- Core: 10 new unit tests (PascalCase / snake_case Display form / case-insensitivity / whitespace-strictness / unknown-token-preserved / Display↔FromStr round-trip) for both `EventType` and `Outcome`.
- MCP: 7 new unit tests for the delegating `parse_event_type` / `parse_outcome`, including the `OutcomeSignal`-rejected-at-boundary case.

## Gate
- `cargo test -p memory-engine -p memory-engine-mcp` — all green
- `cargo clippy -p memory-engine -p memory-engine-mcp --all-targets` — clean (only the 2 pre-existing `consolidation.rs`/`sqlite/mod.rs` storage warnings tracked as #843)
- `cargo fmt --check` — clean
- `cargo build --workspace` — clean

Closes #353. Part of #256.

### DIFF
--- a/memory-engine/bin/mcp/src/tools/mod.rs
+++ b/memory-engine/bin/mcp/src/tools/mod.rs
@@ -683,11 +683,27 @@ fn parse_search_mode(s: &str) -> Result<SearchMode, ErrorData> {
 /// it. The core parser is complete (it accepts every variant); this boundary gate
 /// preserves the schema contract without re-implementing the variant mapping.
 fn parse_event_type(s: &str) -> Result<EventType, ValidationError> {
-    match s.parse() {
-        Ok(EventType::OutcomeSignal) | Err(_) => {
-            Err(ValidationError::UnknownEventType(s.to_owned()))
-        }
-        Ok(et) => Ok(et),
+    // The system-only-reject arm and the unparseable arm share a body, but keeping
+    // them separate is deliberate: it documents the two distinct rejection reasons
+    // and keeps the `EventType` match exhaustive (no `Ok(_)` catch-all), so a new
+    // variant forces a deliberate allow/reject decision here at compile time.
+    #[allow(clippy::match_same_arms)]
+    match s.parse::<EventType>() {
+        // User-facing types — accepted at the MCP ingest/replay boundary. These are
+        // exactly the variants the `ingest` / `replay` JSON schemas advertise.
+        Ok(
+            et @ (EventType::Interaction
+            | EventType::ToolCall
+            | EventType::MemoryOp
+            | EventType::SystemEvent),
+        ) => Ok(et),
+        // System-generated only — emitted internally by `record_outcome`, never
+        // user-ingestible; the schemas deliberately omit it.
+        Ok(EventType::OutcomeSignal) => Err(ValidationError::UnknownEventType(s.to_owned())),
+        // NOTE: intentionally NO catch-all `Ok(_)`. Adding a new `EventType` variant
+        // must force a deliberate allow/reject decision here — the compiler flags
+        // the non-exhaustive match instead of silently making it user-ingestible.
+        Err(_) => Err(ValidationError::UnknownEventType(s.to_owned())),
     }
 }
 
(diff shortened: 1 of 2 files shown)


dutiona:
### [Antigravity Review] (agy — Gemini 3.1 Pro High)

Adversarial pass on the EventType/Outcome SSOT refactor.

**Verified CLEAN:** the lenient/case-insensitive `FromStr` reaches ONLY the two MCP edge parsers (grep-confirmed) — DB persistence, serde derives, and `EventFilter` stay exact-match; the new `Display` (snake_case) does not leak into any wire/serialization path; the new `Parse*Error` types are additive (no API break).

**Found (fixed in `c56c248`):** the `parse_event_type` gate was a *negative blocklist* — it rejected `OutcomeSignal` but accepted everything else via a catch-all `Ok(et) => Ok(et)`, so a future system-only `EventType` variant would silently become user-ingestible. Converted to an **exhaustive match** (positive allowlist of the 4 user-facing variants + explicit `OutcomeSignal` reject, no catch-all) — adding a variant is now a compile error at the gate, forcing a deliberate allow/reject decision (deny-by-default).

Issue #560, opened 2026-06-15
refactor(core): complete MemoryError typed-error hierarchy (remaining String variants)


dutiona:
## Complete the `MemoryError` typed-error hierarchy

#115 (PR #559) established the pattern: `Conflict(String)` → `Conflict(ConflictError)` — a `thiserror` sub-enum whose variants are **derived from real call sites**, with byte-preserved Display messages. This issue tracks extending that pattern toward a full typed hierarchy for the remaining `String`-wrapped `MemoryError` variants.

### Remaining stringly-typed variants (post-#115)

| Variant          | ~Sites | Consumer pattern-match value             |
| ---------------- | ------ | ---------------------------------------- |
| `Internal`       | 45     | Low — "shouldn't happen", propagated     |
| `Migration`      | 36     | Low — fatal/terminal                     |
| `Archive`        | 16     | **Medium** — cold-storage failure modes  |
| `Reranker`       | 11     | **Medium** — provider failure vs timeout |
| `Lineage`        | 10     | Low–medium                               |
| `NotImplemented` | 6      | Low                                      |
| `Pool`           | 2      | Low                                      |
| `Bootstrap`      | 1      | Low                                      |

### Approach

Apply the `ConflictError` pattern **variant-by-variant, where consumer pattern-matching adds value** — likely `Archive` and `Reranker` first. For the terminal/internal variants (`Internal`, `Migration`), prefer **documenting that they are not meant to be matched** over splitting ~80 sites for little benefit. Each split is independent and high-ripple → separate PRs.

Pattern reference: `ConflictError` in `src/error.rs` (PR #559). All variants already carry doc comments (#496).

Follow-up to #115 · part of epic #241.

dutiona:
## Design note: which `MemoryError` variants to split vs. document

_Resolves `status:needs-design`. This is the decision record for #560; implementation follows as separate PRs._

### Methodology

The value of the `ConflictError` split (#559) was **not** renaming `String`→typed — it was giving callers a small set of fixed, distinguishable cases to `match` on (incl. unit variants like `TargetExists`). So the test for each remaining variant is **not** "how many call sites" but:

> Do the construction messages cluster into a **small set (2–6) of fixed, distinguishable failure families** a consumer could meaningfully branch on? And does the variant **earn its keep** at all?

Evidence gathered: every non-test construction site (verbatim messages) + every consumer match-site across the whole workspace (core + cli + mcp + embed).

### Correction to the issue's a-priori table

[...]

PR #880, opened 2026-06-26
refactor(storage): str_to_fact_type/str_to_event_type return a parse error, not NotFound, for invalid variants

### LINKED ISSUE #366: [super-qa] medium: str_to_fact_type and str_to_event_type return MemoryError::NotFound for invalid enum variants

**str_to_fact_type and str_to_event_type return MemoryError::NotFound for invalid enum variants**

- **ID:** `store/design-wrong-error-variant-parse`
- **Severity:** medium
- **Area:** storage · **Type:** refactor
- **Location:** `src/store/facts.rs:29, src/store/events.rs:45`
- **Auto-fixable:** no

## Description
facts.rs:29: `Err(MemoryError::NotFound(format!("unknown fact type: {other}")))` and events.rs:45 does the same for event types. NotFound semantics are 'the requested resource does not exist in the store'. Parse failures are a completely different error class — the input string is malformed, not that a DB row is missing. This causes callers that match `MemoryError::NotFound` to conflate 'fact not in DB' with 'invalid fact_type string from DB'. Target state: add `MemoryError::InvalidData(String)` or `MemoryError::Corruption(String)` variant (since if a bad string comes from the DB column it is data corruption), or reuse `MemoryError::Internal`. Refactoring pattern: Replace Error Code with Exception / Rename. Motivation: correctness, prevents silent conflation of error semantics. Estimated scope: error.rs + facts.rs + events.rs + any match sites on NotFound. Risk: low — purely additive error variant.

## Suggested approach
// In error.rs, add:
#[error("invalid data: {0}")]
InvalidData(String),

// In facts.rs:29:
other => Err(MemoryError::InvalidData(format!("unknown fact_type in DB: {other}"))),

// Same in events.rs:45.

## References
CWE-703

dutiona:
## Summary

`str_to_fact_type` (`store/facts.rs`) and `str_to_event_type` (`store/events.rs`) run on the **read path**, deserializing the `fact_type` / `event_type` column back from the database. Both classified an unparseable value as `MemoryError::NotFound`.

That conflates two unrelated error classes. `NotFound` means *"the requested row is absent"* — a recoverable, expected condition a caller may branch on. But these strings are written by `fact_type_to_str` / `event_type_to_str`, so a value that does **not** parse cannot be a missing row: it is **data-integrity corruption** (the column was tampered with, or a write/read invariant broke). Mapping it to `NotFound` lets a caller matching `NotFound` silently treat a corrupt store as "no such fact".

## Change (per issue)

- **#366** — Map both parse seams to `MemoryError::Internal` instead of `NotFound`.
  - `store/facts.rs::str_to_fact_type` → `Internal("corrupt stored fact_type: {e}")`
  - `store/events.rs::str_to_event_type` → `Internal("corrupt stored event_type: {other}")`

### Why `Internal`, not a new `InvalidData`/`Corruption` variant

Per the **#560** typed-error arc, `Internal` is the documented home for *"a bug or corrupted state the caller cannot meaningfully recover from"* — terminal, propagate-don't-`match`. A corrupt stored enum has **no per-value recovery** (no caller can act differently per offending string), so #560's "one variant per *recoverable* cause" rule argues **against** adding a dedicated `InvalidData`/`Corruption` variant here. `error.rs` is therefore intentionally untouched; this is a pure read-path semantic fix scoped to the two store modules. The message stays greppable, embedding the offending token verbatim.

The legitimate row-absent `NotFound` (e.g. `EventStore::get` / `FactStore::get` on a missing id) is unchanged — only the parse seam moves. No existing test asserted `NotFound` from these parse paths (the one `NotFound` test, `events::get_not_found`, is the genuine missing-row path and stays).

## Test plan (TDD)

Added one focused unit test per seam (red→green): asserts `Internal` (not `NotFound`) and that the message contains the offending token.

| Gate | Result |
| --- | --- |
| `cargo fmt --all` | clean |
| `cargo build --workspace` (default features) | exit 0 |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | No issues found |
| `cargo test -p memory-engine --all-features` | 1378 passed, 71 ignored |

Fixes #366

### DIFF
--- a/memory-engine/lib/memory-engine/src/store/events.rs
+++ b/memory-engine/lib/memory-engine/src/store/events.rs
@@ -36,6 +36,18 @@ pub const fn event_type_to_str(et: &EventType) -> &'static str {
 /// #560), **not** [`MemoryError::NotFound`] — `NotFound` means "the requested row
 /// is absent", a semantically distinct, recoverable condition (#366). The message
 /// embeds the offending token verbatim for diagnostics.
+///
+/// What a *read-path* caller sees, end-to-end: the read site ([`row_to_event`])
+/// boxes this `Internal` into `rusqlite::Error::FromSqlConversionFailure`, which
+/// re-maps to [`MemoryError::Database`] via its `#[from]`. So the surfaced
+/// top-level variant is `Database` (a *data* error — the correct class for corrupt
+/// stored data, and crucially **not** `NotFound`, the conflation #366 reported),
+/// with this `Internal` preserved as the boxed source. The `Internal` value here
+/// is therefore the directly-observable error only for a *direct* caller; the read
+/// path re-classifies it to `Database`. Both surfaced behaviors are covered by
+/// tests (`corrupt_event_type_read_path_surfaces_database_not_notfound` for the
+/// end-to-end path, `str_to_event_type_rejects_unknown_as_internal` for the direct
+/// call).
 fn str_to_event_type(s: &str) -> Result<EventType> {
     match s {
         "Interaction" => Ok(EventType::Interaction),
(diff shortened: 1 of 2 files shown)


dutiona:
**Review thread resolution (codex P2 ×2 — corrupt fact/event type error propagation):**

Investigated end-to-end and confirmed by an agy/Gemini second-model review. You're right that the helper's `MemoryError::Internal` is boxed by `row_to_*` into `rusqlite::Error::FromSqlConversionFailure` and never surfaces as a top-level variant. Resolution and rationale:

- The user-visible read-path error is **`MemoryError::Database`**, carrying `internal error: corrupt stored fact_type: …` as the boxed source. That is the **correct class** for corrupt stored data and matches existing precedent (malformed timestamps, invalid embedding blobs also box to `Database`).
- This fully eliminates #366's actual harm: the old `not found: unknown …` semantic is **gone from the chain**, so no caller can mistake a corrupt enum for an absent row.
- Added **end-to-end tests** (insert → corrupt the enum column via the raw-SQL test seam → `store.get()` → assert the surfaced `Database` error carries the corrupt-stored message). The private helper retains `Internal` for any future *direct* caller.
- Forcing `Internal` to the top level would require unboxing `FromSqlConversionFailure`, mis-classifying disk corruption as a logic bug — declined (agy concurred this would be worse).

Resolving both threads.

PR #884, opened 2026-06-29
test(storage): cover dump_sqlite live-database + null-byte path guards

### LINKED ISSUE #315: [super-qa] high: dump_sqlite live-database safety guard has no test

**dump_sqlite live-database safety guard has no test**

- **ID:** `inspect/testing-dump-live-db-guard`
- **Severity:** high
- **Location:** `src/inspect/dump.rs:196-206`
- **Auto-fixable:** no

## Description
`dump_sqlite` (dump.rs:196-206) canonicalizes both the source DB path and the target path and returns `MemoryError::Conflict` if they resolve to the same file. This is a data-safety mechanism that prevents overwriting the live database. No test verifies it fires. A regression could silently corrupt the live database on a re-dump.

## Suggested approach
Add `sqlite_dump_refuses_to_overwrite_live_db`: open a file-backed engine at path `source.db`, then call `engine.dump_state(&DumpFormat::Sqlite(source_path.clone()))` and assert `matches!(err, MemoryError::Conflict(_))`.

## References
src/inspect/dump.rs:196, src/inspect/dump.rs:201

### LINKED ISSUE #463: [super-qa] medium: dump_sqlite null-byte path guard untested

**dump_sqlite null-byte path guard untested**

- **ID:** `inspect/testing-dump-null-byte-guard`
- **Severity:** medium
- **Location:** `src/inspect/dump.rs:213-218`
- **Auto-fixable:** no

## Description
`dump_sqlite` rejects paths containing null bytes (dump.rs:213-218) with `MemoryError::Internal`. This guard prevents SQL injection via `VACUUM INTO '...'`. No test verifies the rejection. The null-byte check is only reached for file-backed engines (the `!is_file_backed` short-circuit at line 194 means in-memory engines never hit it), so an integration test with a file-backed engine is needed.

## Suggested approach
Add `sqlite_dump_rejects_null_byte_path`: open a file-backed engine, construct a `PathBuf` from bytes containing `\0` (via `std::path::PathBuf::from(OsString::from_vec(b"output\0.db".to_vec()))` on Unix), call `dump_sqlite`, and assert `matches!(err, MemoryError::Internal(_))`.

## References
src/inspect/dump.rs:213

dutiona:
Adds regression coverage for the two untested `dump_sqlite` data-safety guards in `inspect/dump.rs`, from epic #258 (Wave 2). Test-only PR — no production code touched.

### #315 — live-database safety guard
`dump_sqlite` canonicalizes the source DB path and the dump target and refuses (`ConflictError::DumpTargetIsLiveDatabase`) when they resolve to the same file, preventing a re-dump from clobbering the live database. The guard is gated on `is_file_backed`, so an in-memory connection never reaches it.

- **`dump_sqlite_refuses_to_overwrite_live_db`** — opens a *file-backed* connection at `source.db`, dumps onto that same path, asserts the exact `DumpTargetIsLiveDatabase` variant, and verifies the source DB stays a valid SQLite file (was not partially overwritten).
- **`dump_sqlite_refuses_symlink_resolving_to_live_db`** (unix) — corollary proving `canonicalize` follows a symlink, so the guard catches a symlink whose referent is the live DB (not bypassable by indirection; a guard comparing raw paths would miss it).

### #463 — null-byte path guard
A dump target with an interior NUL byte must be rejected before reaching `VACUUM INTO`.

- **`dump_sqlite_rejects_null_byte_path`** (unix) — feeds a `<tempdir>/out\0.db` target (interior NUL, valid parent dir) and asserts the rejection, plus a NUL-free control dump to attribute the failure to the NUL specifically.

**Variant deviation (current main):** the original finding expected `MemoryError::Internal`. After the #836 refactor (`VACUUM INTO` into a server-minted temp), the surviving in-function NUL check guards the *server-minted* temp path, which can never carry a caller NUL. A caller-path NUL is now caught earlier by the unconditional `std::fs::symlink_metadata(path)` probe and typed as **`MemoryError::Io` (`ErrorKind::InvalidInput`)**. The test asserts the live behavior and documents why.

## Test plan
All five CI-matrix gates run from the worktree, all green:

1. `cargo fmt --all` — clean
2. `cargo build --workspace` — Finished, no errors
3. `cargo build --workspace --all-features` — Finished, no errors
4. `cargo clippy --workspace --all-targets --all-features -- -D warnings` — exit 0
5. `cargo test -p memory-engine --all-features` — ok, 1331 passed / 0 failed (+ all sub-suites green)

Each key assertion was **mutation-checked**: asserting the wrong error variant, or that the guard returns `Ok`, makes the corresponding test fail — confirming non-vacuity.

Fixes #315, Fixes #463

### DIFF
--- a/memory-engine/lib/memory-engine/src/inspect/dump.rs
+++ b/memory-engine/lib/memory-engine/src/inspect/dump.rs
@@ -990,9 +990,6 @@ mod tests {
     #[tokio::test]
     async fn dump_sqlite_rejects_null_byte_path() {
         use crate::store::schema::{init_schema, open_connection};
-        use std::ffi::OsString;
-        use std::os::unix::ffi::OsStringExt;
-        use std::path::PathBuf;
 
         let dir = tempfile::tempdir().unwrap();
         let source = dir.path().join("source.db");
@@ -1001,11 +998,10 @@ mod tests {
 
         // Build a target path *inside* the temp dir with an interior NUL byte in
         // the leaf name, so only the NUL — not a missing parent dir — can be the
-        // cause of any rejection. Assemble the bytes directly (`<dir>/out\0.db`)
-        // so the NUL is preserved verbatim through to `dump_sqlite`.
-        let mut raw: Vec<u8> = dir.path().as_os_str().to_os_string().into_vec();
-        raw.extend_from_slice(b"/out\0.db");
-        let bad_path = PathBuf::from(OsString::from_vec(raw));
+        // cause of any rejection. A NUL is a valid `char`, so the literal leaf
+        // `"out\0.db"` carries the byte verbatim through `join` into
+        // `dump_sqlite`.
+        let bad_path = dir.path().join("out\0.db");
         assert!(
             bad_path.as_os_str().as_encoded_bytes().contains(&0),
             "the constructed bad path must actually contain a NUL byte"
@@ -1027,6 +1023,12 @@ mod tests {
         let good_path = dir.path().join("out.db");
         dump_sqlite(&conn, &good_path)
             .expect("an otherwise-identical path without a NUL byte must dump cleanly");
+
+        // Drop the live source connection before verifying the dump: the
+        // verification reopen below must observe the published file without
+        // racing the still-open writer handle on the same temp dir (avoids any
+        // SQLite lock-contention / sidecar-file flakiness).
+        drop(conn);
         rusqlite::Connection::open_with_flags(
             &good_path,
             rusqlite::OpenFlags::SQLITE_OPEN_READ_ONLY,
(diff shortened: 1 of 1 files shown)


dutiona:
Addressed all three: (1) **drop-conn** — `dump_sqlite_rejects_null_byte_path` now `drop(conn)`s the live source connection before the read-only verification reopen, removing the lock-contention flake risk. (2) **simpler NUL path** — replaced the manual `Vec<u8>` assembly with `dir.path().join("out\0.db")` (NUL survives `join`; still rejected as Io/InvalidInput), dropping the unused OsString imports. (3) **contract** — already satisfied by the earlier docstring fixup: the test asserts the documented contract (`Io/InvalidInput` before the NUL reaches VACUUM INTO, framed as defense-in-depth), not an incidental layer. Resolving.

Issue #255, opened 2026-06-01
[super-qa] Area: forgetting — code-quality findings


dutiona:
All `/super-qa` 2026-06-01 findings in **area:forgetting**. Non-autofixable blocker/critical/high/medium are individual sub-issues; low/info are grouped in one sub-issue.

dutiona:
## Epic #255 — area:forgetting super-qa sweep: COMPLETE ✅

All 9 sub-issues resolved across **3 waves + 1 tail wave — 6 PRs merged**. Triaged against *current* main, not the 2026-06-01 snapshot: findings predated the #631 async cutover (prune's write txn moved below the StorageBackend seam into `prune_atomic`) and the #814 bin/lib reorg, so 6/9 cited relocated paths.

### Disposition — 9 sub-issues
| # | type/sev | Outcome |
|---|----------|---------|
| #351 | refactor/med | **PR #888** — relocated `ForgetPolicy`/`PruneStats` traits.rs → forgetting/types.rs (crate-root re-export preserved) |
| #380 | docs/med | **PR #888** — corrected the drifted importance-formula doc (folded into #351's move) |
| #392 | refactor/med | **PR #885** — N+1 importance UPDATE → O(N) `UPDATE...FROM json_each` + non-finite guard |
| #393 | refactor/med | **already-fixed** (7cf493a: `SCORING_COLUMNS`/`list_active_facts_scoring`) — closed with evidence, no PR |
| #312 | test/**high** | **PR #902** — cascade-on-prune + graph-reconcile + degree-weight tests; mutation-verified. Canary clean: coverage gap, **not** a latent bug |
| #379 | docs/med | **PR #905** — `ebbinghaus_decay`/`compute_importance` half_life precondition docs |
| #454 | test/med | **PR #905** — empty-store prune no-op test |
| #455 | test/med | **PR #905** — decay/importance property-based tests |
| #497 | chore/low-info | **PRs #905 + #907** (6 residue findings), 2 already-fixed, 1 carved (#908) — see #497 close-out |

### PRs merged
#888, #885, #902, #905, #907, #916

### Collateral filed
- **#895** — dedup.rs N+1 importance UPDATE (area:consolidation; can reuse #392's bulk method)
- **#903** — orphaned eval harness → **PR #916**: 56 conformance+quality tests had been silently un-run since #814; re-attached, restoring coverage repo-wide (also benefits the concurrent #258/#259 sweeps)
[...]

PR #885, opened 2026-06-29
refactor(forgetting): batch importance_score materialization into one UPDATE

### LINKED ISSUE #392: [super-qa] medium: N+1 individual UPDATE statements for importance_score materialization

**N+1 individual UPDATE statements for importance_score materialization**

- **ID:** `forgetting/perf-n-plus-one-score-update`
- **Severity:** medium
- **Area:** forgetting · **Type:** refactor
- **Location:** `src/forgetting/policy.rs:114-118`
- **Auto-fixable:** no

## Description
`update_importance_score` (store/facts.rs:564-570) executes one `UPDATE facts SET importance_score = ?1 WHERE id = ?2` per active fact inside the transaction. For a store with 100k facts this issues 100k individual prepared-statement executions. SQLite batches writes within a transaction efficiently, but each call still incurs statement preparation overhead and a B-tree write per row. A single bulk UPDATE via `json_each` would collapse this to one statement regardless of corpus size.

## Suggested approach
Add a `update_importance_scores_bulk(scores: &[(i64, f64)])` method to `FactStore` using `json_each`:
```sql
UPDATE facts
SET importance_score = (
    SELECT value
    FROM json_each(?1)
    WHERE key = CAST(facts.id AS TEXT)
)
WHERE id IN (SELECT key FROM json_each(?1))
```
Or use a temporary table approach for very large datasets. Call this once from `prune` with the full scored vec instead of looping `update_importance_score`.

## References
—

dutiona:
## What

The prune path materialized `importance_score` with an N+1 loop of single-row UPDATEs inside `prune_atomic` (`storage/sqlite/graph.rs`): one prepared-statement exec + one B-tree write per active fact. For a 100k-fact store that is 100k statement executions per prune.

This replaces the loop with a single `json_each`-driven bulk UPDATE — one statement regardless of corpus size:

```sql
UPDATE facts
SET importance_score = (SELECT value FROM json_each(?1) WHERE key = CAST(facts.id AS TEXT))
WHERE id IN (SELECT CAST(key AS INTEGER) FROM json_each(?1))
```

## Why it's correct

Behavior-identical to the loop, guarded by a new TDD test (`update_importance_scores_bulk_sets_subset_and_round_trips_ids`):

- (a) sets exactly the scored subset;
- (b) an unscored row is left **untouched** — the `WHERE id IN (...)` guard; without it the correlated subquery would NULL every unmatched row;
- (c) i64 ids round-trip the `json_each` TEXT key cast (a large id, 2^53+1, catches a TEXT/INTEGER mismatch — the issue body's suggested SQL omitted `CAST(key AS INTEGER)`);
- (d) an empty slice is a no-op returning `Ok(())` (early return, no degenerate statement).

The per-row `update_importance_score` is **retained** — still used by the single-fact callers (`graph.rs:219`, `consolidation/dedup.rs:260`).

## Gate

`cargo fmt --check`, `cargo build --workspace`, `cargo test --workspace` (1702 passed), `cargo clippy --workspace --all-targets --all-features -- -D warnings` — all green.

Closes #392
Part of #255

### DIFF
--- a/memory-engine/lib/memory-engine/src/storage/sqlite/graph.rs
+++ b/memory-engine/lib/memory-engine/src/storage/sqlite/graph.rs
@@ -990,10 +990,9 @@ impl FactGraph for SqliteBackend {
                 let fact_store = FactStore::new(&tx, dim);
                 let edge_store = EdgeStore::new(&tx);
 
-                // Materialize importance scores for every active fact.
-                for &(id, score) in &scored {
-                    fact_store.update_importance_score(id, score)?;
-                }
+                // Materialize importance scores for every active fact in one
+                // bulk UPDATE (#392) instead of an N+1 per-row loop.
+                fact_store.update_importance_scores_bulk(&scored)?;
                 // Expire the sub-threshold set + cascade edge expiry.
                 for &fact_id in &to_expire {
                     fact_store.expire(fact_id, now)?;
--- a/memory-engine/lib/memory-engine/src/store/facts.rs
+++ b/memory-engine/lib/memory-engine/src/store/facts.rs
@@ -871,15 +871,19 @@ impl<'a> FactStore<'a> {
     /// `update_importance_score` calls (one prepared-statement exec + one
     /// B-tree write each) into one `json_each`-driven UPDATE, regardless of
     /// corpus size. Behaviorally identical to the loop: only the ids present in
-    /// `scores` are touched — rows outside the payload are left untouched. The
-    /// `WHERE id IN (...)` guard restricts the UPDATE to the scored ids; without
-    /// it the correlated subquery yields NULL for every unmatched row, which the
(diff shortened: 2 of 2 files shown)


dutiona:
**Review thread dispositions (finish-pr):**

Both bots flagged the original correlated-subquery O(N²) concern — **now fixed**, so these threads are resolved against the current code:

- The SET clause is no longer a correlated subquery. The statement is now an `UPDATE facts SET importance_score = s.value FROM json_each(?1) AS s WHERE facts.id = CAST(s.key AS INTEGER)` join (facts.rs:933): the JSON payload is parsed exactly once and index-joined by id — O(N), exactly as both reviewers recommended.
- gemini's comment at facts.rs:936 re-attached the pre-fix finding to the new SHA; it does not apply to the FROM-join code.

Verified: both `update_importance_scores_bulk_*` tests pass and all 7 CI checks are green on 609a51e.

PR #889, opened 2026-06-29
test(storage): pak read_pak fuzz target + UpcasterRegistry/parse_timestamp proptests + parse_timestamp fuzz

### LINKED ISSUE #421: [super-qa] medium: read_pak has no fuzz target — cold-storage bytes from disk are parsed without fuzz coverage

**read_pak has no fuzz target — cold-storage bytes from disk are parsed without fuzz coverage**

- **ID:** `archive/no-fuzz-target-read-pak`
- **Severity:** medium
- **Location:** `src/archive/pak.rs:58-66 (read_pak), no fuzz/ directory exists`
- **Auto-fixable:** no

## Description
read_pak (pak.rs:58) decompresses a zstd stream and then deserializes JSON from it. It is the only function in the codebase that accepts bytes originating from disk storage (i.e., potentially tampered or corrupted .pak files). There is no fuzz/ directory in the workspace. CWE-409 (decompression from untrusted source) and CWE-20 (improper input validation in deserialization) are both relevant. The MAX_PAK_DECOMPRESSED_SIZE cap at pak.rs:64 is the main safety valve, and it is not covered by any test. Fuzzing read_pak is the highest-value fuzz candidate in the archive module because it combines two parsing layers (zstd + serde_json) with a size limit that must hold.

## Suggested approach
Add a fuzz target with cargo-fuzz:
```
# fuzz/fuzz_targets/fuzz_read_pak.rs
#![no_main]
use libfuzzer_sys::fuzz_target;
use memory_engine::archive::pak::read_pak;
use std::io::Write;
fuzz_target!(|data: &[u8]| {
    let dir = tempfile::tempdir().unwrap();
    let p = dir.path().join("fuzz.pak");
    std::fs::write(&p, data).unwrap();
    let _ = read_pak(&p); // must not panic regardless of input
});
```
Add `[package] name = "fuzz" ...` and add memory-engine as path dependency in fuzz/Cargo.toml. Run with `cargo fuzz run fuzz_read_pak`.

## References
src/archive/pak.rs:9, src/archive/pak.rs:58-66

### LINKED ISSUE #486: [super-qa] medium: UpcasterRegistry chain composition lacks property-based tests

**UpcasterRegistry chain composition lacks property-based tests**

- **ID:** `store/proptest-upcaster-chain`
- **Severity:** medium
- **Location:** `src/store/upcaster.rs`
- **Auto-fixable:** no

## Description
The upcaster has example-based tests for chains of length 1 and 2, but no proptest that verifies the fundamental composition property: upcast(payload, from_rev, n) applied iteratively for arbitrary chain lengths equals a direct upcast from from_rev to latest. Properties to verify: (1) upcast is idempotent when from_rev == current_rev; (2) chained single-step upcasts produce the same result as a direct multi-step upcast; (3) attempting to upcast from a revision higher than the latest returns an error. A mutant upcaster that skips one step in the chain could pass all existing tests.

## Suggested approach
Add a proptest module in upcaster.rs that generates arbitrary revision chains (0..=10), registers incrementing upcasters, and verifies the composition invariant. Use proptest::collection::vec to generate chain lengths.

## References
—

### LINKED ISSUE #487: [super-qa] medium: parse_timestamp and parse_optional_timestamp have no property-based roundtrip tests

**parse_timestamp and parse_optional_timestamp have no property-based roundtrip tests**

- **ID:** `store/proptest-timestamp-parse`
- **Severity:** medium
- **Location:** `src/store/mod.rs:63,78`
- **Auto-fixable:** no

## Description
mod.rs has proptest coverage for embedding serialization roundtrips, but not for the two timestamp parse helpers. parse_timestamp and parse_optional_timestamp are called in every row-mapping function across all store modules — a regression here would silently corrupt all datetime fields. The property to verify: for any DateTime<Utc> value d, parse_timestamp(d.to_rfc3339()) == d. parse_optional_timestamp on a non-RFC3339 string should return a rusqlite::Error.

## Suggested approach
Add a proptest block in mod.rs alongside the existing proptest_embedding module, generating arbitrary Unix timestamps and verifying the serialize/parse roundtrip. Also add a negative test with arbitrary non-RFC3339 strings (e.g., via proptest::string::string_regex for invalid formats) asserting parse returns Err.

## References
—

### LINKED ISSUE #488: [super-qa] medium: No fuzz targets exist — parse_timestamp with arbitrary byte strings is highest priority

**No fuzz targets exist — parse_timestamp with arbitrary byte strings is highest priority**

- **ID:** `store/fuzz-timestamp-parse`
- **Severity:** medium
- **Location:** `src/store/mod.rs:63,78`
- **Auto-fixable:** no

## Description
There is no fuzz/ directory in the workspace. parse_timestamp and parse_optional_timestamp both take &str from SQLite column values. While SQLite controls input in production, a corrupted DB or a migration path that stores a malformed timestamp string would cause rusqlite::Error at read time rather than a panic — but the error conversion path (Box::new(e) into FromSqlConversionFailure) has not been stress-tested. A fuzz target is particularly valuable because DateTime::parse_from_rfc3339 from the chrono crate has historically had edge cases around leap seconds, timezone designators, and sub-second precision that break parsing assumptions.

## Suggested approach
Create a fuzz/ directory with cargo-fuzz. Add a fuzz target fuzz_parse_timestamp that takes arbitrary &[u8], converts to str if valid UTF-8, and calls parse_timestamp — asserting it never panics (only returns Ok or Err). Enable with cargo +nightly fuzz run fuzz_parse_timestamp.

## References
—

dutiona:
## Summary

Storage super-qa Wave 2 (epic #258) — four test-coverage gaps in the persistence layer, all property/fuzz-based.

### #421 — `read_pak` had no fuzz target
- `fuzz/fuzz_targets/read_pak.rs` (+ `[[bin]]`) drives the two-layer **zstd-frame + `serde_json::<ArchivePak>`** parser, including its 4 GiB decompression-bomb cap, over fuzzer bytes. Asserts the total contract: every input is `Ok`/`Err`, never a panic. Distinct from `snapshot_restore` (single magic-sniff JSON path).
- **Seam needed (deviation from brief):** `read_pak` is `pub fn` inside the `#[cfg(feature = "archive")] pub(crate) mod archive`, so it is **unreachable** from the workspace-detached fuzz crate. The brief's "read_pak is pub so no fuzz_seam needed" was wrong (verified against live `lib.rs:121`). Added an `archive`-gated `fuzz_seam` re-export, mirroring how `snapshot_restore` reaches `read_snapshot`.

### #486 — `UpcasterRegistry` chain composition lacked property tests
`proptest_chain` in `store/upcaster.rs` drives chains of **0..=10** registered steps and pins three laws:
1. **idempotent at latest** — upcasting an already-latest payload is a byte-for-byte no-op;
2. **chained == direct** — one `upcast` over the whole span equals folding the step `(latest - start)` times by hand (an order/count-sensitive `trace` array catches off-by-one and reordering, not just the final revision);
3. **future revision is a no-op** — a revision `> latest` returns unchanged with the caller's own revision preserved (not clamped).

### #487 — timestamp parsers had no property tests
`proptest_timestamps` in `store/mod.rs`:
- arbitrary whole-second UTC instants survive `to_rfc3339` -> `parse_timestamp` **byte-for-byte**;
- arbitrary non-RFC3339 strings -> `Err(FromSqlConversionFailure)` (both `parse_timestamp` and `parse_optional_timestamp`);
- `parse_optional_timestamp(None)` -> `Ok(None)`. The three optional-parser arms use distinct expected values so a predicate flip is caught.

### #488 — no fuzz target over raw bytes for `parse_timestamp`
- `fuzz_seam` re-export of both timestamp helpers + `fuzz/fuzz_targets/parse_timestamp.rs` (+ `[[bin]]`) asserting the never-panic `Ok`/`Err` contract over arbitrary UTF-8 (reaches UTF-8 boundaries / control chars a `String` strategy seldom produces).

## Test quality (non-vacuous)
Every new assertion was mutation-checked to confirm it fails when the behavior is wrong:
- #487 roundtrip: a `+1s` shift in `parse_timestamp` -> FAIL; a swallow-error (`Ok(epoch)`) -> negative props FAIL.
- #486: loop-bound `<` -> `<=` -> `chained_equals_direct` FAIL; early-return rev-clamp -> `future_revision_is_noop` FAIL.

## Test plan (all gates green)
| Gate | Result |
| --- | --- |
| `cargo fmt --all` | clean |
| `cargo build --workspace` | 0 errors |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | exit 0 (true exit confirmed via `rtk proxy`) |
| `cargo test -p memory-engine --all-features` | 1336 + integration suites pass, 0 failed |
| `cd fuzz && RUSTFLAGS="--cfg fuzzing" cargo check` | exit 0 (detached fuzz crate; both new targets type-check) |

Fixes #421, Fixes #486, Fixes #487, Fixes #488


dutiona:
Valid finding, fixed. The proptest was order-blind: every step was the same `trace_step` whose marker derived only from trace LENGTH, so a misordered lookup with the right count produced an identical trace. Fix: per-revision distinct step functions (a `rev_steps!` macro → `STEP_FNS`) that each append their OWN `from_revision`; `chained_equals_direct` now asserts the exact order-sensitive `start..latest` sequence. Mutation-verified: reversing the lookup order in `upcast` now FAILS (`[2,1]` vs `[1,2]`) while the length stayed equal. Tests-only. Resolving.

PR #713, opened 2026-06-21
feat(storage): move HNSW ownership into SqliteBackend (#631 Stage B)

dutiona:
## Summary

**Stage B of #631** (epic #628) — additive prep, **engine byte-identical**. Moves HNSW vector-search ownership into `SqliteBackend` (completes #630's D3 deferral) so its `vector_search` keeps HNSW acceleration. The engine keeps its own `hnsw_strategy` until the Stage E cutover; this is a parallel, isolation-tested backend index.

## What

- **Struct + ctor:** `SqliteBackend` gains `search_config: Option<SearchConfig>` and `#[cfg(feature="ann")] hnsw: Option<Arc<HnswStrategy>>`; a `with_search_config` builder builds the index via `HnswStrategy::build_from_db` under `ann` + `ann_threshold < usize::MAX` (`from_pool` unchanged → brute-force, exactly #630).
- **Dispatch:** `vector_search` replicates the engine's `should_use_hnsw` predicate exactly (`active_count() >= ann_threshold`), plus a filter-compat gate — HNSW only for `Active` + `fact_type`/`scope_ids` (the dims `HnswStrategy::check_fact_filters` honors); richer `FactFilter` dims fall through to the brute path, which (post-#684) honors them via `build_filter_sql`. `f32→f64` widening at the seam boundary.
- **Maintenance:** `notify_insert`/`notify_expire` wired into the backend's owned write methods (insert / reinforce / atomic-insert / batch / expire / delete), post-commit. `notify_insert` is idempotent (tombstones the prior entry), so reinforce updates rather than duplicates. Cycle-apply returns `to_index` for the caller (Stage A contract).
- **Snapshot:** serialize/deserialize relocated to backend methods for Stage E to wire the engine snapshot/Drop.

## Verification

- `cargo test --all-features` — **1239 passed**; `cargo test --workspace` — **1533 passed**
- `cargo clippy --workspace --all-targets --all-features` + `--features async` (no-ann) — clean; `cargo build` (default) compiles
- `src/engine/` **byte-identical**; rebased onto current `main`.

## Fidelity

Self-conducted review (committed at `docs/plans/2026-06-21-stage-B-699-hnsw-fidelity-review.md`; the delegated reviewer hit the monthly subagent spend limit) verified: dispatch-predicate parity, the filter-compat gate composing correctly with #684's full-filter brute path, idempotent maintenance (reinforce-safe), `search_config==None`→never-HNSW, and engine byte-identity.

## Testing

[...]

PR #896, opened 2026-06-29
test(storage): integration test exercising --all-features store interactions

### LINKED ISSUE #489: [super-qa] medium: No integration test exercises --all-features store interactions

**No integration test exercises --all-features store interactions**

- **ID:** `store/integration-feature-flag-combos`
- **Severity:** medium
- **Location:** `tests/roundtrip_test.rs`
- **Auto-fixable:** no

## Description
The integration tests in tests/roundtrip_test.rs and tests/read_only_test.rs run without feature flags. The archive feature gates ArchiveManifestStore; the ann feature gates HNSW vector search. No integration test verifies that (1) a DB opened with the archive feature behaves correctly when archive_manifest records coexist with live facts, or (2) the ann feature's HNSW index is correctly maintained across ingest/expire cycles. The workspace-level cargo test --all-features is in the CLAUDE.md gate but no test file specifically targets these combinations.

## Suggested approach
Add a tests/feature_flags_test.rs that is conditionally compiled per feature (using #[cfg(feature = "archive")] and #[cfg(feature = "ann")]) and exercises at least one round-trip per gated code path.

## References
—

dutiona:
## Summary

Adds `tests/all_features_store.rs`, a new integration test gated on `all(feature = "ann", feature = "archive", feature = "compress-zstd")` that drives a single cross-feature store flow through the `MemoryEngine` facade — the feature combination a `cargo test --all-features` build links but that no existing test exercised *together*.

Resolves #489 (epic #258, Wave 2). Previously each feature was covered only in isolation: `ann_recall_test.rs` (vector recall, no expire), `archive_test.rs` (archive+live coexistence), and the per-format roundtrips in `restore_roundtrip_test.rs`.

## Per-issue coverage (#489)

#489 flagged two uncovered `--all-features` interactions:

1. **archive ↔ live coexistence** — *already covered on current main* by `tests/archive_test.rs` (`#![cfg(feature = "archive")]`): forget → archive → `list_archives` / `verify_archives` → archive-search vs live-search → pinned-skip. This half was closed by a Wave-1 merge; no new work needed (the brief confirms PARTIALLY_FIXED).

2. **ann HNSW maintenance across ingest/expire cycles** — *this was the gap*. The HNSW index's behaviour across an ingest → expire → archive cycle (tombstoning via `notify_expire` + the `t_expired IS NULL` post-filter; the rebuild-from-DB path on restore) was unit-tested only inside `src/search/ann.rs`, never through the facade. This PR covers it end-to-end, and folds in the archive + compress legs so the *combination* is exercised in one flow.

### The flow exercised

- Ingest one Episodic fact per topic; HNSW vector search resolves each topic to its own fact (pre-expire).
- Pin even topics, `forget` aggressively → exactly the odd (unpinned, non-decay-exempt) topics expire, driving `expire_fact → notify_expire` (tombstone) per fact (#713).
- Assert HNSW drops the tombstoned facts but still surfaces survivors; live-fact accounting matches the survivor set.
- Archive the expired facts; verify `pak_path`/size/hash, manifest `fact_count`, and `verify_archives` integrity.
- Same text query: live search finds nothing for an expired topic, archive search finds it tagged `MatchType::Archive` with `archive_paks_scanned > 0`.
- `zstd`-dump the survivors, restore into a fresh in-memory engine whose HNSW index is rebuilt from the restored rows; assert it resolves a survivor and does not contain an expired topic.

## Test quality (non-vacuity)

- A one-hot-per-topic embedder makes "nearest neighbour to topic K is fact K" a sharp, mutation-sensitive assertion (orthogonal topic vectors).
- The survivor/expired split is asymmetric (even survive / odd expire) so a predicate flip or tuple swap fails rather than passing vacuously.
- **Mutation-verified**: pinning the wrong parity makes the post-expire HNSW assertion fail as expected (`expired topic 1 must NOT surface` → `left == right` panic), proving the assertions bind to real index behaviour, not a tautology.

## Test plan — full CI matrix, all green

| Gate | Result |
| --- | --- |
| `cargo fmt --all` | clean |
| `cargo build --workspace` | 0 errors (2 pre-existing unrelated `dead_code` warnings) |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | exit 0, no issues |
| `cargo test -p memory-engine --all-features` | exit 0; `ann_archive_compress_end_to_end ... ok`; 0 failed across all suites |

## Notes

- `memory-engine/lib/memory-engine/Cargo.toml` has `autotests = false` and every integration test is registered with an explicit out-of-tree `path`, so a one-line `[[test]]` registration block for the new file is mechanically required (appended after `roundtrip_test`).

Fixes #489

### DIFF
--- a/memory-engine/lib/memory-engine/Cargo.toml
+++ b/memory-engine/lib/memory-engine/Cargo.toml
@@ -127,6 +127,10 @@ path = "../../../tests/restore_roundtrip_test.rs"
 name = "roundtrip_test"
 path = "../../../tests/roundtrip_test.rs"
 
+[[test]]
+name = "all_features_store"
+path = "../../../tests/all_features_store.rs"
+
 # --- examples (top-level ../../../examples/) ---
 [[example]]
 name = "basic_roundtrip"
--- a/tests/all_features_store.rs
+++ b/tests/all_features_store.rs
@@ -34,7 +34,8 @@ use memory_engine::search::SearchConfig;
 use memory_engine::traits::{EmbeddingProvider, ForgetPolicy};
 use memory_engine::types::{AddFactRequest, FactType};
 use memory_engine::{
-    ArchivePolicy, EmbeddingFingerprint, MatchType, MemoryQuery, SearchMode, SearchQuery,
+    ArchivePolicy, EmbeddingFingerprint, EngineConfig, MatchType, MemoryQuery, SearchMode,
+    SearchQuery,
 };
 
 /// Embedding width — one slot per topic plus headroom, so each topic maps to a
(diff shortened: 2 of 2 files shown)


dutiona:
**Review thread resolution:**
- *Restore with ANN before asserting HNSW rebuild (codex P2):* fixed. `restore_json_memory` calls `init_from_pool(.., None, ..)`, dropping the ANN `SearchConfig` so HNSW was never materialized and the assertions passed via brute-force. Now restores via `restore_json` carrying `SearchConfig { ann_threshold: 0 }` so the index is rebuilt from the restored rows and vector search is HNSW-served — the assertions bite on the rebuild (a regressed restore-time rebuild now fails). Also loops all survivor topics.
- *`is_multiple_of` not available (gemini):* respectfully declined — false positive. `u*::is_multiple_of` is stable since Rust 1.87; MSRV is 1.88 and the Build + MSRV CI gates are green on this commit, so it compiles. No change.

PR #902, opened 2026-06-29
test(forgetting): cover edge cascade + graph reconcile + degree-weight on prune

### LINKED ISSUE #312: [super-qa] high: Graph edge cascade on prune is never tested

**Graph edge cascade on prune is never tested**

- **ID:** `forgetting/testing-missing-graph-edge-cascade`
- **Severity:** high
- **Area:** forgetting · **Type:** test
- **Location:** `src/forgetting/policy.rs:120-131`
- **Auto-fixable:** no

## Description
The `prune` function performs two operations when a fact is expired: it calls `edge_store.expire_by_fact(fact_id, now)?` (SQL cascade) and `graph.remove_edges_by_fact(fact_id)` (in-memory graph). All existing inline tests (`prune_expires_low_importance_facts`, `prune_skips_pinned_facts`, `prune_materializes_importance_scores`) and all integration tests pass an empty `MemoryGraph::new()` with no edges added. The SQL edge-expiry and in-memory graph-removal branches are therefore never executed by the test suite. A regression in `EdgeStore::expire_by_fact` or `MemoryGraph::remove_edges_by_fact` would be invisible. Additionally, the `graph_degree_weight` signal (the 3rd of 4 scoring signals) is always zero in every test because `graph.degree(fact_id)` always returns 0 on an empty graph, meaning the graph-connectivity branch of `compute_importance` is never meaningfully exercised in an end-to-end scenario.

## Suggested approach
Add an inline test that pre-populates both `facts` and `edges` tables (via `EdgeStore::insert`) and a `MemoryGraph` with edges, then calls `prune` and asserts: (1) the expired fact's edges are marked expired in SQLite, (2) `graph.degree(fact_id)` is 0 after prune, (3) adjacent non-expired facts retain their edges. Also add an integration test where a well-connected fact survives while an isolated one is pruned under equal importance, exercising the graph-degree signal.

## References
—

dutiona:
Closes the #312 coverage gap. The `prune` → in-memory reconcile loop and the `graph_degree_weight` survival signal were untested **end-to-end** because every prior prune test ran against an **empty** `MemoryGraph` (degree always 0, no edges to cascade). The constituent primitives — `EdgeStore::expire_by_fact`, `MemoryGraph::remove_edges_by_fact`, `prune_atomic` atomicity — were each independently tested; only their *composition* was uncovered.

## Tests added

**Unit** — `forgetting/policy.rs::prune_cascades_edges_and_reconciles_graph`
Seeds edges in BOTH the SQLite `EdgeStore` (via the storage port) and the in-memory `MemoryGraph`, prunes a low-importance Episodic victim adjacent to a decay-exempt Semantic survivor, and asserts:
- (a) the victim's edge is **soft-expired** in SQLite (`t_expired` set, still present), while a non-victim edge stays active;
- (b) `graph.degree(victim) == 0` after prune — the in-memory reconcile loop ran;
- (c) the surviving adjacent fact retains its non-victim edge in **both** projections (cascade is scoped to the victim, not a blanket wipe).

**Integration** — `tests/eval/quality/forgetting.rs::well_connected_fact_survives_prune`
Two Episodic facts with **equal** `base_importance`, age, and access count differ only in graph connectivity: one is the hub of a co-session clique (built through the real engine API — `ingest` + `link_session_facts`), the other isolated. Under a threshold tuned strictly between their computed scores, the well-connected fact survives while the isolated one is pruned — isolating `graph_degree_weight` as the decisive signal, exercised end-to-end.

## Canary result

**Both tests pass on current behavior — no product bug found.** `prune` cascades edges, reconciles the in-memory graph, and the degree signal influences survival exactly as designed. Verified under both default features and `--all-features` (the `ann` HNSW `notify_expire` path in `prune_atomic`).

## Collateral: re-attach the orphaned eval harness

While wiring the integration test I found the entire two-tier eval harness (`tests/eval/main.rs` — 56 conformance + quality tests, including the 4 pre-existing `forgetting` tests) was **not being built or run** by `cargo test --workspace`. Cargo auto-discovered it when the repo root was a package; the bin/lib restructure left the core crate's directory without a `tests/` subdir, so the harness silently stopped compiling. Added a `[[test]] name = "eval"` target (`path = "../../../tests/eval/main.rs"`) mirroring the other explicitly-attached top-level integration tests. Without this the new #312 integration test would not run at all. The harness compiles and passes clean (56 passed, 3 `#[ignore]` Phase-5 skeletons).

## Gate (CI-exact, all green)

- `cargo fmt --all --check` ✅
- `cargo build --workspace` ✅
- `cargo test --workspace` ✅ (0 failures; eval harness 56 passed / 3 ignored)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` ✅

No production code change. No `unwrap`/`expect` in lib code (the `#[cfg(test)]` module is exempt).

Closes #312
Part of #255

### DIFF
--- a/memory-engine/lib/memory-engine/Cargo.toml
+++ b/memory-engine/lib/memory-engine/Cargo.toml
@@ -75,15 +75,6 @@ harness = false
 
 # --- integration tests (top-level ../../../tests/) — attached explicitly since this
 #     crate dir has no tests/ subdir for Cargo to auto-discover ---
-# The two-tier eval harness (#16): a multi-file binary rooted at tests/eval/main.rs.
-# Cargo auto-discovered it when the repo root was a package; after the bin/lib
-# restructure the core crate's dir has no tests/ subdir, so it must be attached
-# explicitly like every other top-level integration test below — otherwise the
-# whole conformance + quality suite silently stops building under `cargo test`.
-[[test]]
-name = "eval"
-path = "../../../tests/eval/main.rs"
-
 [[test]]
 name = "activity_stream_test"
 path = "../../../tests/activity_stream_test.rs"
--- a/memory-engine/lib/memory-engine/Cargo.toml
+++ b/memory-engine/lib/memory-engine/Cargo.toml
@@ -75,6 +75,15 @@ harness = false
 
 # --- integration tests (top-level ../../../tests/) — attached explicitly since this
 #     crate dir has no tests/ subdir for Cargo to auto-discover ---
+# The two-tier eval harness (#16): a multi-file binary rooted at tests/eval/main.rs.
+# Cargo auto-discovered it when the repo root was a package; after the bin/lib
+# restructure the core crate's dir has no tests/ subdir, so it must be attached
+# explicitly like every other top-level integration test below — otherwise the
+# whole conformance + quality suite silently stops building under `cargo test`.
+[[test]]
+name = "eval"
+path = "../../../tests/eval/main.rs"
+
 [[test]]
 name = "activity_stream_test"
 path = "../../../tests/activity_stream_test.rs"
--- a/memory-engine/lib/memory-engine/src/forgetting/policy.rs
+++ b/memory-engine/lib/memory-engine/src/forgetting/policy.rs
@@ -746,4 +746,198 @@ mod tests {
             "explicit override must win over the default exemption"
         );
     }
+
+    /// #312: the prune → graph-reconcile composition. Every prior prune test runs
+    /// against an EMPTY `MemoryGraph`, so neither the SQL edge cascade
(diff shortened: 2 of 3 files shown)


dutiona:
**Review disposition (finish-pr):** gemini medium at policy.rs:802 (`NewFact::builder` vs manual struct construction in the test fixture): **declined** — non-blocking test-fixture style. The explicit construction is clear and intentional, and the test is *mutation-verified* (deleting the edge cascade at graph.rs:1000 or gutting the reconcile loop at policy.rs:195 makes it fail), so the fixture style has no bearing on correctness. Not worth a re-gate/CI cycle on a proven test. Resolving.

Issue #497, opened 2026-06-01
[super-qa] forgetting: low/info findings (8)


dutiona:
> ⚠️ Agent analysis; not source-verified. Grouped low/info for **area:forgetting**.

Grouped low/info findings for area **forgetting** (8):

- **low** `src/forgetting/policy.rs:26-68` — `compute_importance` inherits the undocumented `half_life > 0` precondition from `ebbinghaus_decay` (`forgetting/documentation-compute-importance-precondition`)
- **low** `src/forgetting/policy.rs:97-118` — compute_importance called twice per fact — once to classify, once to materialize (`forgetting/refactoring-double-score-computation`)
- **low** `src/forgetting/policy.rs:36-68` — compute_importance output bounds never asserted (`forgetting/testing-missing-compute-importance-output-bounds`)
- **low** `src/forgetting/policy.rs:49-50` — Negative age clamping (.max(0.0)) path is untested (`forgetting/testing-missing-future-last-accessed`)
- **low** `tests/eval/quality/forgetting.rs:47-51, tests/eval/helpers.rs:223-225` — Integration eval tests rely on wall-clock time with non-injectable now (`forgetting/testing-flaky-time-in-eval-tests`)
- **low** `memory-engine-mcp/src/tools/mod.rs:1007-1030, memory-engine-mcp/tests/tools.rs:127-179` — MCP handle_forget does not test invalid half_life_overrides key or non-object type (`forgetting/testing-mcp-forget-missing-invalid-override-key`)
- **info** `src/forgetting/policy.rs:11-16` — Constant name `*_NORMALIZATION_CEILING` stores `n+1`, not `n` — doc comment is ambiguous (`forgetting/documentation-ceiling-constant-naming`)
- **info** `src/forgetting/mod.rs:7` — pub use policy::prune has no doc-test; prune docstring only describes errors (`forgetting/testing-no-doctest-on-public-reexport`)

dutiona:
Closing — all 8 low/info sub-findings resolved (epic #255 area:forgetting sweep):

1. `compute_importance` half_life precondition doc → **PR #905** (folded with #379)
2. double-score computation → **already-fixed** (prune scores each fact once: `list_active_facts_scoring` → single `prune_atomic` pass)
3. `compute_importance` output bounds never asserted → **PR #905** (proptest property: score ∈ [0, Σweights], weight-sum computed from the policy)
[...]

PR #907, opened 2026-06-29
test(mcp): handle_forget negative-path coverage

dutiona:
Closes the #497 sub-finding #6 coverage gap.

The MCP `handle_forget` handler (`memory-engine/bin/mcp/src/tools/handlers/p1.rs`) already validates `half_life_overrides`:
- invalid override keys are rejected via `parse_fact_type` (fact-type parse error);
- a non-object `half_life_overrides` value returns the must-be-a-JSON-object validation error.

Only the happy path (`Episodic: 30.0`) was tested. This adds the missing negative-path coverage in `memory-engine/bin/mcp/tests/tools.rs`:
- `test_forget_invalid_override_key` — a `NotARealType` key must error;
- `test_forget_non_object_overrides` — a JSON array and a JSON string must each error.

Behavior-preserving, test-only. Both paths confirmed genuinely validated by the handler (no real gap). Note #497 is closed at the #255 epic wrap-up.

Part of #255

### DIFF
--- a/memory-engine/bin/mcp/tests/tools.rs
+++ b/memory-engine/bin/mcp/tests/tools.rs
@@ -218,7 +218,11 @@ async fn test_forget_invalid_override_key() {
         &memory_engine::ActivityFilterConfig::default(),
     )
     .await;
-    assert!(result.is_err());
+    // An unknown FactType key surfaces as `INVALID_PARAMS` via the
+    // `parse_fact_type` -> `ValidationError::UnknownFactType` -> `ErrorData`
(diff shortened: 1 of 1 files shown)


dutiona:
**Review disposition (finish-pr):** the gemini `is_err()` false-positive threads are **addressed** in 23369a1 — both negative-path tests now assert the specific `ErrorCode::INVALID_PARAMS` (plus the "must be a JSON object" message for the non-object cases), matching the sibling `test_forget_validation_error` idiom. A mutation kill-test confirmed the assertions fail iff the handler validation is removed. Resolving.

PR #912, opened 2026-06-29
fix(storage): archive() handles boundary edges (one archived endpoint) without FK violation

### LINKED ISSUE #847: fix(storage): archive() fails with FK violation when a boundary edge crosses the archive set

dutiona:
## Root cause

`MemoryEngine::archive` failed with `FOREIGN KEY constraint failed` whenever an edge had **exactly one** endpoint in the archive candidate set — a *boundary* edge (`archived → survivor`, or `survivor → archived`).

`commit_archive_atomic` (`storage/sqlite/cold_storage.rs`) deleted edges via `EdgeStore::hard_delete_by_facts`, which by design removes only edges whose **both** endpoints are archived. A surviving boundary edge therefore still referenced an about-to-be-hard-deleted archived fact through `edges.source_fact_id/target_fact_id REFERENCES facts(id)` (the FK has **no** `ON DELETE CASCADE`). The subsequent `FactStore::hard_delete_ids` then tripped that FK and aborted the entire archive transaction.

## Fix

The archived fact leaves the live DB for cold storage, and the boundary edge is **not** part of the `.pak` (only internal — both-endpoints-archived — edges are archived). So the live link is genuinely dangling and must go with the fact.

- New `EdgeStore::hard_delete_incident_to_facts(fact_ids)` deletes every edge with **either** endpoint in the set (internal + boundary).
- `commit_archive_atomic` now calls it instead of `hard_delete_by_facts`, inside the same transaction → FK satisfied.
- A mere soft-expire (the issue's "Option A") would **not** work: it leaves the edge row, and its FK column, in place, so the fact hard-delete would still violate the FK. Only a hard-delete is referentially safe here ("the link is gone with the fact").
- The existing `hard_delete_by_facts` (both-endpoints contract) is left untouched.
- The in-memory graph already prunes by single fact (`remove_edges_by_fact`), so it stays consistent: the survivor keeps its node but loses the dangling boundary edge, mirroring the DB.

## Test plan

TDD — the regression tests fail without the fix (pre-fix: `Storage(Backend("FOREIGN KEY constraint failed"))`):

- `commit_archive_atomic_removes_boundary_edge_without_fk_violation` (store-level, direct atomic-commit path).
- `archive_removes_boundary_edge_without_fk_violation` (engine-level, full `archive()` path + in-memory graph consistency via `graph_degree`).
- `hard_delete_incident_to_facts_*` (3 unit tests for the new EdgeStore method: internal+boundary removal, unrelated-edge preservation, empty-slice no-op).

Full CI matrix, all green:

| Gate | Result |
| --- | --- |
| `cargo fmt --all` | clean |
| `cargo build --workspace` | OK (2 pre-existing warnings unrelated to this PR) |
| `cargo build --workspace --all-features` | OK |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | No issues found |
| `cargo test --workspace --all-features` | 1879 passed, 76 ignored, 0 failed |

Fixes #847

### DIFF
--- a/memory-engine/lib/memory-engine/src/engine/archive.rs
+++ b/memory-engine/lib/memory-engine/src/engine/archive.rs
@@ -527,11 +527,11 @@ mod tests {
         //    hard-deletes them from the DB and prunes the nodes from the graph)
         //  - survivor: survivor -> survivor  (must be left untouched)
         //
-        // Boundary edges (exactly one archived endpoint) are deliberately *not*
-        // exercised here: they are an orthogonal, pre-existing DB-side concern
-        // (archive's `hard_delete_by_facts` only removes edges with *both*
-        // endpoints archived, so a surviving boundary edge trips the facts FK on
-        // hard-delete). That is unrelated to #332's in-memory two-phase update.
+        // Boundary edges (exactly one archived endpoint) are covered by the
+        // dedicated #847 regression below
+        // (`archive_removes_boundary_edge_without_fk_violation`); this test keeps
+        // its focus on #332's in-memory two-phase update with internal/survivor
+        // edges only.
         engine
             .storage()
             .insert_edge(&make_edge(archived_ids[0], archived_ids[1]))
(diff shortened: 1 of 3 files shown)


dutiona:
Fixed. `hard_delete_incident_to_facts` now binds the id set as a single serialized JSON array via `json_each(?1)`, reused in both IN clauses — bound-variable count is O(1) regardless of N, so SQLite's `SQLITE_MAX_VARIABLE_NUMBER` (32766 in our bundled build) is never approached. Mirrors the existing `list_internal_by_facts` / facts.rs json_each idiom. Applied the same fix opportunistically to the sibling `hard_delete_by_facts` (identical latent positional-placeholder landmine; AND semantics preserved). Added a non-vacuous regression test driving a 35000-id set (past 32766) with real facts+edges — verified it fails under the old positional binding ("too many SQL variables") and passes under json_each. Resolving.

Issue #923, opened 2026-06-30
test(qa): expand doc-test / proptest / fuzz / feature-combo coverage for storage·archive·inspect


dutiona:
Deferred test-infra subset of #500 (grouped low/info storage findings). These are genuine but **disproportionate-to-cram** items — substantial test-authoring against an async-DB public surface — split out of the #500 cleanup PR so they get scoped work rather than a rushed batch. Triaged STILL_VALID against `main` (`49afc2e`).

### Checklist
- [ ] **Runnable doc-tests for public store API** (`store/{facts,events,edges,summaries}.rs`) — `store/doctest-public-api`. Use `no_run`/`ignore` examples where a live DB+embedder is needed.
- [ ] **Runnable doc-tests for the archive public API** (`archive/{pak,search,types}.rs`) — `archive/missing-doc-tests`.
- [ ] **Runnable doc-tests for the public `MemoryEngine` inspection methods** (`engine/inspect.rs` / `inspect/*`) — `inspect/testing-no-doc-tests`.
- [ ] **Property/fuzz coverage for `build_filter_query`** (`store/events.rs`) — a `proptest` over an arbitrary `EventFilter` (and optionally a `cargo-fuzz` target) — `store/fuzz-filter-query-builder`.
- [ ] **archive feature-combo CI coverage** — `tests/archive_test.rs` runs only under `--features archive`; add a CI matrix arm (or `#[cfg(all(feature="archive", feature="ann"))]` test) exercising archive×ann — `archive/no-feature-combo-coverage` (area:build overlap).
- [ ] **Dedup the remaining integration-test embedder stubs** (`tests/inspect_test.rs`, `tests/restore_roundtrip_test.rs`) by exposing the shared `MockEmbedder` behind the `test-util` feature — `inspect/testing-fake-embed-duplication` (the `src/` unit-test copies were already deduped in #120).

### Out of this scope (dispositioned on #500)
The non-test #500 items (docs, upcaster/build_pak clones, deterministic checkpoint ordering, non-zero stats test, FactType proptest, ArchivePolicy intent doc) are in the #500 cleanup PR. Info-severity design observations (`UpcasterFn` fn-ptr→closure, `commit_archive` 7-arg param-object, VACUUM-INTO `format!` path) are dispositioned as won't-fix/accept on #500 with rationale.

Refs #500.

dutiona:
## Durable triage detail (pasted for posterity)

[...]

Issue #500, opened 2026-06-01
[super-qa] storage: low/info findings (29)


dutiona:
> ⚠️ Agent analysis; not source-verified. Grouped low/info for **area:storage**.

Grouped low/info findings for area **storage** (29):

- **low** `src/archive/types.rs:28-35` — `ArchivePolicy::default()` calls `Utc::now()`, making `Default` non-deterministic and untestable (`archive/design-archivepolicy-nondeterministic-default`)
- **low** `src/engine/archive.rs:219-268` — `commit_archive` takes 7 arguments — introduce a local value-object (`archive/design-commit-archive-too-many-args`)
- **low** `src/archive/search.rs:31-33` — `search_archives` `# Errors` omits `MemoryError::Serialization` returned by `read_pak` (`archive/documentation-search-archives-incomplete-errors`)
- **low** `src/archive/pak.rs:14-19` — `write_pak` suppressed as dead_code with forward reference to unimplemented restore tooling (`archive/documentation-write-pak-dead-code-forward-ref`)
- **low** `src/archive/types.rs:12-44` — Public API structs re-exported at crate root lack field-level documentation (`archive/documentation-public-struct-fields-undocumented`)
- **low** `src/inspect/types.rs:62-65` — FactProvenance fields importance and importance_score are completely undocumented (`inspect/documentation-factprovenance-undocumented-fields`)
- **low** `src/store/activities.rs:218, src/store/checkpoints.rs:98` — unwrap_or_default silently replaces corrupt stored JSON with null/{} (`store/documentation-silent-json-swallow`)
- **low** `src/store/scopes.rs` — ScopeStore public methods lack # Errors doc sections (`store/documentation-missing-errors-scopes`)
- **low** `src/store/activities.rs` — ActivityStore public methods lack # Errors doc sections (`store/documentation-missing-errors-activities`)
- **low** `src/store/checkpoints.rs` — CheckpointStore public methods lack # Errors doc sections (`store/documentation-missing-errors-checkpoints`)
- **low** `src/store/facts.rs` — Multiple FactStore public methods lack # Errors doc sections (`store/documentation-missing-errors-facts`)
- **low** `src/store/facts.rs` — Several FactStore methods call expect on serde_json::to_string without # Panics doc (`store/documentation-missing-panics-facts`)
[...]

PR #924, opened 2026-06-30
docs(storage): #500 cleanup — docs, redundant-clone + flaky-order fixes, coverage

dutiona:
Addresses the valuable, reasonable-effort subset of grouped issue #500 (low/info storage findings), triaged against current `main`. Three atomic commits:

1. **docs** — additive `# Errors`/`# Panics` doc sections on `store::{scopes,activities,checkpoints,facts}` methods that lacked them; `# Errors` for `search_archives` (adds `MemoryError::Serialization`); public field docs on archive `types.rs`; `FactProvenance::importance_score` doc; a note on `ArchivePolicy::default()` clarifying its now-relative cutoff is **by design** (a retention policy's default is inherently relative to now — not an oversight, not "fixed").
2. **perf/correctness** — eliminate the per-step `String` clone in the upcaster hot loop (nested `HashMap<String, HashMap<u16, _>>` so the event-type chain is looked up once by `&str`; `(type,rev)→fn` mapping, `registered_count`, and `Debug` semantics all preserved); deterministic `list_recent` ordering (`ORDER BY checkpoint_at DESC, session_id DESC` — fixes a `Utc::now()`-tie flake). `build_pak` left by-reference (the caller reuses the slices, so by-value would only relocate the clone — documented).
3. **tests** — non-zero `EdgeStats`/`ScopeStats` coverage (nested scopes; `max_depth` was always 0 before); `FactType` round-trip through the store for every variant.

**Disposition of the rest of #500:** 6 sub-findings already-fixed by earlier waves (closed with evidence on #500); the large test-infra items (public-API doc-tests, fuzz harness, archive feature-combo CI, integration-test embedder dedup) are tracked in **#923**; info-severity design observations (`UpcasterFn` fn-ptr, `commit_archive` 7-arg, VACUUM-INTO `format!` path — an inherent SQLite limitation, already mitigated) dispositioned won't-fix/accept on #500.

Refs #500.

## Verification
All CI gates pass locally: fmt, `clippy --workspace --all-targets --all-features -D warnings`, build all-features, `test --workspace --all-features` (1967 pass), cargo doc all-features.

### DIFF
--- a/memory-engine/lib/memory-engine/src/archive/search.rs
+++ b/memory-engine/lib/memory-engine/src/archive/search.rs
@@ -125,6 +125,8 @@ const fn entry_is_prunable(_entry: &ArchiveManifestEntry, _query: &MemoryQuery) 
 /// # Errors
 ///
 /// Returns `MemoryError::Archive` on I/O or decompression failure.
+/// Returns `MemoryError::Serialization` if a `.pak` file's JSON is corrupt or
+/// truncated (surfaced by `read_pak` during decompression and deserialization).
 pub fn search_archives(
     archive_dir: &Path,
     manifest_entries: &[ArchiveManifestEntry],
--- a/memory-engine/lib/memory-engine/src/archive/types.rs
+++ b/memory-engine/lib/memory-engine/src/archive/types.rs
@@ -56,14 +56,20 @@ impl Default for ArchivePolicy {
     /// Returns a policy with a 30-day look-back cutoff relative to the current
     /// wall-clock time and a 100-fact minimum batch size.
     ///
-    /// The `expired_before` cutoff is intentionally now-relative: a retention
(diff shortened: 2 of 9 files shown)


dutiona:
Applied the doc + deprecation fixes: `ArchivePolicy::default()` now (a) modernizes `chrono::Duration::days(30)` → `chrono::TimeDelta::days(30)`, and (b) documents that the cutoff is **frozen at construction** with explicit guidance to build a fresh `ArchivePolicy` (or set `expired_before`) per archival run in a long-running process. The deeper refactor — changing the field from an absolute `DateTime<Utc>` to a relative `TimeDelta` so the horizon slides automatically — is a **breaking public-API change** (needs-design) and out of scope for this low/info cleanup; documenting the construct-per-run contract is the proportionate fix here. Resolving.

Issue #903, opened 2026-06-29
fix(build): re-attach orphaned eval test harness — 56 tests silently dark since #814


dutiona:
## What

The two-tier eval harness (`tests/eval/main.rs` — **56 conformance + quality tests**, including the per-area `quality/{forgetting,...}.rs` suites) stopped being compiled/run by `cargo test --workspace` after the **#814 bin/lib restructure**.

## Why

Cargo auto-discovers integration tests under a package's own `tests/` dir. #814 moved the core crate to `memory-engine/lib/memory-engine/`, but the eval harness lives at the **repo root** `tests/eval/` — which is no longer adjacent to any crate manifest. With no explicit `[[test]]` target pointing at it, Cargo silently skipped the entire harness. No error, no warning — the 56 tests just stopped running.

## Blast radius (repo-wide)

These tests have been **dark since #814**. Any PR that added or modified `tests/eval/` tests since then had **false-green CI** — the new tests never actually ran. This affects the concurrent **#258 (storage)** and **#259 (temporal)** super-qa sweeps if they touched `tests/eval/`, as well as the pre-existing `forgetting` eval tests the #255 triage referenced.

## Fix

Add an explicit `[[test]] name = "eval"` target (path to `tests/eval/main.rs`) in `memory-engine/lib/memory-engine/Cargo.toml`, mirroring the other explicitly-attached top-level integration tests. The harness then compiles and runs under `cargo test --workspace` again.

Discovered during the #255 area:forgetting sweep (PR #902, whose #312 integration test could not run while the harness was orphaned). Split out per the no-silent-descope contract. Part of #818 (Engineering Health). Refs #255 #902 #814.

Issue #816, opened 2026-06-26
refactor(build): Wave 2 — decompose core into per-primitive crates + opt-in thin-bin/dylib


dutiona:
Follow-up to **Wave 1** (PR #814), which did the mechanical `bin`/`lib`/`utils` layout
move. Wave 2 is the substantive part: make the library/binary boundary and link graph
explicit, and decompose the monolith.

### Scope
- **Decompose the core** (`memory-engine/lib/memory-engine`) into per-primitive crates:
  `ingest` / `query` / `consolidate` / `forget` / `resolve`, plus `storage` and
  `types`/`traits`. `memory-engine` becomes the **facade** that re-exports them.
  - Untangle any inter-primitive cycles (crates must form a DAG).
  - Re-gate `pub(crate)` → `pub` carefully at the new crate boundaries.
- **Link legibility**: a `just`/`make` target surfacing the link graph + shipped surface
  (`cargo tree`, `cargo-bloat`, `nm -D` of release artifacts, `--print native-static-libs`).
- **Opt-in thin-bin + `dylib`** build mode (default stays **fat-bin, static LTO**).
  Eyes open: Rust `dylib` ABI is unstable → ship lib+bins as one matched set; no
  independent upgrade. The C-ABI options are off the table (no foreign consumers).
- **Revisit test-fixture placement**: the Wave-1 `include_str!("../../../../../tests/...")`
  in `src/engine/tests.rs` is brittle (re-breaks if the file moves in the split).

### Status
Placeholder tracking issue — promote/redesign as a proper `type:epic` with your own
sub-issue breakdown if preferred. Full rationale (Rust linking model, trade-offs) is in
the PR #814 description and the Wave-1 design discussion.

dutiona:
**Wave 2 (#816) design + plan are archived in #925** — the canonical, navigable record (design §1–13 + plan preamble + §A–M as comments, plus the §13.2 epic routing table). In-tree copies under `docs/superpowers/specs/` and `docs/plans/` are removed at the Wave 2 finish-PR per the archive-plans-as-issues convention. Per-slice PRs (S1–S6) link to #925.

dutiona:
## ✅ S2 COMPLETE — Wave 2 status + S3→S6 roadmap & carry-forward gotchas

**main @ `911e70d`** · workspace = **10 members** · all worktrees/branches cleaned (only `main`).

### Done

| Slice | Crates carved | Merge |
|---|---|---|
[...]

Issue #633, opened 2026-06-19
feat(storage): PgBackend pool + schema + migrations


dutiona:
`PgBackend` skeleton behind the `backend-postgres` feature: `deadpool-postgres` over `tokio-postgres`, a fresh PG migration chain (no FK-rebuild hack; `GENERATED ALWAYS AS (to_tsvector(...)) STORED` instead of FTS triggers), `SchemaManager` impl, read-only via `default_transaction_read_only`. `compile_error!` if neither backend feature is enabled.

**Part of** #628 (epic).
**Depends on** #628 A1.
**Spec:** `docs/superpowers/specs/2026-06-19-pluggable-storage-backend-design.md` — §Section 4

dutiona:
## 🧭 Bootstrap handoff for #633 (PgBackend pool + schema + migrations)

### Current state (main `adb137a`, green + clippy-clean under `--all-features`)
A1 #629 ✅ (trait family) · A2 #630 ✅ (`SqliteBackend`) · A3 #631 ✅ (engine async-native, holds `Arc<dyn StorageBackend>`) · **A4 #632 ✅ MERGED** — the cross-backend conformance battery (`src/storage/conformance/`) is the harness #633's `PgBackend` must eventually pass. Already scaffolded and waiting: the inert **`backend-postgres = []`** feature, a `PgFactory` with 3 `todo!()`s (`factory.rs:131/135/140`), and `#[ignore = "#633"]` tokens on the `postgres`/`postgres_cold` suite arms (`mod.rs:135/147`).

#633 is **B1 — the FIRST PG piece**. #634 (CRUD: FactGraph/EventLog/ConsolidationStore/SessionStore) and #635 (SearchIndex: lexical tiers + pgvector) follow. **`PgBackend` does not become a full `StorageBackend` until #634+#635** (the blanket impl needs all 6 bounded traits), so **#633 does NOT flip the conformance `#[ignore]`** and cannot wire `PgFactory::make()` to return `Arc<dyn StorageBackend>` yet. #633 tests its own SchemaManager + pool + migrations against a real PG testcontainer.

### What #633 is
[...]

PR #931, opened 2026-07-02
refactor(core): decouple rusqlite from me-types, delete MemoryError::Database

### LINKED ISSUE #926: refactor(core): decouple rusqlite from MemoryError / me-types (L0 leaf hygiene)
## Problem

`rusqlite` (the SQLite driver) leaks onto the workspace's **public, user-facing error type** and, via the orphan rule, pins itself to the **L0 base crate** `me-types`.

The single offending item is [`MemoryError::Database(#[from] rusqlite::Error)`](../blob/main/memory-engine/lib/me-types/src/error.rs):

```rust
/// An underlying `SQLite` operation failed (query, statement, or connection).
#[error("database error: {0}")]
Database(#[from] rusqlite::Error),
```

`MemoryError` is the public error type of the whole engine (`MemoryEngine::* -> Result<T, MemoryError>`). Two consequences:

1. **Public-API leak** — the concrete SQLite driver type `rusqlite::Error` is exposed in the public surface. A backend-agnostic engine should not name a driver type in its error enum.
2. **Orphan-rule pin** — `#[from]` synthesises `impl From<rusqlite::Error> for MemoryError`, which must live in the crate that defines `MemoryError` (now `me-types`, the L0 base of the Wave 2 #816 DAG). So `me-types` is forced to depend on `rusqlite` + `bundled`, compiling the SQLite C library into **every** crate in the workspace — including a hypothetical Postgres-only build.

This is a **half-finished migration**: #560 already introduced the correct driver-free abstraction — `StorageError` (same file), whose doc states driver errors "are mapped to an opaque `String` inside each backend and never cross the trait seam" and that the `Database` variant merely "stays for `SQLite`-internal use". This issue finishes that arc.

## Why it is NOT urgent (and why it is tracked, not folded)

`rusqlite` is an _external_ crate, so `me-types -> rusqlite` is **not** a workspace cycle — the Wave 2 DAG is acyclic and green with it in place. This is an API-cleanliness + build-hygiene concern, not a layering cycle. It was deliberately kept **out** of the E.4b `me-storage` carve (whose scope is the crate-boundary move) to avoid ballooning a delicate transactional refactor with an orthogonal ~224-site error change.

## Blast radius (measured on `worktree-wave2-s1-keystone`)

- `MemoryError::Database` references (non-test): **224**, across ~48 files.
- `rusqlite` references total: 226 — `me-types` 7, `me-traits`/`me-embed` 0, core `memory-engine` 205 (of which `store/` = 121, the legacy SQLite internals bound for `me-backend-sqlite` in S2), CLI 9, MCP 5.

Removing `#[from]` breaks every `?` that currently auto-converts `rusqlite::Error -> MemoryError::Database`; each becomes an explicit map at the backend.

## Design options (decide at execution)

1. **`Database(String)`** — keep the variant, change its payload to `String`; backends map `rusqlite::Error`/`tokio_postgres::Error -> String` at the seam. Smallest consumer-side churn (matches on `Database(_)` still compile).
2. **Delete `Database`, route through `StorageError::Backend(String)`** — the fuller #560 "one variant per recoverable cause" arc. Cleaner end state, but consumers matching `MemoryError::Database(_)` must move to `MemoryError::Storage(StorageError::Backend(_))` (larger consumer-side change).

## Payoff / timing

Do it at **S1-tail or S2-start**, when the backend split makes the win concrete:

- `me-types` (and thus a Postgres-only build) sheds the bundled-SQLite compile.
- Unblocks **#633**'s future `dep:rusqlite` feature-gating (`rusqlite` compiled only under `backend-sqlite`).

## Acceptance

- [ ] `me-types/Cargo.toml` no longer lists `rusqlite`.
- [ ] No `rusqlite` (or any driver) type appears in `me-types` public API.
- [ ] Full CI gate green (`--workspace --all-features`), no test-count regression.
- [ ] `cargo tree -p me-types` shows no `rusqlite`.

_Filed from the Wave 2 #816 S1 execution session; part of the epic #628 pluggable-storage arc._

dutiona:
## Summary

Decouples `rusqlite` from the L0 `me-types` crate by deleting the public `MemoryError::Database(#[from] rusqlite::Error)` variant. Backend driver errors now surface via `MemoryError::Storage(StorageError::Backend(String))`, mapped at each backend call site. This is the S2-start L0-hygiene prerequisite for the Wave 2 backend carve (epic #816) and the real unblocker for #633's `dep:rusqlite` feature-gating.

Fixes #926.

## Why

`MemoryError::Database(#[from] rusqlite::Error)` embedded the SQLite driver's error type in the workspace's public error enum. The orphan rule forced `impl From<rusqlite::Error> for MemoryError` to live in `me-types` (where `MemoryError` is defined), pinning `rusqlite` + bundled SQLite into L0 — compiled by **every** workspace crate, including a hypothetical Postgres-only build.

## What (Option 2 — the fuller #560 driver-free-error arc)

- **L0 (`me-types`):** delete the `Database` variant; add a generic `StorageError::backend(impl Display)` helper (takes any `Display`, so L0 needs no driver dep); drop `rusqlite` from `me-types/Cargo.toml`. `cargo tree -p me-types` → **0 rusqlite**.
- **Seam guarantee:** delete `storage::sqlite::map_seam_err`. The D4 "no driver type crosses the seam" invariant moves from a runtime boundary match (`block_read`/`block_write`/`for_each_streamed`) to **compiler-enforced source-mapping** via `.map_err(StorageError::backend)?`. The postgres backend's `pg_err` already worked this way — unaffected, and now the two backends are symmetric.
- **~430 call sites:** `rusqlite::Error` `?`-sites → `.map_err(StorageError::backend)?`; explicit constructions/matches repointed to `Storage(Backend)`. `#[non_exhaustive]` on `MemoryError` means only explicit `Database(_)` arms broke.
- **2 semantic corrections** (the deletion forced each mis-mapped site to declare its true kind): a `serde_json` failure that was wrapped as a *fake* `rusqlite::Error::ToSqlConversionFailure` now maps to `MemoryError::Serialization`. The `QueryReturnedNoRows`→`NotFound` inspections already lived in separate, untouched arms — no semantic distinction was collapsed.
- **179 doc references** repointed `MemoryError::Database` → `MemoryError::Storage` (fixes 23 intra-doc gate-blocking links, incl. one in `me-storage` the variant deletion broke).
- **ADR-0018:** approved public-API-break ledger extended 2 → 3 (this is the 3rd deliberate, gated Wave-2 break).

## Verification

- Full CI-exact §I gate: fmt, clippy (`-D warnings`, pedantic+nursery), build ×2, test (`--all-features`), doc ×2 (broken-intra-doc-link denied), `cargo deny`, **both MSRV `--tests --examples` poles** — all green.
- **Acceptance criteria (#926):** `me-types/Cargo.toml` has no `rusqlite`; no driver type in `me-types` public API; `cargo tree -p me-types` shows no `rusqlite`; full gate green.
- **anti-#903:** net **−1 test** — mcp `database_maps_to_internal` removed because its subject variant `MemoryError::Database` no longer exists; its behavior (a backend error → MCP `internal_error`) is preserved by the pre-existing `storage_maps_to_internal` (ex-`Database` errors now flow through `Storage`). Not a dark test.

## Blast radius

Behavior-preserving except the intentional `serde_json`→`Serialization` correction and the documented −1 test. Public API: one variant removed (gated, ADR-recorded). Consumers matching `MemoryError::Database(_)` migrate to `MemoryError::Storage(StorageError::Backend(_))`.

**MCP surface (one user-visible text change):** backend/SQL failures previously surfaced through the deleted `Database` arm as `"database error: …"`; they now flow through the pre-existing `Storage` arm as `"storage error: …"` — **same** `INTERNAL_ERROR` JSON-RPC code, only the human-readable message *prefix* changes (no machine-parseable contract moved). The `storage_maps_to_internal` test now pins the exact `"storage error:"` prefix, restoring — on the new text — the guard the removed `database_maps_to_internal` provided (this is why the net is −1 test, not −2: coverage was consolidated, not dropped).

### DIFF
--- a/memory-engine/bin/mcp/src/error.rs
+++ b/memory-engine/bin/mcp/src/error.rs
@@ -27,8 +27,6 @@ pub fn to_mcp_error(err: MemoryError) -> ErrorData {
         // maps to invalid_params alongside `Conflict` — not the internal-error default.
         MemoryError::Cycle(e) => ErrorData::invalid_params(format!("cycle error: {e}"), None),
 
-        MemoryError::Database(e) => ErrorData::internal_error(format!("database error: {e}"), None),
-
         MemoryError::Serialization(e) => {
             ErrorData::internal_error(format!("serialization error: {e}"), None)
         }
@@ -152,13 +150,6 @@ mod tests {
         let mcp = to_mcp_error(err);
         assert_eq!(mcp.code, rmcp::model::ErrorCode::INVALID_PARAMS);
         assert!(mcp.message.contains("model mismatch"));
-    }
-
-    #[test]
-    fn database_maps_to_internal() {
-        let err = MemoryError::Database(rusqlite::Error::QueryReturnedNoRows);
-        let mcp = to_mcp_error(err);
-        assert_eq!(mcp.code, rmcp::model::ErrorCode::INTERNAL_ERROR);
     }
 
     #[test]
--- a/memory-engine/bin/mcp/src/error.rs
+++ b/memory-engine/bin/mcp/src/error.rs
@@ -167,7 +167,10 @@ mod tests {
         let err = MemoryError::Storage(StorageError::Backend("boom".into()));
         let mcp = to_mcp_error(err);
         assert_eq!(mcp.code, rmcp::model::ErrorCode::INTERNAL_ERROR);
-        assert!(mcp.message.contains("storage error"));
+        // #926: backend failures used to surface via the deleted `Database` arm as
+        // "database error: …"; they now flow through `Storage` as "storage error: …".
+        // Pin the exact prefix so that user-visible text change can never silently regress.
+        assert!(mcp.message.starts_with("storage error:"), "{}", mcp.message);
     }
 
     #[test]
--- a/memory-engine/lib/memory-engine/src/store/activities.rs
+++ b/memory-engine/lib/memory-engine/src/store/activities.rs
@@ -1,5 +1,6 @@
 //! Store operations for activity records.
 
+use crate::error::StorageError;
 use rusqlite::{Connection, OptionalExtension, params};
 
 use crate::error::{MemoryError, Result};
@@ -26,7 +27,7 @@ impl<'a> ActivityStore<'a> {
     ///
     /// # Errors
     ///
-    /// Returns `MemoryError::Database` on SQL failure (e.g. FK violation if
+    /// Returns `MemoryError::Storage` on SQL failure (e.g. FK violation if
     /// `activity.scope_id` references a non-existent scope).
     pub fn insert_or_dedup(
         &self,
@@ -61,7 +62,7 @@ impl<'a> ActivityStore<'a> {
                 |row| Ok((row.get(0)?, row.get(1)?)),
             )
             .optional()
-            .map_err(MemoryError::Database)?;
+            .map_err(StorageError::backend)?;
 
         if let Some((existing_id, count)) = existing {
             self.conn
@@ -74,7 +75,7 @@ impl<'a> ActivityStore<'a> {
                      WHERE id = ?4",
                     params![count + 1, ts, activity.result_summary, existing_id],
                 )
-                .map_err(MemoryError::Database)?;
+                .map_err(StorageError::backend)?;
             return Ok((existing_id, true));
         }
 
@@ -96,7 +97,7 @@ impl<'a> ActivityStore<'a> {
                     activity.scope_id,
                 ],
             )
-            .map_err(MemoryError::Database)?;
+            .map_err(StorageError::backend)?;
 
         let id = self.conn.last_insert_rowid();
         Ok((id, false))
@@ -110,7 +111,7 @@ impl<'a> ActivityStore<'a> {
     /// # Errors
     ///
     /// Returns `MemoryError::NotFound` if no activity with `id` exists.
-    /// Returns `MemoryError::Database` on SQL failure.
+    /// Returns `MemoryError::Storage` on SQL failure.
     pub fn get(&self, id: i64) -> Result<Activity> {
         self.conn
             .query_row(
@@ -125,7 +126,7 @@ impl<'a> ActivityStore<'a> {
                 rusqlite::Error::QueryReturnedNoRows => {
                     MemoryError::NotFound(format!("activity {id}"))
                 }
-                other => MemoryError::Database(other),
+                other => StorageError::backend(other).into(),
             })
     }
 
@@ -136,7 +137,7 @@ impl<'a> ActivityStore<'a> {
     ///
     /// # Errors
     ///
-    /// Returns `MemoryError::Database` on SQL failure.
+    /// Returns `MemoryError::Storage` on SQL failure.
     pub fn list_by_session(&self, session_id: &str, limit: Option<usize>) -> Result<Vec<Activity>> {
         let base = "SELECT id, session_id, tool_name, args_hash, args, result_summary,
                         outcome_class, status, occurrence_count, first_seen, last_seen,
@@ -146,12 +147,12 @@ impl<'a> ActivityStore<'a> {
                  ORDER BY last_seen DESC";
         let limit_i64: i64 = limit.map_or(-1, |n| i64::try_from(n).unwrap_or(i64::MAX));
         let sql = format!("{base} LIMIT ?2");
-        let mut stmt = self.conn.prepare(&sql).map_err(MemoryError::Database)?;
+        let mut stmt = self.conn.prepare(&sql).map_err(StorageError::backend)?;
         let rows = stmt
             .query_map(params![session_id, limit_i64], row_to_activity)
-            .map_err(MemoryError::Database)?;
+            .map_err(StorageError::backend)?;
         rows.collect::<std::result::Result<_, _>>()
-            .map_err(MemoryError::Database)
+            .map_err(|e| StorageError::backend(e).into())
     }
 
     /// List recent activities for a set of scope IDs, most recent first.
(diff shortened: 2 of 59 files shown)


memory-engine/bin/mcp/src/error.rs
```diff
@@ -27,8 +27,6 @@ pub fn to_mcp_error(err: MemoryError) -> ErrorData {
         // maps to invalid_params alongside `Conflict` — not the internal-error default.
         MemoryError::Cycle(e) => ErrorData::invalid_params(format!("cycle error: {e}"), None),
 
-        MemoryError::Database(e) => ErrorData::internal_error(format!("database error: {e}"), None),
-
         MemoryError::Serialization(e) => {
```
dutiona, line 30: [Codex Review] Deleting the dedicated Database arm changes the MCP error text for backend failures from "database error: ..." to "storage error: ..." because SQLite faults now arrive through the Storage variant. The PR description for #931 only calls out the serde reclassification and the deleted MCP test as intended behavior changes, so this is a third user-visible contract change on the MCP surface. If any client or golden test keys off the old wording, it will regress silently because the dedicated database-mapping test was removed rather than rewritten. Please either preserve the legacy database prefix for backend-storage failures here or document this contract change and add a replacement test.

memory-engine/lib/memory-engine/src/store/activities.rs
```diff
@@ -160,14 +161,12 @@ impl<'a> ActivityStore<'a> {
     ///
     /// # Errors
     ///
-    /// Returns `MemoryError::Database` on SQL failure.
+    /// Returns `MemoryError::Storage` on SQL failure.
```
dutiona: [Codex Review] This doc comment is stale after the reclassification in this PR. The function now does `serde_json::to_string(scope_ids).map_err(MemoryError::Serialization)?`, so callers can see `MemoryError::Serialization` before any SQL runs, but the rustdoc still says only `Storage` is possible. Since `list_recent_by_scope` is one of the two call sites whose error kind changed semantically in #931, the doc sweep should mention the new Serialization path explicitly.

memory-engine/lib/memory-engine/src/store/events.rs
```diff
@@ -39,7 +40,7 @@ pub const fn event_type_to_str(et: &EventType) -> &'static str {
 ///
 /// What a *read-path* caller sees, end-to-end: the read site ([`row_to_event`])
 /// boxes this `Internal` into `rusqlite::Error::FromSqlConversionFailure`, which
-/// re-maps to [`MemoryError::Database`] via its `#[from]`. So the surfaced
+/// re-maps to [`MemoryError::Storage`] via its `#[from]`. So the surfaced
```
dutiona: [Codex Review] The code now correctly asserts `MemoryError::Storage(StorageError::Backend(_))`, but this explanatory block still says the read path surfaces `Database`, and the test name/message keep the old contract too. That leaves the maintainer-facing docs contradicting the actual assertion immediately below. Since #931 is specifically the Database-variant removal PR, these leftover references should be scrubbed here and in the matching `store/facts.rs` test block so the docs reflect the new seam error kind consistently.

dutiona (commented):
[Codex Review] I found 3 issues.

1. `memory-engine/bin/mcp/src/error.rs`: deleting the `Database` arm changes MCP backend-failure text from `database error: ...` to `storage error: ...`, which is an extra user-visible behavior change beyond the ones called out in the PR description.
2. `memory-engine/lib/memory-engine/src/store/activities.rs`: `list_recent_by_scope` now returns `MemoryError::Serialization` before any SQL runs, but its rustdoc still says only `Storage` is possible.
3. `memory-engine/lib/memory-engine/src/store/events.rs` (and the matching `store/facts.rs` block): the explanatory prose/test naming still says corrupt read-path errors surface `Database` even though the assertion now checks `Storage(StorageError::Backend(_))`.

Inline comments have details.

memory-engine/bin/mcp/src/error.rs
dutiona: [Codex Review reply] Addressed in 12cc52b. Correct — the prefix changed ("database error:" -> "storage error:") as a consequence of deleting the Database arm; same INTERNAL_ERROR code, only the human-readable message. Rather than re-add database_maps_to_internal (it tested a now-impossible variant), I strengthened storage_maps_to_internal to pin the exact prefix: assert(mcp.message.starts_with("storage error:")) (was .contains("storage error")). That restores the guard on the new text so the prefix cannot silently regress. Also documented in the PR body under Blast radius / MCP surface.

memory-engine/lib/memory-engine/src/store/activities.rs
dutiona: [Codex Review reply] Fixed in 12cc52b. list_recent_by_scope # Errors doc now reads: "Returns MemoryError::Serialization if the scope-id list cannot be JSON-encoded, or MemoryError::Storage on backend/SQL failure." — matching the serde_json::to_string(...).map_err(MemoryError::Serialization)? reclassification.

memory-engine/lib/memory-engine/src/store/events.rs
dutiona: [Codex Review reply] Fixed in 12cc52b across all three axes (prose, test name, assertion message), in both events.rs and the mirrored facts.rs block. The prose now describes the actual new two-hop path: the row-mapper boxes Internal into rusqlite::Error::FromSqlConversionFailure, the call site .map_err(StorageError::backend) stringifies it into StorageError::Backend, and ? lifts that to MemoryError::Storage (the old #[from]-on-rusqlite mechanism is gone). Surfaced variant: Storage(StorageError::Backend), with the boxed Internal message preserved as a substring of the Backend string. Tests renamed ..._surfaces_database_not_notfound -> ..._surfaces_storage_not_notfound; assertion messages now say Storage(Backend). Same corrections applied to embedding_spaces.rs and the facts.rs timestamp test.

dutiona (commented):
[Codex Review] Confirmation pass on delta `12cc52b1fbee5ad7b32c45b0be777a4cdde3e88b`: the three round-1 findings are resolved, and I found no residual issues in this delta.

Issue #940, opened 2026-07-04
refactor(core): S4 — carve MODERATE primitives me-query/me-ingest/me-consolidate/me-archive (Wave 2 #816)


dutiona:
S4 of the Wave-2 decomposition (plan #925). Carve **me-query** (RRF/hybrid merge + MemoryQuery; the backend search machinery already sits in me-backend-sqlite behind SearchIndex — XR-2), **me-ingest**, **me-consolidate** (the compute_* orchestration; apply_* SQL already in me-backend-sqlite), **me-archive** (.pak format DTOs + logic). Includes **public API-break #2**: VectorSearchStrategy::search → &dyn SearchIndex. Carve DoD applies.

dutiona:
### S4 sub-PR 1 (`me-ingest`) ✅ MERGED

PR #974 squash-merged → `main@2cd5167`. The **Ingest** primitive is now L3 crate `me-ingest` (5 methods + 5 helpers as free fns over `MemoryCtx` + scope_tree/embedder/classifier). Workspace = **14 crates**.

- Full super-review **4/4 CLEAN** (Codex + Antigravity + advisor + code-reviewer subagent byte-diffing all 10 fns + 2 copies).
- anti-#903: **1967/0** exact; me-ingest ships 0 tests (facade tests stay); zero back-deps.
- The S4 **template-setter** — proved the `mem_ctx()` seam scales to a multi-method, multi-capability primitive (EmbeddingProvider + PersistenceClassifier + scope_tree + spawn_blocking).
- Follow-up **#973**: `cache_scope_chain` (shared ScopeTree hydration) carries a deliberate inline copy in me-ingest; shared-home decision deferred to a coordinated sweep (sequenced with me-consolidate).

**Next: sub-PR 2 = `me-query`** (backend-free RRF/hybrid/`MemoryQuery` — the search SQL already moved to me-backend-sqlite in S2) + **API-break #2** (`VectorSearchStrategy::search` → `&dyn SearchIndex`, bundled per maintainer decision).

dutiona:
### S4 sub-PR 2 — `me-query` ✅ MERGED (squash `e42fc57`, PR #975)

[...]

Issue #628, opened 2026-06-19
epic(storage): pluggable StorageBackend abstraction (SQLite + Postgres)


dutiona:
## Summary

Introduce a pluggable **`StorageBackend`** abstraction so consumers can run `memory-engine` on **PostgreSQL** — while **SQLite stays the default in-process backend**. Today persistence is wired directly to SQLite (~219 inline SQL sites across `src/store/*` and the engine, ~42 files touching `rusqlite`, no abstraction trait).

**Driving requirement:** an adopter standardized on PostgreSQL for cloud-scaled deployment should not have to operate a second DB technology. We support Postgres in **two tiers** plus SQLite, all behind one trait family:

1. **SQLite** (default, in-process) — `bm25()` baseline.
2. **Postgres, bundled-BM25** — self-hosted reference `postgres + bm25` Docker image (`pg_search`/`pg_textsearch`). True BM25 parity.
3. **Postgres, stock managed** — RDS/Cloud SQL/Azure with `tsvector` + `ts_rank_cd` + `pgvector`. Lexical degraded (no IDF); vector first-class via `pgvector`.

We **do not silently degrade**: `LexicalMode::RequireBm25` fails fast at open if no BM25 extension is present. We **measure** the degradation (piece D) rather than asserting it.

> 📄 Full design spec committed in-repo: `docs/superpowers/specs/2026-06-19-pluggable-storage-backend-design.md`
> The complete spec is also reproduced **verbatim, section by section, in the comments below**.

## Why this matters (research highlights)

- PostgreSQL has full-text search but **not BM25**: `ts_rank`/`ts_rank_cd` use **no corpus IDF**, no TF saturation, length-norm off by default (PG docs: _"the ranking functions do not use any global information"_).
- Real-BM25 Postgres extensions are **not installable on any managed cloud** (RDS/Aurora, Cloud SQL/AlloyDB, Azure Flexible Server). `pgvector`, by contrast, **is** on every managed allowlist — so **BM25 is the only capability that needs the self-hosted bundle**.
- Our hybrid retrieval merges via **RRF**, which consumes **rank order only, not raw scores** — so the BM25-vs-`ts_rank` gap is a _measurable quality property, not an API incompatibility_. The `SearchIndex` trait returns **ranked ids**, absorbing the difference at the seam.

## Decisions (the four forks)

| #   | Decision                | Choice                                                                           |
| --- | ----------------------- | -------------------------------------------------------------------------------- |
[...]

Issue #634, opened 2026-06-19
feat(storage): PgBackend graph/event/consolidation/session CRUD


dutiona:
Implement FactGraph, EventLog, ConsolidationStore, SessionStore (and ColdStorage under `archive`) for PgBackend. `FactFilter`/`MetadataPredicate` translate to `jsonb` operators; timestamps as `timestamptz`; embeddings via `pgvector` `vector(dim)`. Must pass the A4 conformance suite.

**Part of** #628 (epic).
**Depends on** #628 B1.
**Spec:** `docs/superpowers/specs/2026-06-19-pluggable-storage-backend-design.md` — §Section 4, §Section 6

dutiona:
## 🧭 Bootstrap handoff for #634 (PgBackend graph/event/consolidation/session CRUD)

### Current state (main, B1 #633 ✅ MERGED — `09acffc`)
The `backend-postgres` feature now ships a **`PgBackend` skeleton**: a `deadpool-postgres` pool (`src/storage/postgres/pool.rs`), a fresh native-PG v14 migration chain (`migrations.rs` — all 12 tables exist, incl. `vector(N)` embedding columns, `embedding_spaces`, `fact_vectors`), the `SchemaManager` impl (`schema.rs`), the **native-async `with_client` seam** (`mod.rs` — no `spawn_blocking`; MVCC, so no read-pool/write-mutex), and a **gated testcontainer harness** (`tests.rs` — the `spin_pg`/`migrated_backend` helpers + a schema-parity introspection proof, 18 `#[ignore]` tests). #634 **extends** this module.

### What #634 is
Implement the four **data-CRUD** bounded traits for `PgBackend`, in PG dialect: **`FactGraph`** (the big one — add/expire/edges/bi-temporal/`FactFilter` queries), **`EventLog`** (append-only event sourcing + upcasting), **`ConsolidationStore`**, **`SessionStore`**, plus **`ColdStorage`** under `archive`. Mirror the `src/storage/sqlite/{graph,event_log,consolidation,session,convert,cold_storage}.rs` impls — but see the #1 trap. After #634, `PgBackend` has 5 of 6 traits; **`SearchIndex` (#635) completes the set**, so **#634 does NOT flip the conformance arm** (the blanket impl is all-or-nothing; `PgFactory::make()` stays `todo!()` until #635). #634 tests CRUD via the **direct gated harness** (extend `spin_pg`); #635 flips the arm + runs the full battery over #634's CRUD.

[...]

Issue #117, opened 2026-03-22
[super-qa] medium: scope resolution duplicated across query paths


dutiona:
## Finding: Scope Resolution Duplication

- **ID:** M17
- **Severity:** medium
- **Category:** refactoring
- **Location:** src/engine.rs:431-502, 583-647

## Description

Both \`query()\` and \`execute_query()\` contain identical scope resolution logic (read scope_tree, call resolve_query, return empty on None) and ANN strategy dispatch.

## Suggested Approach

Extract: \`fn resolve_scope_and_strategy(&self, scope: Option<&ScopeQuery>) -> Result<(Option<Vec<i64>>, &dyn VectorSearchStrategy)>\`

PR #976, opened 2026-07-12
refactor(build): hoist shared query+math vocabulary to me-types

dutiona:
## What

**S4 sub-PR 3a of 5** — a **layer-correction relocation**. No new crate, no public-API break. It is the prerequisite that makes sub-PR 3b (carving `me-archive` as a clean L3 leaf) legal.

## Why

Triage-against-current-main, run *before* writing the me-archive carve spec, surfaced a **DAG blocker the #925 plan did not anticipate**: the archive code reaches three items sitting in the wrong layer.

| item archive needs | was in | layer | edge it would force |
|---|---|---|---|
| `MemoryQuery` | `me-query` | **L3** | `me-archive → me-query` = **L3→L3 sibling edge — ILLEGAL** |
| `cosine_similarity` | `me-backend-sqlite` | L2 | `me-archive →` **concrete backend** |
| `CURRENT_SCHEMA_VERSION` | `me-backend-sqlite` | L2 | `me-archive →` **concrete backend** |

The two L2 edges are the insidious ones: **legal by layer number, architecturally fatal**. They would drag **rusqlite + bundled SQLite into a Postgres-only build**, silently gutting epic #628 while every gate stayed green.

## The changes

**1. `MemoryQuery` → `me-types`.** A pure data+builder DTO (verified: zero `me-query`-internal deps) whose **entire sibling vocabulary** — `SearchQuery`, `SearchResult`, `QueryDiagnostics`, `QueryResponse`, `SearchMode`, `MatchType` — was *already* in me-types. It was the odd one out. `me-query` re-exports it, so `me_query::MemoryQuery` still resolves. Kills the L3→L3 edge.
  - Consequence: `me-query`'s `archive` feature is dropped (its only job was gating `MemoryQuery::include_archives`, which now rides me-types' pre-existing `archive` feature — the same one already gating `QueryDiagnostics`' archive fields).

**2. `cosine_similarity` → `me-types`.** Pure math, no SQL/driver. **All five consumers and the public re-export are re-pointed** to `me_types::math`, so the facade no longer reaches a backend for math. This also unblocks **sub-PR 4** — `consolidation/{dedup,cluster}` (me-consolidate's move-set) consume it too.

**3. The `.pak` schema gate is now backend-independent.** ⚠️ **Corrected assumption:** `CURRENT_SCHEMA_VERSION = 14` is **not** a cross-backend contract — Postgres numbers its migrations independently (`CURRENT_PG_SCHEMA_VERSION = 1`, and its docs explicitly forbid "syncing" the two). Binding a **portable archive format** to one backend's migration counter is a category error that only worked while SQLite was the sole backend: a `.pak` is a zstd+JSON blob of `me-types` DTOs, so *"can this build read this pak?"* is a question about **DTO shape**.
  - New `me_types::types::archive::ARCHIVE_SCHEMA_VERSION` (L0, backend-independent). **Both** the write stamp and the read gate read that one constant → symmetric **by construction**, and the guard stays structurally unbypassable (`read_pak`'s original signature is unchanged; no parameter, no wrapper).
  - The value stays **14** deliberately: the gate rejects stamps *greater* than supported, so restarting at 1 would make every `.pak` already on disk read as "from the future" and be rejected.

## Review

Full 4-voice gate. **Codex: LGTM. agy: LGTM.** The adversarial pass found a **MEDIUM** (the cosine hoist had re-pointed *zero* call sites — the PR's own purpose was unmet) and a **latent HIGH** (the `.pak` gate's write and read halves were free to diverge; a PG engine would stamp 14, check against 1, and fail to read the pak it just wrote — with the archive fallback swallowing the error, **silently dropping every archived fact**). Fixed in `4fb6758`.

A **second** adversarial round then attacked that fix and found three more: it had traded a structural coupling for an unenforced maintainer obligation, its regression test could not detect the regression (both constants equal 14, so a re-point is value-invisible), and a stale doc still carried the wrong instruction. Fixed in `80fc051` — including `archive_pak_serde_shape_is_pinned`, a golden test that pins the serialized shape of `ArchivePak`/`Fact`/`Edge` so a DTO change **fails to compile** rather than relying on someone remembering to bump the version.

Collateral **#977** (under epic #628): the **snapshot** compat gate is this gate's identical twin and is still split across backends. Filed separately, not folded in.

## Test plan

- [x] `cargo fmt --all --check` · `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo build --workspace` (default + `--all-features`)
- [x] `cargo test --workspace --all-features` — **1970 passed / 0 failed**
      (base 1968, **+2** deliberate: `pak_write_stamp_matches_read_gate` and `archive_pak_serde_shape_is_pinned`; a test-fn-name set-diff vs base confirms nothing went dark)
- [x] `cargo test --workspace --features archive` **alone** — green (the config `--all-features` structurally masks; this is where a feature-gating regression would surface)
- [x] `cargo build -p memory-engine --no-default-features --features backend-sqlite` (archive **off**) — green
- [x] `cargo doc` (facade + me-types, `-D broken_intra_doc_links -D private_intra_doc_links`)
- [x] DAG: `me-types` has **zero `me-*` dependencies** — still a true L0 leaf

## Public API

**No break.** `memory_engine::search::MemoryQuery`, `memory_engine::search::cosine_similarity`, and `read_pak`'s signature all resolve exactly as at base. One *additive* path appears (`memory_engine::types::search::MemoryQuery`), consistent with its sibling DTOs, which already sit at both paths.

Refs #940

### DIFF
--- a/memory-engine/lib/me-backend-sqlite/src/search/vector.rs
+++ b/memory-engine/lib/me-backend-sqlite/src/search/vector.rs
@@ -49,29 +49,14 @@ static BRUTE_FORCE_WARN_ONCE: std::sync::Once = std::sync::Once::new();
 
 /// Compute the cosine similarity between two vectors.
 ///
-/// Returns 0.0 if either vector has zero magnitude (avoids NaN).
-///
-/// # Length mismatch
-///
-/// When `a.len() != b.len()` the similarity is computed over the first
-/// `min(a.len(), b.len())` components — the `zip` silently truncates to the
-/// shorter slice, so any trailing components of the longer slice are ignored.
-/// This function performs no length check; the **caller is responsible for
-/// passing equal-length vectors**. In-engine callers already enforce dimension
-/// uniformity at the store/query layer (e.g. `vector_search_filtered` rejects
-/// a wrongly-sized query, and stored embeddings are validated against
-/// `embed_dim`), so the truncation never fires on the real read path.
-#[must_use]
-pub fn cosine_similarity(a: &[f32], b: &[f32]) -> f32 {
-    let (mut dot, mut norm_a, mut norm_b) = (0.0_f32, 0.0_f32, 0.0_f32);
-    for (&x, &y) in a.iter().zip(b.iter()) {
-        dot = x.mul_add(y, dot);
-        norm_a = x.mul_add(x, norm_a);
-        norm_b = y.mul_add(y, norm_b);
-    }
-    let denom = norm_a.sqrt() * norm_b.sqrt();
-    if denom == 0.0 { 0.0 } else { dot / denom }
-}
+/// Relocated to [`me_types::math::cosine_similarity`] (Wave 2 #816 / S4,
+/// sub-PR 3a) — it is pure math with no SQL/driver dependency, so keeping it
+/// here would have forced the forthcoming `me-archive` crate (an L3 leaf) to
+/// depend on this concrete backend crate just to reuse a scoring function.
+/// Re-exported so every existing `me_backend_sqlite::search::vector::cosine_similarity`
+/// / `crate::search::cosine_similarity` call site in this crate keeps resolving
+/// unchanged.
+pub use me_types::math::cosine_similarity;
 
 /// Brute-force vector similarity search over active facts.
 ///
@@ -225,77 +210,6 @@ mod tests {
     }
 
     #[test]
-    fn cosine_identical_vectors() {
-        let a = [1.0_f32, 0.0];
-        let b = [1.0_f32, 0.0];
-        let sim = cosine_similarity(&a, &b);
-        assert!((sim - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn cosine_orthogonal_vectors() {
-        let a = [1.0_f32, 0.0];
-        let b = [0.0_f32, 1.0];
-        let sim = cosine_similarity(&a, &b);
-        assert!(sim.abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn cosine_opposite_vectors() {
-        let a = [1.0_f32, 0.0];
-        let b = [-1.0_f32, 0.0];
-        let sim = cosine_similarity(&a, &b);
-        assert!((sim - (-1.0)).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn cosine_mismatched_length_truncates_to_shorter() {
-        // Documented behavior (#479): `zip` truncates to the shorter slice, so
-        // the trailing component of `a` (9.9) is ignored and the result equals
-        // the equal-length computation over the common prefix.
-        let truncated = cosine_similarity(&[1.0_f32, 0.0, 9.9], &[1.0_f32, 0.0]);
-        let control = cosine_similarity(&[1.0_f32, 0.0], &[1.0_f32, 0.0]);
-        assert!(
-            (truncated - control).abs() < f32::EPSILON,
-            "trailing component of the longer slice must be ignored: \
-             truncated={truncated}, control={control}"
-        );
-        // The common prefix [1,0] vs [1,0] is identical, so the score is 1.0.
-        assert!((truncated - 1.0).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn cosine_mismatched_length_symmetric_truncation() {
-        // Truncation is independent of which argument is longer: swapping the
-        // operands yields the same prefix-only score.
-        let a_longer = cosine_similarity(&[1.0_f32, 2.0, 7.0], &[1.0_f32, 2.0]);
-        let b_longer = cosine_similarity(&[1.0_f32, 2.0], &[1.0_f32, 2.0, 7.0]);
-        let control = cosine_similarity(&[1.0_f32, 2.0], &[1.0_f32, 2.0]);
-        assert!((a_longer - control).abs() < f32::EPSILON);
-        assert!((b_longer - control).abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn cosine_zero_vector_returns_zero() {
-        let a = [0.0_f32, 0.0];
-        let b = [1.0_f32, 0.0];
-        let sim = cosine_similarity(&a, &b);
-        assert!(sim.abs() < f32::EPSILON);
-        // Also both zero
-        let sim2 = cosine_similarity(&a, &a);
-        assert!(sim2.abs() < f32::EPSILON);
-    }
-
-    #[test]
-    fn cosine_empty_slices_return_zero() {
-        // Empty zip -> norm 0 -> denom 0 -> returns 0.0 (NaN-avoidance promise).
-        // Use abs-epsilon (not assert_eq!) to satisfy clippy::float_cmp under the
-        // CI `-D warnings` gate, matching the sibling zero-vector test convention.
-        assert!(cosine_similarity(&[], &[]).abs() < f32::EPSILON);
-        assert!(cosine_similarity(&[], &[1.0]).abs() < f32::EPSILON);
-    }
-
-    #[test]
     fn vector_search_rejects_wrong_query_dimension() {
         let conn = setup();
         let wrong_dim_query = [1.0_f32, 0.0]; // DIM is 4, query is 2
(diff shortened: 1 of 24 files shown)


dutiona:
[Codex Review] (gpt-5.6-sol, xhigh — full 4-voice gate; quota restored this round)

Reviewed the vocabulary hoist against the six focus areas (relocation faithfulness, feature-gating drift, public API, parameterization correctness, DAG/layering, anti-#903).

**LGTM** — no findings.

_Orchestrator note, recorded for the record: this verdict was rendered against commit `e938a8f`. The adversarial in-house pass subsequently found a MEDIUM and a latent HIGH at that same commit (see the `[Code Reviewer]` comment below), both since fixed in `4fb6758`. Logged transparently — a clean external verdict is evidence, not proof._

dutiona:
[Antigravity Review] (Gemini 3.1 Pro (High))

Reviewed all six focus areas. All findings LOW; **LGTM**.

- **(A) Relocation faithfulness** — `MemoryQuery` and `cosine_similarity` bodies moved intact; tests preserved.
- **(B) Feature-gating** — `me-query` dropped its `archive` feature; since `me-query` contains no code reading `include_archives` (only the facade does, behind its own `archive` gate), behavior does not differ in any configuration.
- **(C) Public API** — `memory_engine::search::MemoryQuery`, `memory_engine::search::cosine_similarity`, and `read_pak`'s signature all resolve exactly as at base.
- **(D) Parameterization** — the compat check correctly uses the new `supported_schema_version` parameter, and every caller injects `CURRENT_SCHEMA_VERSION`. *"The parameterization allows the forthcoming Postgres backend to pass its own schema version rather than being bound to the SQLite constant."*
- **(E) DAG** — `me-types` retains zero `me-*` dependencies; a true L0 leaf.
- **(F) anti-#903** — exactly 1968; nothing added or dropped.

_Orchestrator note: finding (D) above is worth preserving verbatim, because it is **wrong in an instructive way**. It endorsed the parameterization on exactly the reasoning that made it dangerous — a PG engine passing "its own" version (1) would fail to read paks stamped by the still-hard-wired write side (14), silently dropping every archived fact. The review faithfully validated the design **as documented**; the flaw was in the design and its doc, which this review had no reason to distrust. Fixed in `4fb6758`. A faithfulness lens cannot catch a guard whose two halves are free to diverge — only a structural lens asks that question._

dutiona:
[Code Reviewer] (in-house adversarial pass — Opus)

Two real findings at `e938a8f`, both **verified against source** and both **fixed in `4fb6758`**. Codex and agy each returned LGTM on this same code.

---

**[MEDIUM] The `cosine_similarity` hoist did not cut the edge it claimed to cut — zero call sites were re-pointed.**

The function moved to `me-types`, but all five consumers still resolved it through `crate::search::vector` → `pub(crate) use me_backend_sqlite::search::{strategy, vector}`:
`archive/search.rs:16`, `consolidation/dedup.rs:10`, `consolidation/cluster.rs:8`, `engine/dormant.rs:2`, `engine/cycle/dbscan.rs:11`.

So `archive/search.rs` — **the file sub-PR 3b carves into `me-archive`** — still depended on the concrete SQLite backend, and the PR's stated purpose ("the prerequisite that makes the me-archive carve legal") was **not met**. The same trap awaited `me-consolidate` (sub-PR 4) via `cluster.rs`/`dedup.rs`.

**Fixed:** all five re-pointed to `me_types::math::cosine_similarity`; the facade's public re-export now sources L0 directly and the `vector` import is gone. The facade no longer reaches a backend for math.

---

**[HIGH — latent, would fire at #634] The `.pak` schema gate was split into two halves free to diverge.**

The **read** check was parameterized (`read_pak(path, supported_schema_version)`), but the **write** stamp at `engine/archive.rs:203` stayed hard-wired to `me-backend-sqlite`'s `CURRENT_SCHEMA_VERSION` (= 14). And the new doc on `read_pak` explicitly instructed the #634 implementer that *"a Postgres-backed engine passes its own schema version here."*

Postgres numbers its migrations **independently** — `CURRENT_PG_SCHEMA_VERSION = 1` (`me-backend-postgres/src/migrations.rs:35`), whose own docs state: *"Version numbering — logical-equivalence, NOT numeric-equality … Do NOT 'sync' this number to SQLite's."*

Failing scenario, following the PR's own doc:
1. PG-backed engine archives → `.pak` stamped `engine_schema_version = 14` (write side, still the SQLite constant).
2. Same engine queries with `include_archives` → `search_archives_fallback` passes `CURRENT_PG_SCHEMA_VERSION = 1`.
3. `read_pak` → `14 > 1` → `SchemaVersionUnsupported`.
4. `engine/query.rs` **swallows it** (best-effort `tracing::warn!`) → **every archived fact silently vanishes from every archive query.** A PG engine cannot read the pak it just wrote.

**Root cause is a category error, not the asymmetry.** A `.pak` is a portable zstd+JSON blob of `me-types` DTOs, so "can this build read this pak?" is a question about **DTO shape**, not any backend's migration history. Binding it to a backend counter only worked while SQLite was the sole backend.

**Fixed** by introducing `me_types::types::archive::ARCHIVE_SCHEMA_VERSION` — the backend-independent logical content-schema version of the `.pak` format (L0). **Both** the write stamp and the read gate read that one constant → symmetric **by construction**. This made the parameterization unnecessary: `read_pak`'s original signature is restored and the facade wrapper dropped, returning the guard to **structurally unbypassable** rather than a caller obligation (which was itself the degradation flagged by the completeness sweep).

The value stays **14** deliberately: existing `.pak`s stamp 14 and the gate rejects stamps *greater* than supported, so restarting at 1 would make every archive already on disk read as "from the future" and be rejected — the fix would have shipped its own data-loss bug.

New test `pak_write_stamp_matches_read_gate` pins the invariant: **what the write side stamps is exactly what the read side accepts.**

---

**Also verified clean:** relocation faithfulness (both bodies byte-identical to base; all 42 `MemoryQuery` tests + all 6 unit tests / 4 proptests for `cosine_similarity` moved intact; the doctest survives and runs), public API (no break; one *additive* path `memory_engine::types::search::MemoryQuery`), DAG (`me-types` = zero `me-*` deps), anti-#903 (**1969** after the fix — 1968 + the one new symmetry test), and the `#117` scope-miss guard from sub-PR 2 is untouched and still green.

dutiona:
[Code Reviewer] — **round 2: adversarial re-review of the fix `4fb6758`** (in-house, Opus)

A substantial fix deserves fresh eyes, so `4fb6758` was re-reviewed as an attack target rather than confirmed. It found **three problems with the fix itself**. All fixed in `80fc051`.

**Verified sound first:** `ARCHIVE_SCHEMA_VERSION = 14` is correct (the gate is a one-sided *upper* bound — `stamp > supported → reject` — so older paks stamped `< 14` still read; restarting at 1 would reject every 14-stamped pak on disk). Write/read symmetry is complete for the pak path (whole-workspace grep: the only production write is `build_pak`, the only production read is `read_pak`; both now source the L0 constant; `CURRENT_SCHEMA_VERSION`'s legitimate SQLite-migration uses are untouched). The cosine re-point is complete and behavior-inert (`grep "search::vector"` across the facade → zero hits; the backend's internal chain intact for `ann.rs`). Public surface is at base parity — `read_pak`'s signature is byte-identical to base again. anti-#903 verified by an independent run **and** a test-fn-name set-diff vs base: exactly one added, zero removed.

---

**[WARNING] The fix swapped a structural coupling for an unenforced human obligation.**

Before: the stamp inherited `CURRENT_SCHEMA_VERSION`, so *any* store migration moved it — over-approximating, but **automatic**. After: `ARCHIVE_SCHEMA_VERSION` is frozen at 14 and moves **only if a maintainer bumps it** when the `.pak`'s DTOs change.

Nothing enforced that, and the failure is silent: `Fact` has no `deny_unknown_fields` and several `#[serde(default)]` fields (`me-types/src/types/facts.rs:141,143,155,157`). A build that adds a `Fact` field writes paks still stamped 14; an older build accepts them (`14 <= 14`) and **silently ignores the new field** — where the old design would have produced a clean `SchemaVersionUnsupported`. Decoupling the archive from a backend counter was right; losing the automatic bump was not.

**Fixed:** `archive_pak_serde_shape_is_pinned` pins the serialized key sets of `ArchivePak`/`Fact`/`Edge`. Its fixture enumerates every field explicitly (no `..Default::default()`), so adding a DTO field **fails to compile** — the shape question cannot be skipped. Plus a back-reference from `CURRENT_SCHEMA_VERSION`'s docs so the obligation is visible where someone would trip it. (The test caught `Fact::surfaced_at` — a field I did not know about, and itself a `#[serde(default)] Option`, i.e. a live specimen of the hazard.)

**[WARNING] `pak_write_stamp_matches_read_gate` cannot catch the bug it was written for — today.**

Both assertions compare *values*, and `ARCHIVE_SCHEMA_VERSION` == `CURRENT_SCHEMA_VERSION` == **14**. So re-pointing the write stamp back at the backend constant is **value-invisible** and the test still passes. It becomes a live tripwire only once the two diverge (SQLite v15). It is a **delayed** guard, not an immediate one — now documented as such, rather than left to be over-trusted. The genuinely structural guard lands in **sub-PR 3b**: once `build_pak` moves into `me-archive` (a crate with no `me-backend-sqlite` dependency), the backend constant becomes *unnameable* from either half.

**[WARNING] Stale doc still carried the wrong instruction.**

`archive/search.rs` still documented the deleted `supported_schema_version` param, telling the reader to supply *"the active backend's current schema version"* — **the exact instruction that caused the HIGH** — and it sits in the file sub-PR 3b `git-mv`s into `me-archive` verbatim. No gate catches prose. Rewritten, plus two further doc corrections (`engine_schema_version`'s field doc, which post-fix describes precisely what it is *not*; and `search/mod.rs`'s carve map, now listing `vector` among the items no longer re-exported).

---

**[COLLATERAL → #977]** The **snapshot** compat gate is this gate's identical twin and is *still* split: write stamps the live DB's `schema_version` (runtime, backend-numbered — a PG engine writes 1), read compares against SQLite's compile-time `CURRENT_SCHEMA_VERSION` (14). Same category error, reachable at #634. Lower severity (restore fails loudly rather than swallowing), so filed separately under epic #628 rather than folded in.

Issue #981, opened 2026-07-12
refactor(cognitive): carve me-cognitive — requires inverting the public DreamContext into a me-traits capability trait


dutiona:
## Summary

The Phase-5 **cognitive/dream layer** (`engine/cognitive.rs` + `engine/cycle/*`, ~3,100 LOC) is the one part of the engine that Wave 2's decomposition leaves in the facade. Carving it is blocked on a **public API break** that was deliberately scoped out of S4 sub-PR 4 (#940), and it deserves its own design pass rather than being bolted onto the last slice of an in-flight epic.

## Why it can't just be carved

`DreamContext` is **facade-bound by construction**:

```rust
pub struct DreamContext<'a> {
    engine: &'a MemoryEngine,          // <- a reference to the L4 facade
}
```
and its methods delegate straight through (`self.engine.query(...)`, `self.engine.list_active_facts(...)`). `CycleContext` then *wraps* `DreamContext` (`engine/cycle/context.rs:27`).

So moving `cognitive.rs`/`cycle/*` into any L3 crate would force that crate to **name `MemoryEngine`** — an **L3 → L4 back-edge to the facade**. That is not merely an illegal sibling edge (like the L3→L3 that #976 had to fix); it is a cycle straight back to the top of the DAG.

`DreamContext` is also **public API** (`lib.rs:295`: `pub use engine::cognitive::{DreamContext, INSIGHT_MARKER_KEY}`), and consumers write `DreamCycle`/`DeltaProposer` implementations against it.

## The fix has a precedent — S1 already did it once

This is the *same shape* as the S1 keystone: `DreamCycle::run(&CycleContext)` became `run(&dyn CycleCtx)`, where `CycleCtx` is a `me-traits` **capability trait** enumerating the exact read-set the shipped implementations use, and `CycleContext` *implements* it. (ADR-0018 records that as gated public-API break #1.)

The same inversion applies here: **`DreamContext` → a `me-traits` capability trait** (`DreamCtx`?), which the facade implements. Then a `me-cognitive` crate can name the trait instead of the facade.

## Scope of work

1. Enumerate `DreamContext`'s **full live call set** — the S1 lesson was that the trait must cover every method the shipped impls actually use, or the inversion has to be redone. Grep `DefaultDreamCycle`, `LlmDreamCycle`, `HttpDeltaProposer`, and the MCP/CLI paths.
2. Define the capability trait in `me-traits`; have the facade's `DreamContext` implement it.
[...]

Issue #972, opened 2026-07-11
enhancement(core): fail-fast read-only gate (ctx.ensure_writable) for write primitives on MemoryCtx


dutiona:
## Context

Surfaced by the PR #971 (`me-resolve`) super-review — a valid gemini observation, deferred to keep #971 a behavior-preserving carve (maintainer decision 2026-07-11).

`me_resolve::resolve_conflict` (and, as they adopt `MemoryCtx`, the other **write** primitives) currently rely on the **downstream** read-only guard: `resolve_conflict_atomic` → `block_write` → `ConnectionPool::try_write()` (`me-backend-sqlite/src/sqlite/mod.rs:440`) fails with `MemoryError::ReadOnly`. So a read-only engine executes `get_fact` **and the consumer-provided `ConflictArbiter::arbitrate`** (which may perform expensive work — network / LLM calls) **before** failing.

`MemoryCtx::ensure_writable()` (`me-storage/src/ctx.rs:59`) exists *specifically* for this — "the write gate; write primitives call this before mutating."

## Task

Add `ctx.ensure_writable()?` at the entry of each **write** primitive, right after `ctx.ensure_open()?`, so a read-only engine **fails fast** before any read or consumer-trait call. Apply uniformly (not piecemeal) as each primitive lands on `MemoryCtx`:

- [ ] `me_resolve::resolve_conflict` (on `MemoryCtx` now, #971)
- [ ] `me_forget::prune` (adopts `MemoryCtx` via #970)
- [ ] `me-ingest` write ops (S4)
- [ ] `me-consolidate` write ops (S4)

## Behavior note

This is a **deliberate, benign behavior change**: the observable error stays `MemoryError::ReadOnly`, but on a read-only engine the arbiter/consumer side-effects (logging/network/LLM) that currently fire **before** the failure will no longer fire. That is why it was kept out of the #971 faithful extraction and tracked here for a coordinated sweep.

Refs #816. Best folded into #970 + the S4 extractions, or done as one small sweep once the write primitives are all on `MemoryCtx`.

Issue #742, opened 2026-06-25
feat(core): different-dimension reconstruction (engine effective-dim transition)


dutiona:
> **Reviewed implementation plan** (3-lens max-effort panel → synthesis → clean-slate adversarial review; 1 HIGH + 2 MEDIUM folded). Builds on #623 (merged, `1f235aa`). Ready to implement on a fresh worktree off `main`.

# Plan — #742: different-dimension background reconstruction (reopen-at-D′)

**Issue:** #742 `feat(core): different-dimension reconstruction (engine effective-dim transition)`
**Type/Area:** `type:feature` / `area:core` (+ `area:storage`) · **Epic:** #610 Wave 2 · builds on #623 (merged, `1f235aa`)
**Engagement:** Thorough · **Severity:** Default
**Mechanism (chosen from a 3-lens panel):** **reopen-at-D′** — accept a brief reopen between promote-commit and serving at the new dimension. True no-downtime in-place transition is a **deferred epic** (priced below).

---

## BLUF

#623 shipped same-dim reconstruction; its storage layer is **already dimension-agnostic** (the promote copy-swaps blobs; `fact_vectors` has no dim column; `PromoteOutcome.new_fingerprint.dim` already carries the target dim — designed in for #742). The _only_ thing freezing the engine to one dimension is `embed_dim`, consumer-passed at open and cached immutably in `MemoryEngine` / `SqliteBackend` / `ConnectionPool` / `HnswStrategy` + every `deserialize_embedding(blob, embed_dim)` read site. After a D→D′ promote, `facts.embedding` is D′-wide but the live engine still holds D.

**Plan:** let D′ through the guards (targeting the dim checks at D′), and at the moment the promote commits, **fence** the now-stale engine — every embedding-touching read/write returns a typed `EmbeddingReopenRequired { new_dim }` until the consumer drops the handle and **reopens at D′** (which re-validates clean — `embedding_meta` now says D′ — and rebuilds HNSW@D′ via `build_from_db` _for free_). No engine-side `embed_dim` mutation. The crucial safety fact (risk-first): `deserialize_embedding` is a **checked length compare** (`store/mod.rs:48`), so a stale-D reader on a D′ blob returns `Err(EmbeddingDimension)` — **loud and recoverable, never UB / silent wrong-space**. The fence upgrades that loud-but-cryptic error into an actionable one.

[...]

PR #982, opened 2026-07-12
refactor(build): carve the me-consolidate crate

### LINKED ISSUE #940: refactor(core): S4 — carve MODERATE primitives me-query/me-ingest/me-consolidate/me-archive (Wave 2 #816)

dutiona:
## What

**S4 sub-PR 4 of 5 — the final one.** Carves the **`me-consolidate`** crate (the Consolidate primitive) out of the facade. Workspace **16 → 17 crates**. No behavior change, no public-API change.

**This completes slice S4.** `Closes #940`.

## Scope — the Consolidate *primitive*, deliberately not the cognitive layer

`me-consolidate` = **Consolidate**, one of the five documented primitives (Ingest → Query → **Consolidate** → Forget → Resolve).

The Phase-5 **cognitive/dream layer** (`engine/cognitive.rs` + `engine/cycle/*`, ~3,100 LOC) **stays in the facade**, by maintainer decision. It is facade-bound *by construction*:

```rust
pub struct DreamContext<'a> { engine: &'a MemoryEngine }   // <- a reference to the L4 facade
```
and `CycleContext` wraps it. Moving it into an L3 crate would force that crate to **name `MemoryEngine`** — an **L3 → L4 back-edge to the facade**, i.e. a cycle straight back to the top of the DAG. Carving it properly needs the same inversion S1 performed for `DreamCycle::run` (`&CycleContext` → `&dyn CycleCtx`, a `me-traits` capability trait) — but `DreamContext` is **public API**, making that a 4th deliberate API break. Deferred with its full design analysis to **#981** rather than bolted onto the epic's final slice. ADR-0018's L3 list has no `me-cognitive`; the cognitive layer *consumes* primitives, it isn't one.

## The carve

Triage against current main found the primitive was **already carve-ready**, and that the plan was stale:

- `consolidation/{mod,dedup,cluster,global}.rs` are **production-pure**. Every backend import (`rusqlite::Connection`, `FactStore`/`SummaryStore`, `get_config`/`set_config`, the `me_backend_sqlite::consolidation::apply_*` re-exports) is **`#[cfg(test)]`-gated** — S2 sub-PR 2b moved the SQL into the backend, and #976 cut the `cosine_similarity` edge. The comment at `mod.rs:36` says it outright: *"`Connection`/`get_config`/`set_config` are production-unused here."*
- There is **no me-query callback** — S1's `CycleCtx` inversion already removed that need (`CycleCtx`'s method set is `time_window`/`prior_wisdom`/`prior_reports`/`list_undreamt_in_period`/`outcome_counts_batch`; no `query`).

So: the four files move wholesale, **tests included**. The `#[cfg(test)]` integration wrappers (`consolidate`, `consolidate_with_caps`, `local_dedup`, `cluster_fusion`, `global_integration` — which compose the pure `compute_*` with the backend's `apply_*` over a raw `Connection`) travel with their files, backed by a **`me-backend-sqlite` dev-dependency**. A dev-dep is **not** a production edge, and this mirrors `me-query`'s existing dev-dependency on the same crate.

`engine/consolidation.rs`'s single `consolidate` method becomes a free fn over `MemoryCtx` (the me-ingest/me-query/me-archive template); the facade keeps a thin delegate with its rustdoc byte-for-byte:

```rust
pub async fn consolidate(
    ctx: MemoryCtx<'_>, graph: &RwLock<MemoryGraph>,
    generator: Arc<dyn SummaryGenerator>, embedder: Arc<dyn EmbeddingProvider>,
    config: &ConsolidationConfig,
) -> Result<ConsolidationStats>
```

## Completeness sweep (the question that found a real bug on three consecutive PRs)

> *Did any guard enforced **structurally** — by code position, an upstream early-return, or a feature flag — get relocated across the new crate boundary such that it is no longer enforced?*

**No.** Both named risks checked explicitly:

1. **`ensure_open` / read-only.** Base's `MemoryEngine::consolidate` called **only** `self.ensure_open()` — it never had a read-only pre-check; `ReadOnly` surfaces later, when `apply_plan`'s write hits the pool's `try_write()` guard. The free fn calls `ctx.ensure_open()?` in the **identical first position** and deliberately does **not** add `ctx.ensure_writable()`: that would fail *before the read phase* instead of during the write phase — a real behavior change wearing a safety costume. (Note this is the **opposite** verdict from #978's me-archive carve, and correctly so: *there*, base **did** have an explicit read-only pre-check, so the facade kept it. A faithful carve is not free to be safer. The systemic `ensure_writable` sweep is **#972**, where that change can be made deliberately across all primitives at once.)
2. **`#[cfg(test)]` crossing a crate boundary.** Not the known trap: the items the moved tests reach (`apply_plan`, `apply_dedup`, `apply_clusters`, `apply_global`, `load_snapshot*`) are unconditionally `pub` in `me-backend-sqlite`; only the *local re-export binding* inside me-consolidate is `cfg(test)`-gated, and that cfg is this crate's own. Verified empirically both ways: `cargo build -p me-consolidate` compiles with those names absent from the graph entirely; `cargo test -p me-consolidate` is green.

## Test plan

- [x] `cargo fmt --all --check` · `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo build --workspace` (default + `--all-features`)
- [x] `cargo test --workspace --all-features` — **1970 passed / 0 failed** (anti-#903: pure carve, tests relocate, count unchanged)
- [x] `cargo check -p memory-engine` (default) + `--no-default-features --features backend-sqlite` (the facade-alone gates added in #978)
- [x] `cargo doc` (facade + me-consolidate, `-D broken_intra_doc_links -D private_intra_doc_links`)
- [x] **DAG gate 1:** `cargo tree -p me-consolidate -i memory-engine` → did-not-match (zero back-deps on the facade)
- [x] **DAG gate 2:** `cargo tree -p me-consolidate --edges normal` → **no `me-backend-*`** (production deps are `me-types`/`me-traits`/`me-storage`/`me-index` + chrono/parking_lot/tokio/tracing; the backend dev-dep is test-only and invisible here)

## Public API — unchanged

The only facade line removed is `pub(crate) mod consolidation;` (never public). `MemoryEngine::consolidate`'s signature and rustdoc are preserved byte-for-byte; only its body became a one-line delegate.

Closes #940


dutiona:
[Code Reviewer] (in-house adversarial pass — Opus)

**No blocker.** The carve is behaviourally sound and every headline claim held under independent verification. Three warnings, **all in the documentation/verification layer** — fixed in `37c8855`.

**Claims I verified myself (not the PR's word):**

| claim | verdict | evidence |
|---|---|---|
| anti-#903 = 1970, net-zero | ✅ | My runs: base **1970/0**, HEAD **1970/0**. Cross-check: workspace-wide `#[test]`-fn **name** set-diff = **1908 vs 1908, zero added, zero removed**. |
| Extraction faithful | ✅ | Normalized code-only diff (comments/imports stripped, paths remapped) of all 4 moved files: **only rustfmt reflow**. 48 tests, 2 proptests, 140 asserts — identical both sides. |
| **No feature-unification regression** (the #978 hazard) | ✅ | Resolved-feature-set diff (`cargo tree --workspace -e features`), base vs HEAD, **default AND `--all-features`**: the only delta is the new `me-consolidate` row *with an empty feature set*. The fragile facade-alone archive-OFF gate is unaffected. |
| No production edge to a backend | ✅ | **Compiler-enforced**, not convention: dev-deps aren't linked into the lib target, so un-`cfg(test)`-ing any `me_backend_sqlite` import fails `cargo build -p me-consolidate` outright. |
| No dark tests | ✅ | The 48 run **both** standalone and inside the `--workspace --all-features` gate. |

**On the guard-order claim (B.i) — it is TRUE.** `MemoryCtx::ensure_open` is byte-equivalent to `MemoryEngine::ensure_open`, and `mem_ctx()` is pure — so for open / closed / read-only / #742-fenced, order and error are identical to base. "ReadOnly is caught below the seam" is real, not hand-waved: `apply_plan` → `block_write` → `try_write()` → `MemoryError::ReadOnly`, pinned by `block_write_on_read_only_pool_yields_read_only`.

---

**[WARNING] Two files in this PR asserted opposite things about where a guard lives.** ✅ fixed

me-consolidate's comment said *"The facade delegate performs the identical check first"* — but the delegate is a bare one-line forward, and `engine/consolidation.rs`'s **own module doc** says so (*"the delegate stays a one-line forward with no pre-flight of its own"*). The sentence was copy-pasted from `me-archive`, where it **is** true (that delegate keeps a pre-flight, because its file-backed check must run before `archive_dir` can be resolved). No runtime impact — but it encodes a "belt-and-braces" model for the **single** guard this carve moved across a crate boundary: exactly the artefact that had to be corrected on sub-PR 2 after the scope-miss bug. *(Codex and agy independently flagged this too — three-way convergence.)*

**[WARNING] That guard had zero test coverage.** ✅ fixed

Deleting `ctx.ensure_open()?` left **all 1970 tests green**. The two `force_reopen_fence` tests assert only on `add_fact`; `read_only_test.rs` never calls `consolidate`; and me-consolidate's own 48 tests route through a `#[cfg(test)]` wrapper that never constructs a `MemoryCtx`, so they *structurally cannot* reach the guard. A fenced handle would have consolidated against the **stale `embed_dim`**. Faithful to base (base's was equally untested) — but a carve is precisely the event that makes an untested guard evaporate. Now asserted in the existing fence enumeration, at **no anti-#903 cost**.

**[WARNING] The canonical design records would have shipped wrong — and this PR closes the slice.** ✅ fixed

ADR-0018 justified the *already-shipped* S1 API break as *"necessary so `me-traits` does not depend on **me-consolidate** … **me-consolidate's** `CycleContext` implements it."* `CycleContext` never landed there — it is in the facade. `crate-layout.md` still routed `engine::cycle`/`engine::cognitive` **into** me-consolidate: not merely lagging, but wrong about the destination. The me-query PR *did* amend ADR-0018 when it diverged; this one had to as well.

---

**Also filed, not folded:** **#983** (the 50k caps duplicated across me-consolidate and me-backend-sqlite with **no agreement test** — if the backend's cap drifts below the pipeline's, `load_snapshot` marks `over_both_caps` while `compute_plan_capped` disagrees: a silent skip/no-skip mismatch) · **#984** (`spawn_join_err` now has four private copies; wants one home before S5 makes it five).

dutiona:
[Codex Review] (gpt-5.6-sol, xhigh)

**[LOW]** — ✅ **fixed in `37c8855`.**

> The guard rationale says the facade delegate "performs the identical check first," but the delegate is only a direct call to `me_consolidate::consolidate` and has no pre-flight check (`engine/consolidation.rs:56`); the sole `ensure_open` call is in the extracted function (`me-consolidate/src/lib.rs:107`). This contradicts the delegate module documentation and makes the stated check-order proof internally inconsistent. Correct the comment at `me-consolidate/src/lib.rs:93-95` to say that the base check moved into the free function; the code behavior itself remains faithful.

Correct on every point, including that the **behaviour** is faithful — the defect was purely in the rationale. The comment had been copy-pasted from `me-archive`, whose delegate genuinely *does* keep a pre-flight (its file-backed check must run before `archive_dir` can be resolved). Now documented as the sole enforcement point: drop that line and the fence is gone.

Independently found by all three reviewers this round (Codex, agy, and the adversarial pass) — the strongest corroboration any finding has had in this epic.

dutiona:
[Antigravity Review] (Gemini 3.1 Pro (High)) — **LGTM**, 1 LOW.

**[LOW] Completeness sweep (B.i)** — ✅ **fixed in `37c8855`.**

> The comment in `me-consolidate/src/lib.rs` states "…The facade delegate performs the identical check first…". However, `engine/consolidation.rs` shows `self.ensure_open()?` was removed from the facade delegate, leaving the `ensure_open` check entirely to the new primitive. While the execution behavior and error order remain identical (because the primitive performs the check first-thing, correctly omitting `ensure_writable`), the comment is factually incorrect.

Right on both halves — the behaviour is identical, and the comment was wrong. Same finding as Codex and the adversarial pass; three-way convergence.

One clarification on the parenthetical *"the intended defense-in-depth layer at the facade was lost"*: there was no such layer at base. Base's `MemoryEngine::consolidate` had a **single** `self.ensure_open()`, which the carve **moved** (not removed) into the primitive. The mistake was the comment claiming a second, belt-and-braces check that never existed — not the loss of one. (Contrast `me-archive`, where the facade delegate genuinely does retain a pre-flight, and where this sentence was copy-pasted from.)

**Verified passing:** extraction faithfulness (logic and tests verbatim), crate separation (`cfg(test)` intact; `proptest` + `me-backend-sqlite` correctly `[dev-dependencies]`), DAG, public API, anti-#903.

Issue #941, opened 2026-07-04
refactor(core): S5 — facade shrink (Wave 2 #816)


dutiona:
S5 of the Wave-2 decomposition (plan #925). Shrink memory-engine (L4) to a thin re-export + orchestration layer once the L3 primitives are carved. Public API stays stable via re-exports.

dutiona:
## ⚠️ Hard constraint on the freeze ratchet — discovered in S5-1 (PR #986)

**`cargo public-api` cannot implement this freeze.** Do not build the ratchet on it. This is not a preference; it was established empirically, twice, during #986's super-review.

### What happened

#986 (hoist `ForgetPolicy` L3 → L0) claimed *"public API unchanged"*, evidenced by `cargo public-api` reporting **4018 → 4018 items** with no names added or removed.

**The claim was false.** The facade blanket-re-exports `me_types::types` (`memory-engine/src/lib.rs:122`), and `me_types::types::forgetting` is a `pub mod`. Hoisting `ForgetPolicy` into that module made a **new** path writable by consumers:

```
memory_engine::types::forgetting::ForgetPolicy      <-- NEW (module held only PruneStats at base)
```

Proven with an out-of-tree compile probe: a consumer crate naming that path builds at HEAD (**exit 0**) and fails at base `f1a308b` (**exit 101, E0433: cannot find ForgetPolicy in forgetting**). A `PruneStats` control compiled on *both* refs — so the module was already public; only `ForgetPolicy` became reachable.

### Why the tool is blind — this is structural, not a bug

`cargo public-api` canonicalizes a type to a **single definition site** and enumerates **items**. It cannot observe a *path* becoming writable. Measured directly on this repo:

```
$ grep -c "memory_engine::types::" papi_base.txt
0
```

The entire `memory_engine::types::*` surface renders as **one opaque line** — `pub use memory_engine::types`. The tool does not see into the re-exported module **at all**.

### The double failure mode

Had #941 installed a `cargo public-api` byte-diff as the ratchet, on this very PR it would have:

- ❌ **FAILED** the PR over **4 cosmetic lines** — the definition site re-rendering as `me_forget::types::ForgetPolicy` → `me_types::types::forgetting::ForgetPolicy` inside `forget()` signatures. Every Wave-2 relocation produces these. A gate that reddens on every refactor gets disabled within a month.
- ❌ **MISSED** the one **real** API addition entirely.

[...]

PR #986, opened 2026-07-12
refactor(build): hoist ForgetPolicy to me-types (L0) — rejoin PruneStats, unblock DreamCtx

### LINKED ISSUE #985: refactor(build): hoist ForgetPolicy to me-types (L0) — rejoin PruneStats, unblock DreamCtx
## Context

Wave 2 (#816), slice **S5**, sub-PR 1 of 3 — a **prerequisite** for #981 (`DreamCtx` / `me-cognitive`).

`ForgetPolicy` and `PruneStats` are the **input and output of the same operation** (`forget`), but they live **two layers apart**:

| type | crate | layer |
| --- | --- | --- |
| `ForgetPolicy` (config in) | `me-forget/src/types.rs` | **L3** |
| `PruneStats` (data out) | `me-types/src/types/forgetting.rs` | **L0** |

`PruneStats` was hoisted to `me-types` in Wave 2 (#816 E.4b Phase B). `ForgetPolicy` was left behind on a rationale that is now stale — from `me-forget/src/types.rs`:

> *"Lives here (rather than in `traits.rs`) because it isn't a trait — `traits.rs` declares the consumer behaviour contracts, while this is the concrete forgetting-layer type it configures."*

That answers **"traits.rs or forgetting?"** — a pre-Wave-2 question. Wave 2 changed the question to **"which layer?"**, and a peer-relative rationale cannot migrate. `ForgetPolicy` is `#[derive(Debug, Clone)]` over plain scalar fields and imports exactly one thing — `me_types::error::Result`. It has **zero L3 dependencies**. It is pure config data sitting two layers too high.

## Why it blocks #981

#981 promotes the dream-cycle capability bag into a **me-traits (L0.5)** trait:

```rust
trait DreamCtx: Send + Sync {
    async fn forget(&self, policy: &ForgetPolicy) -> Result<PruneStats>;   // <-- here
    …
}
```

A trait in **L0.5** cannot name a type in **L3** — that is exactly the back-edge ADR-0018's break #1 exists to prevent. So the hoist must land first.

## Scope

- Move `ForgetPolicy` (+ its `Default` and inherent impls) `me-forget/src/types.rs` → `me-types/src/types/forgetting.rs`, rejoining `PruneStats`.
- Re-export from `me-forget`'s root so `me_forget::ForgetPolicy` still resolves; facade `crate::forgetting::ForgetPolicy` unchanged.
- **Public API unchanged** (re-export preserves every path).
- Replace the stale doc rationale with a **property-based** one ("pure config data → L0"), not a peer-relative one.

## Gates

The CI-exact matrix; **anti-#903 = exactly 1970** (a move: no test added or dropped).

Part of #816 · unblocks #981 · S5 sub-PR 1 of 3.

dutiona:
## What

Hoists **`ForgetPolicy`** from `me-forget` (**L3**) to `me-types` (**L0**), rejoining its sibling `PruneStats` in `me_types::types::forgetting`.

Wave 2 (#816), slice **S5**, sub-PR **1 of 3**. **Closes #985.** **Unblocks #981.**

## Why

`ForgetPolicy` (config **in**) and `PruneStats` (data **out**) are the input and output of the *same* operation, but they sat **two layers apart**:

| type           | was                                | layer  |
| -------------- | ---------------------------------- | ------ |
| `ForgetPolicy` | `me-forget/src/types.rs`           | **L3** |
| `PruneStats`   | `me-types/src/types/forgetting.rs` | **L0** |

`ForgetPolicy` is `#[derive(Debug, Clone)]` over scalar fields; its only non-local imports are `me_types::error::Result` and `me_types::types::FactType` — **both already L0**. It has **zero L3 dependencies**. It is pure config data sitting two layers too high.

### The stale rationale

Both sides of the split carried a rationale that no longer holds. From `me-forget/src/types.rs`:

> *"Lives here (rather than in `traits.rs`) because it isn't a trait…"*

That answers **"traits.rs or forgetting?"** — a *pre-Wave-2* question. Wave 2 changed the question to **"which layer?"**, and a **peer-relative** rationale cannot migrate. The `me-types` side was worse — it was circular:

> *"`ForgetPolicy` stays in the monolith's `forgetting/types.rs` (an L3 type; **L0 me-types cannot link up to it**)"*

"L0 cannot link up to L3" is true of *every* type; it is not an argument for `ForgetPolicy` **being** L3.

Both are replaced with a **property-based** rationale (*pure config data, no L3 deps → L0, colocated with the `PruneStats` it is paired with*). A rationale naming a **property** survives the plan changing under it; one naming a **peer** does not.

### Why #981 needs it

#981 promotes the dream-cycle capability bag into a **me-traits (L0.5)** trait:

```rust
trait DreamCtx: Send + Sync {
    async fn forget(&self, policy: &ForgetPolicy) -> Result<PruneStats>;   // <-- here
    …
}
```

An **L0.5** trait cannot name an **L3** type — exactly the back-edge ADR-0018's break #1 exists to prevent. So this lands first.

## Public API — additively extended by exactly one path (corrected during review)

> **This section originally claimed "unchanged". Codex refuted that in super-review, and it was right.**

`cargo public-api` on the facade, `main` vs HEAD, `--all-features`: **4018 items → 4018 items**, no names added or removed, with 4 lines differing only in how `ForgetPolicy`'s *definition site* renders inside signatures.

**That is not a reachable-path proof, and presenting it as one was the error.** `cargo public-api` canonicalizes a type to a single definition site — it counts **items**, so it structurally *cannot see a path becoming newly writable*.

What actually changed: the facade blanket-re-exports `me_types::types` (`lib.rs:122`), and `me_types::types::forgetting` is a `pub mod`. Hoisting `ForgetPolicy` into that module therefore makes a **new** path writable by consumers:

```
memory_engine::types::forgetting::ForgetPolicy      <-- NEW (module held only PruneStats at base)
```

**This is accepted, not contorted away:**

- It is **additive** — nothing is removed, so nothing can break.
- It **restores symmetry**. `PruneStats` already had *both* a flat (`memory_engine::PruneStats`) and a nested (`memory_engine::types::forgetting::PruneStats`) path. The L3/L0 split was the *only* reason `ForgetPolicy` had just one. Rejoining the pair rejoins their surfaces.
- The alternative — making the module private — would **remove** the pre-existing `memory_engine::types::forgetting::PruneStats`. That is a genuine break, and strictly worse than an additive path.

Every pre-existing path still resolves unchanged: `memory_engine::ForgetPolicy`, `me_forget::ForgetPolicy`, `crate::forgetting::ForgetPolicy`, and both `PruneStats` paths.

The new surface is now **pinned** in `reexports_are_accessible` (`9a4f45c`) — a compile-time probe that *names paths*, which is the only kind of check that can see this class of change.

> ### ⚠️ This constrains #941 (S5-3, the re-export freeze)
> The freeze ratchet **must compare reachable paths, not `cargo public-api` item counts.** A public-api byte-diff gate would have been blind to this very change — and, worse, it would have *failed* this PR for the 4 cosmetic definition-site lines while *missing* the one real API addition. A gate that cries wolf on refactors and sleeps through real changes is worse than no gate. `reexports_are_accessible` is the pattern to build on.

## Doc correction folded in

`docs/reference/crate-layout.md` claimed **me-traits** held *"policy types (`ForgetPolicy`, …)"*. `ForgetPolicy` has **never** been defined in `me-traits` — it lived in `me-forget`, and now lives in `me-types`. The doc was wrong *before* this PR and would be wrong *differently* after it, so it is corrected **in the PR that diverges** rather than left to rot (the recurring Wave-2 lesson).

## Verification — CI-exact, run unpiped

| gate | result |
| --- | --- |
| `cargo fmt --all --check` | ✅ 0 |
| `cargo clippy --workspace --all-targets --all-features -- -D warnings` | ✅ 0 |
| `cargo build --workspace` | ✅ 0 |
| `cargo build --workspace --all-features` | ✅ 0 |
| `cargo test --workspace --all-features` | ✅ **1970 passed / 0 failed** |
| `cargo check -p memory-engine` (facade-alone, archive-OFF gate) | ✅ 0 |
| `cargo check -p memory-engine --no-default-features --features backend-sqlite` | ✅ 0 |
| `cargo doc --no-deps` (`memory-engine`, `me-types`, `me-forget`, `--all-features`) | ✅ 0 |

**DAG gate:** `cargo tree -p me-types --edges normal` shows **no `me-forget` edge** — L0 did not gain an L3 dependency.

**anti-#903 = exactly 1970.** A pure relocation: the **18** `ForgetPolicy` unit tests move from `me-forget`'s test target to `me-types`'s, verified **by name** (all 18 `types::forgetting::tests::*` present), not merely by total. `me-forget`'s unittest binary now runs 20 tests, all `policy::tests::*`.

## Test plan

- [x] `cargo test --workspace --all-features` — 1970 passed / 0 failed
- [x] Public-API set-diff vs `main` — 4018 → 4018, zero names added/removed
- [x] DAG gate — `me-types` has no `me-forget` production edge
- [x] All three `ForgetPolicy` paths still resolve (`memory_engine::`, `me_forget::`, `crate::forgetting::`)
- [x] 18 relocated tests present by name in `me-types`

Part of #816 · S5 sub-PR 1 of 3.


dutiona:
[Codex Review] — super-review, round 1 (Full gate)

**Primary question — "did any structurally-enforced guard get relocated across the crate boundary such that it is no longer enforced?"**

> No structurally enforced guard was lost. At both base `f1a308b` and HEAD, the sole production entry point still calls `policy.validate()?` as its **first executable statement**, before the storage read (`me-forget/src/policy.rs:193-205`). The moved `#[cfg(test)]` module remains local to its defining crate (`me-types/src/types/forgetting.rs:156-159`); no test-only helper crossed the boundary. `me-types` carries the same test-only unwrap exemption with the #725 justification (`me-types/src/lib.rs:15-19`), while production still inherits `clippy::unwrap_used = "deny"` because the allow is gated by `cfg(test)`. A normalized base/HEAD comparison of the struct, `Default`, both inherent methods, and all 18 tests is **byte-identical** after only changing L0-qualified paths. Fields and method bodies use only `std`, `crate::types::FactType`, and `crate::error`. `cargo tree -p me-types --edges normal` contains no `me-forget` edge.

---

**[MEDIUM] — The "Public API unchanged" claim misses a newly reachable facade path.** ✅ **ACCEPTED — fixed in `9a4f45c`**

> The facade publicly re-exports the entire `me_types::types` module (`memory-engine/src/lib.rs:123`); that module publicly exposes `forgetting` (`me-types/src/types/mod.rs:15`), which now publicly defines `ForgetPolicy`. Consequently consumers can newly write `memory_engine::types::forgetting::ForgetPolicy`; at the base commit that path ended at a module containing only `PruneStats`. The old `me_forget::types::ForgetPolicy` being private proves only that the old *definition-site* path was unwritable; it does not neutralize the new path leaked through the facade's pre-existing blanket module re-export. **The reported `cargo public-api` set comparison evidently canonicalizes/deduplicates the relocated type and therefore is not a reachable-path proof.**

**Verified against source — Codex is correct, and the PR body's claim was wrong.**

Resolution: the new path is **accepted, not contorted away**. It is *additive* (nothing removed), and it *restores symmetry* — `PruneStats` already had both a flat (`memory_engine::PruneStats`) and a nested (`memory_engine::types::forgetting::PruneStats`) path; the L3/L0 split was the only reason `ForgetPolicy` had just one. Making the module private instead would **remove** the pre-existing `memory_engine::types::forgetting::PruneStats` — a genuine break, strictly worse.

Instead the surface is now **pinned**: `reexports_are_accessible` (a compile-time probe that names paths) is extended to assert both nested paths. The PR body has been corrected to claim *additively extended*, not *unchanged*.

**The finding's real value is about the gate, not the path.** `cargo public-api` canonicalizes a type to one definition site — it counts *items*, so it structurally **cannot observe a path becoming newly writable**. Its 4018 → 4018 result was never the reachable-path proof it was presented as. This directly constrains **#941** (the re-export freeze, S5-3): the ratchet must compare **reachable paths**, not public-api item counts — otherwise the freeze ships blind to exactly this class of change.

dutiona:
[Antigravity Review] — super-review, round 1 (Full gate)

**Primary question — "did any structurally-enforced guard get relocated across the crate boundary such that it is no longer enforced?"**

1. **Is `ForgetPolicy::validate()` still called on every path it was called on before the move?**
   > Yes. The invocation `policy.validate()?` in `me-forget/src/policy.rs:199`, inside `pub async fn prune(...)`, was entirely untouched by the move. This path remains strictly enforced.

2. **Did a `#[cfg(test)]` item something depended on silently vanish?**
   > No. The test modules for `ForgetPolicy` successfully migrated to `me-types/src/types/forgetting.rs`. No test items, fixtures, or utilities vanished across the boundary.

3. **Is the `clippy::unwrap_used` opt-out correctly scoped in `me-types` — could the move have weakened the lint for NON-test code?**
   > `#![cfg_attr(test, allow(clippy::unwrap_used))]` was already present at `me-types/src/lib.rs`'s crate root. It does **not** weaken the lint for non-test code: `cargo clippy --workspace --all-targets` runs on the non-test library target where `cfg(test)` is false, so `unwrap_used` is still denied on production paths.

4. **Any runtime behavior change at all?**
   > No. All structurally enforced behaviors, fields, defaults, validations, and trait implementations (`Debug`, `Clone`) were relocated verbatim.

**Independent verification of the PR's claims:**

- **[LOW] Public API reachability** — `me_forget::types::ForgetPolicy` was the definition site but `mod types;` was private; the struct was reachable via `pub use types::ForgetPolicy;`. At HEAD that becomes `pub use me_types::types::forgetting::ForgetPolicy;`. The reachable path `me_forget::ForgetPolicy` is strictly preserved.
- **[LOW] anti-#903** — verified: base and HEAD yield the same test count; the 18 `ForgetPolicy` tests are preserved and execute within `me-types`.
- **[LOW] "No L3 dependencies"** — verified: fields, `is_decay_exempt()`, and `validate()` rely solely on `Result` and `FactType`, both L0.
- **[LOW] DAG gate** — verified: `me-types/Cargo.toml` has no dependency on `me-forget`; `me-types` remains a leaf.

**Verdict: no blocking findings.**

> ⚠️ **Orchestrator's note.** This clean verdict is recorded but **not** treated as sufficient. Earlier in this same epic (PR #976) agy *actively endorsed* a design that would have silently dropped every archived fact under Postgres — faithfully validating the code against the intent it had been given, when **the intent was the bug**. A faithfulness reviewer has no reason to distrust the stated intent, which is exactly why this PR also carries an adversarial structural lens and Codex — and it was **Codex** that refuted this PR's public-API claim (see the `[Codex Review]` comment). Agreement between faithful reviewers is not evidence; it is correlated blindness.

dutiona:
[Adversarial Review] — super-review, round 1 (Full gate, structural lens)

Tasked to **refute** the PR's claims, not validate them.

**Primary question — "did any structurally-enforced guard get relocated across the crate boundary such that it is no longer enforced?"**

> **No. Could not refute.** All four probes negative:
>
> 1. **`ForgetPolicy::validate()`** — call sites enumerated at base and HEAD *independently* (`git grep` on each ref, not read off the diff). Exactly two production invocations, at **identical file:line on both refs**: `me-forget/src/policy.rs:199` (first statement of `prune()`) and `bin/mcp/src/tools/handlers/p1.rs:102`. Both engine entry points route through `prune()` — `engine/forgetting.rs:30` and `engine/cognitive.rs:106` (`DreamContext::forget` → `self.engine.forget`). **No path reaches scoring without validation.**
> 2. **`#[cfg(test)]` cross-crate leak** — at base, `git grep "mod types|super::types"` in me-forget returns exactly 2 hits (`lib.rs:7`, `policy.rs:13`); at HEAD, **zero**. Nothing outside the file depended on the moved module.
> 3. **`clippy::unwrap_used` deny** — `me-types/src/lib.rs:19` already `#![cfg_attr(test, allow(...))]`; the moved *production* code (`Default`, `is_decay_exempt`, `validate`) contains **zero** `unwrap`. Deny not weakened.
> 4. **Runtime behavior — none, proven mechanically.** Normalized the base `me-forget/src/types.rs` (strip header, `me_types::`→`crate::`) and ran `difflib` against the HEAD file minus the pre-existing `PruneStats` block: **zero-line diff.** Not a reading — a diff.

---

### Independent reproduction of the `[Codex Review]` finding

The reviewer reproduced the public-API defect **from scratch**, with stronger evidence than either the PR or Codex had — an **out-of-tree compile probe**:

```rust
pub fn probe()   -> memory_engine::types::forgetting::ForgetPolicy { … }
pub fn control() -> memory_engine::types::forgetting::PruneStats   { … }
```

- against **HEAD**: `cargo build` → **exit 0**
- against **main** (`f1a308b`): **exit 101**, two errors, `ForgetPolicy` only —
  `error[E0433]: cannot find ForgetPolicy in forgetting`

The `PruneStats` control compiled on **both** refs — so the *module* was already public; **only `ForgetPolicy` is newly reachable.** Confirmed.

And it proved *why the gate is blind*, which is the part that matters:

> `cargo public-api -p memory-engine --simplified` on both refs: **1796 → 1796** items. `grep -c "memory_engine::types::" papi_base.txt` → **0**. The entire `memory_engine::types::*` surface renders as a **single opaque line** (`pub use memory_engine::types`). **The tool cannot see into it at all.**

---

### [ACCEPTED] Residual gap — the pin catches *removal*, not *addition*

> The pin at `lib.rs:385-394` guards against **removal** of that path (E0433 under `cargo test`). It does **not** guard against the next accidental ***addition*** — which is the exact failure class this PR just hit. Any future hoist into a `pub mod` under `me_types::types` will silently widen `memory_engine::*` again, and nothing in CI will notice.

**Correct, and this is the single most valuable line in the review.** A pin is a *floor*, not a *ratchet*. It has been carried into **#941** as a hard requirement — see the constraint comment posted there.

### [FIXED in `d1f6147`] Two inexact doc claims — the PR's own thesis, turned on itself

- **`crate-layout.md`** attributed the hoist of *both* `ForgetPolicy` **and** `PruneStats` to S5. `PruneStats` moved in **S1** (`6ae1a46`); only `ForgetPolicy` moves here. **Mis-dating the sibling's move, in the very PR whose argument is that stale rationales rot.**
- **`me-types/src/types/forgetting.rs`** enumerated the deps as *"nothing beyond `Result` and `FactType`"* and presented it as exhaustive. `validate()` also names `MemoryError` and `ConflictError` (all L0 — the *substantive* claim holds, but an enumeration that is silently incomplete cannot carry a "property-based rationale" load).

### [FIXED] NIT — PR body cited `lib.rs:123`; the actual line is **122**.

---

**Verdict: no blocker.** Gates re-run independently at `9a4f45c`, all exit 0, **1970 passed / 0 failed**; 18 relocated tests verified **by name**, each running exactly once; workspace attribute census `#[test]` 1201→1201 and `#[tokio::test]` 731→731 (base vs HEAD).

Issue #994, opened 2026-07-19
enhance(build): Wave 2 follow-up — dylib ship mode (opt-in thin bins)


dutiona:
## Context

S6 of the Wave 2 decomposition (#942, plan #925 §8 / §G) included an **opt-in dylib ship
mode** as a *time-boxed, explicitly deferrable* spike: build the facade `memory-engine` as
a shared library and link `cli`/`mcp` against it (`-Cprefer-dynamic`) to produce thin
binaries, instead of the default fat static bins. §G's box: *"if the crate-type/profile
wiring isn't clean, file this issue and STOP — S6's required output is the legibility
tooling, not the dylib."* The legibility tooling (`just linkmap` / `just symbols`) shipped;
the dylib is deferred here.

This issue records the probe results so whoever picks it up starts from data, not zero.

## What already works (prototyped, then reverted)

- Declaring `crate-type = ["rlib", "dylib"]` on the facade **builds cleanly** — `cargo build
  --workspace` is green and produces `libmemory_engine.so`.
- The **default build is unchanged**: with no `-Cprefer-dynamic`, `cli`/`mcp` still link the
  `rlib` statically (verified: `ldd target/debug/memory-engine-cli` shows no
  `libmemory_engine.so`). The dylib is only *preferred* under the flag.

## Why it is boxed (the two blockers)

1. **`-Cprefer-dynamic` is incompatible with the release profile's `lto = "thin"`.**
   ```
   error: cannot prefer dynamic linking when performing LTO
     = note: only 'staticlib', 'bin', and 'cdylib' outputs are supported with LTO
   ```
   The plan (§8) names *both* "static rlib + **thin-LTO**" (the default) and "opt-in thin via
   `-Cprefer-dynamic`" — but rustc forbids `prefer-dynamic` under LTO. So thin bins need a
   **dedicated non-LTO profile** (e.g. `[profile.thin] { inherits = "release", lto = false }`
   + `cargo build --profile thin`), not just a `RUSTFLAGS` toggle. That is added profile
   machinery, and it means the thin bins are **not** LTO-optimized.

2. **Declaring the `dylib` crate-type inflicts an always-on `.so` build.** Cargo builds
   *every* declared crate-type whenever the lib compiles, and crate-type **cannot** be
   feature- or profile-gated. So a `["rlib", "dylib"]` facade produces a large
   `libmemory_engine.so` (~142 MiB debug; it statically absorbs the rlib deps) on **every**
[...]

PR #995, opened 2026-07-19
refactor(build): S6 — just linkmap/symbols legibility tooling + fuzz-crate gate (#942, #993)

### LINKED ISSUE #942: refactor(build): S6 — dylib spike + just linkmap (Wave 2 #816)
S6 of the Wave-2 decomposition (plan #925). Optional opt-in dynamic-library build spike (default stays fat static-LTO) + a linkmap task. Deferrable.

### LINKED ISSUE #993: test(build): the detached fuzz crate consumes the facade API and NO gate covers it
## The gap

`fuzz/` depends on `memory-engine`, `memory-engine-embed`, and `memory-engine-mcp` **by path**, and names facade API types directly:

```rust
fuzz/fuzz_targets/dispatch.rs:23         use memory_engine::{ActivityFilterConfig, MemoryEngine};
fuzz/fuzz_targets/fact_deserialize.rs:22 use memory_engine::{Edge, Event, Fact};
fuzz/fuzz_targets/fts_query.rs:24        memory_engine::fuzz_seam::fuzz_fts_query(&query);
```

But it is **deliberately detached from the workspace** — its own header says so:

> *"cargo-fuzz crate for memory-engine — **DELIBERATELY OUTSIDE the main workspace**. The empty `[workspace]` table below detaches this package from the parent workspace so that `cargo build --workspace --all-features` at the repo root **never compiles it**. The fuzz targets require the nightly-only libFuzzer sanitizer runtime…"*

That detachment is **correct** — pulling nightly + a sanitizer runtime into the default members would force nightly on the stable main build. But it has a consequence nobody wrote down:

> **A facade API removal can break a real in-repository Rust consumer, and no gate detects it.**

Every gate in the matrix (`build --workspace`, `test --workspace --all-features`, the facade-alone checks, clippy) skips `fuzz/` **by construction**. `cargo +nightly fuzz build` is the only thing that would catch it, and it is not in the matrix.

This is a hole **today**, independent of any freeze decision. It has presumably been open since the fuzz crate was detached.

## Why it matters more now

1. **CI is `workflow_dispatch`-only (#989)** — the local matrix is the *primary* verification. A gap in the matrix is no longer backstopped by anything.
2. **Wave 2 (#816) broke the facade's public API four times.** Each break was checked against cli/mcp/embed via `--workspace`. **None was checked against `fuzz/`.** It may already be broken — nobody has looked.
3. It refuted a claim in ADR-0018's S5 amendment (*"the in-workspace consumers break at compile time, loudly, in the same `--workspace` run"* — true, but not a complete account of the consumers). The ADR is now narrowed to say so explicitly, and points here.

## First question — is it broken right now?

Before designing a gate, find out:

```bash
cargo +nightly fuzz build      # needs: rustup toolchain install nightly; cargo install cargo-fuzz
```

If it fails, Wave 2 already broke it and the repair is part of this issue.

## Options for the gate

1. **Add `cargo +nightly fuzz build` to the gate matrix** (CLAUDE.md / AGENTS.md / GEMINI.md / qa-sweep) and to `ci.yml`. Honest and complete. Cost: every contributor needs nightly + `cargo-fuzz` installed, and it is a slow build. Heavy for a matrix that is already 11 commands.
2. **Add it to `ci.yml` only**, as its own job. Cheap for contributors. ⚠️ But CI is manual (#989) — it would only run when dispatched, so it is not a gate so much as a periodic check.
3. **A cheap compile-only proxy:** a `cargo check`-style probe that names the facade paths `fuzz/` uses, living *inside* the workspace so the normal matrix covers it. Catches removals of exactly the API surface fuzz depends on, without nightly. Does not catch `#[cfg(fuzzing)]`-gated seam drift.
4. **Document-only:** a note in the gate matrix — "if you remove or rename a facade public item, run `cargo +nightly fuzz build`". Cheapest; relies on discipline, which this repo has repeatedly shown decays (cf. #987, #990, #986).

**Recommendation: (1) or (3).** (4) is the option this repo keeps proving does not work — a rule that says "remember to check X" is exactly the shape of guard that Wave 2 kept finding rotted. (2) is weak while CI is manual.

## Acceptance

- [ ] Determine whether `cargo +nightly fuzz build` currently passes on `main`; if not, repair it
- [ ] A gate exists that fails when a facade API removal breaks `fuzz/`
- [ ] The gate is reachable by whatever actually runs (given CI is manual — see #989)
- [ ] ADR-0018's narrowed claim points at the resolution

Surfaced by **[Codex Review]** on **#992** (S5-3). Related: #989 (CI manual → the matrix is the only gate), #941 (the freeze deferral whose premise this narrowed).

dutiona:
## Summary

**S6 — the last slice of the Wave 2 crate decomposition (#942)** — plus its collateral **fuzz-crate gate (#993)**. No production code changes: a new `justfile`, one CI job, and doc/matrix updates. The anti-#903 suite stays at **1975** tests.

## What / Why / How

**Legibility tooling (#942, S6's required half).** The 18-crate split is only legible if you can *see* the link graph. New `justfile`:
- `just linkmap` — the crate DAG (`cargo tree --edges normal`), per-crate binary-size contribution (`cargo bloat --release --crates`), and native link libs (`rustc --print native-static-libs`, via an isolated throwaway `staticlib` build in `target/linkmap-nsl` so it never churns the normal release artifacts).
- `just symbols` — the exported dynamic symbols of the release bins (`nm -D --defined-only`, guarded for non-Linux). A leaf executable exports ~nothing — this surface only gets interesting for the deferred #994 dylib.
- `just` documented as a dev prerequisite (CLAUDE.md).

**dylib ship mode (#942 spike) — DEFERRED to #994 per plan §G's box.** Probed: `crate-type=["rlib","dylib"]` builds cleanly and leaves default bins static, but (a) `-Cprefer-dynamic` bins are incompatible with the release `lto="thin"` profile (`error: cannot prefer dynamic linking when performing LTO`), and (b) the crate-type inflicts an always-on ~142M `.so` on every facade build with no clean opt-in (crate-type can't be feature/profile-gated). §11 lists deferral-with-tracking-issue as a co-equal success outcome. Crate-type reverted; #994 filed under #816 with the full probe record.

**Fuzz-crate gate (#993).** The detached `fuzz/` workspace consumes the facade public API by path, but no `--workspace` command compiles it — so a facade API removal could break a real in-repo consumer unnoticed. `cargo +nightly fuzz build` (confirmed passing on this branch) is now in the gate matrix (CLAUDE.md / AGENTS.md / GEMINI.md / qa-sweep — byte-identical) + a dedicated `fuzz` ci.yml job (cache scoped to `fuzz/target`) + a `just fuzz-build` wrapper. ADR-0018's narrowed claim updated to record the resolution.

## Super-review

Reviewed by 3 diverse-lens adversarial subagents (shell-correctness / CI-gate-integrity / decision-completeness) + **Codex** (external, `gpt-5.6-sol`) + `gemini-code-assist[bot]` — each *ran* the recipes/gate to verify, not just read. All findings triaged against source and fixed:
- subagents: 1 MED (CI rust-cache missed `fuzz/target`) + 6 LOW/NIT → `e3c5141`
- Codex: 1 MED (`nm -D` doesn't prove self-containment) + 2 LOW (native-libs probe status, moving-nightly cache caveat); gemini: 1 MED (non-Linux `nm -D` guard) → `5e2171a`

Codex's external pass caught what the Claude subagents blessed (the "self-contained" framing) and refined a subagent-suggested fix. agy timed out (degraded — Codex covered the external voice). All 4 review threads resolved.

## Test plan

- [x] `just linkmap` reports the crate DAG + per-crate bloat + `native-static-libs`
- [x] `just symbols` reports exported dynamic symbols of the release cli + mcp bins
- [x] `cargo +nightly fuzz build` passes (the #993 gate)
- [x] The fuzz matrix line is byte-identical across CLAUDE.md / AGENTS.md / GEMINI.md / qa-sweep and matches ci.yml
- [x] Full CI-exact gate matrix green; anti-#903 = 1975 exact

Closes #942. Closes #993. Refs #994.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,8 +108,9 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@nightly
       # The detached `fuzz/` workspace builds into `fuzz/target`, which rust-cache's default
-      # (`. -> target`) does NOT cover — cache it explicitly so this nightly build is not cold
-      # on every dispatch (the #989 runner-minute motive that made CI manual).
+      # (`. -> target`) does NOT cover — cache it explicitly. Caveat: `@nightly` moves daily and
+      # rust-cache keys on the rustc commit, so this warms repeated dispatches on the SAME
+      # nightly, not across days — still worth it given the #989 runner-minute motive.
       - uses: Swatinem/rust-cache@v2
         with:
           workspaces: "fuzz -> target"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,28 @@ jobs:
       - name: cargo test (all features)
         run: cargo test --workspace --all-features
 
+  fuzz:
+    name: Fuzz build (#993)
+    runs-on: ubuntu-latest
+    # The detached `fuzz/` workspace (empty `[workspace]`) consumes the facade public API
+    # by path, but NO `--workspace` job compiles it — it needs nightly + the libFuzzer
+    # sanitizer runtime. This is the only gate that catches a facade API removal/rename
+    # breaking a real in-repo consumer. Mirrors `cargo +nightly fuzz build` in the local
+    # gate matrix (CLAUDE.md / AGENTS.md / GEMINI.md / qa-sweep). See #993.
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@nightly
+      # The detached `fuzz/` workspace builds into `fuzz/target`, which rust-cache's default
+      # (`. -> target`) does NOT cover — cache it explicitly so this nightly build is not cold
+      # on every dispatch (the #989 runner-minute motive that made CI manual).
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: "fuzz -> target"
+      - name: install cargo-fuzz
+        run: cargo install cargo-fuzz --locked
+      - name: cargo +nightly fuzz build
+        run: cargo +nightly fuzz build
+
   cargo-deny:
     name: Supply-chain (cargo-deny)
     runs-on: ubuntu-latest
--- a/justfile
+++ b/justfile
@@ -1,0 +1,91 @@
+# memory-engine — developer task runner.
+#
+# Wave 2 (#816 / S6, #942) legibility tooling: make the 18-crate link graph and
+# the shipped symbol surface inspectable, so the decomposition is legible in
+# practice — not just on paper. The crate split already delivered the structural
+# value; these recipes let you *see* it.
+#
+# Requires `just` (https://github.com/casey/just):  cargo install just
+# Individual recipes name any extra tooling they need and print an install hint
+# when it is missing, so a fresh clone degrades gracefully instead of erroring.
+
+# List the available recipes (runs on a bare `just`).
+default:
+    @just --list --unsorted
+
+# The link graph of the shipped facade: the layered crate DAG (normal edges = what
+# actually links, dev-deps excluded), each crate's contribution to binary size,
+# and the native system libraries the final link pulls in. This is the "can I SEE
+# the 18-crate split?" report the Wave 2 plan (#925 §8 / §11) calls for.
+
+# Report the crate DAG, per-crate binary-size split, and native link libs.
+linkmap:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    sep() { printf '\n── %s ──\n' "$1"; }
+
+    sep "Crate DAG — facade \`memory-engine\`, normal edges (the ship graph)"
+    cargo tree -p memory-engine --edges normal
+
+    sep "Per-crate contribution to the release \`memory-engine-cli\` binary"
+    if command -v cargo-bloat >/dev/null 2>&1; then
+        cargo bloat --release --crates --bin memory-engine-cli || echo "  (cargo bloat failed)"
+    else
+        echo "  (skipped — install with: cargo install cargo-bloat)"
+    fi
+
+    sep "Native static libs the final link requires"
+    # `--print native-static-libs` only emits for a `staticlib` crate-type, so override
+    # it for this one rustc invocation (the facade's real crate-type is untouched). Use a
+    # dedicated target-dir so this throwaway staticlib build does NOT invalidate the normal
+    # release artifacts `just symbols` shares.
+    nsl=$(cargo rustc -p memory-engine --release --quiet --lib --crate-type staticlib \
+        --target-dir target/linkmap-nsl -- --print native-static-libs 2>&1 || true)
+    if echo "$nsl" | grep -qi 'native-static-libs:'; then
+        echo "$nsl" | grep -i 'native-static-libs:'
+    elif echo "$nsl" | grep -qiE 'error(\[|:)'; then
+        echo "  (staticlib probe failed to compile — rerun verbose: cargo rustc -p memory-engine --release --lib --crate-type staticlib -- --print native-static-libs)"
+    else
+        echo "  (rustc reported none; run \`ldd target/release/memory-engine-cli\` for runtime shared libs)"
+    fi
+
+# The shipped symbol surface: the *dynamic* symbols the release binaries export for
+# dynamic linking (`nm -D`, Linux/ELF-specific). A fat static bin exports almost nothing
+# dynamically — that near-empty list IS the signal that it is self-contained. The `dylib`
+# ship mode that would make this surface explode into the ABI is deferred (#994); today
+# these are the default fat static bins.
+
+# Report exported dynamic symbols of the release cli + mcp binaries.
+symbols:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    cargo build --release --bins
+    for art in target/release/memory-engine-cli target/release/memory-engine-mcp; do
+        name=$(basename "$art")
+        if [ ! -e "$art" ]; then printf '── %s: (not built) ──\n' "$name"; continue; fi
+        count=$(nm -D --defined-only "$art" 2>/dev/null | wc -l | tr -d ' ')
+        printf '── %s: %s exported dynamic symbols ──\n' "$name" "$count"
+        nm -D --defined-only "$art" 2>/dev/null | awk '{print "   " $NF}' | sort | head -30 || true
+        [ "${count:-0}" -gt 30 ] && echo "   … ($((count - 30)) more; full: nm -D --defined-only $art)" || true
+    done
+
+# The fuzz crate (`./fuzz`) is a DETACHED workspace: the repo-root
+# `cargo build --workspace` never compiles it, yet it consumes the facade public
+# API by path (memory-engine / -embed / -mcp). A facade API removal can break it
+# with no default gate noticing (#993). This recipe IS that gate — it builds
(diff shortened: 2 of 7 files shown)


justfile
```diff
@@ -0,0 +1,91 @@
...
+        echo "$nsl" | grep -i 'native-static-libs:'
+    elif echo "$nsl" | grep -qiE 'error(\[|:)'; then
+        echo "  (staticlib probe failed to compile — rerun verbose: cargo rustc -p memory-engine --release --lib --crate-type staticlib -- --print native-static-libs)"
+    else
+        echo "  (rustc reported none; run \`ldd target/release/memory-engine-cli\` for runtime shared libs)"
+    fi
+
+# The shipped symbol surface: the *dynamic* symbols the release binaries export for
+# dynamic linking (`nm -D`, Linux/ELF-specific). A fat static bin exports almost nothing
+# dynamically — that near-empty list IS the signal that it is self-contained. The `dylib`
```
dutiona: [Codex Review] [MEDIUM] `nm -D --defined-only` cannot support the self-contained/static claim here. On the current release artifacts it reports 0 defined dynamic symbols for both binaries, while `file` and `readelf -d` show both are dynamically linked and require shared libraries (`libssl`, `libcrypto`, `libgcc_s`, `libm`, and `libc`; MCP also needs the ELF loader). The option deliberately omits undefined imports, so a near-empty result is not evidence of self-containment; likewise a future `-Cprefer-dynamic` binary would carry facade imports as `U`, while the definitions would live in the dylib. Please either narrow this recipe and prose to binary exports only, or add `readelf -d`/`ldd` plus undefined-symbol reporting; when #994 adds the dylib, inspect the dylib itself.

justfile
```diff
@@ -0,0 +1,91 @@
...
+        echo "  (skipped — install with: cargo install cargo-bloat)"
+    fi
+
+    sep "Native static libs the final link requires"
+    # `--print native-static-libs` only emits for a `staticlib` crate-type, so override
+    # it for this one rustc invocation (the facade's real crate-type is untouched). Use a
+    # dedicated target-dir so this throwaway staticlib build does NOT invalidate the normal
+    # release artifacts `just symbols` shares.
+    nsl=$(cargo rustc -p memory-engine --release --quiet --lib --crate-type staticlib \
+        --target-dir target/linkmap-nsl -- --print native-static-libs 2>&1 || true)
```
dutiona: [Codex Review] [LOW] This discards the definitive Cargo exit status with `|| true` and then guesses failure from a narrow English diagnostic regex, so the advertised compile-failure-versus-no-output distinction is not reliable. `set -o pipefail` adds another failure mode: both `echo "$nsl" | grep -q ...` tests can report false when a long diagnostic payload fills the pipe and `echo` receives SIGPIPE. I reproduced that exact pattern with a matching 200k-line payload: Bash printed `echo: write error: Broken pipe` and selected the non-match branch. Capture the command status with `if nsl=$(... 2>&1); then ... else ... fi`, use a here-string for matching, and classify failure from the status rather than output text.

.github/workflows/ci.yml
```diff
@@ -96,6 +96,28 @@ jobs:
...
+    # by path, but NO `--workspace` job compiles it — it needs nightly + the libFuzzer
+    # sanitizer runtime. This is the only gate that catches a facade API removal/rename
+    # breaking a real in-repo consumer. Mirrors `cargo +nightly fuzz build` in the local
+    # gate matrix (CLAUDE.md / AGENTS.md / GEMINI.md / qa-sweep). See #993.
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@nightly
+      # The detached `fuzz/` workspace builds into `fuzz/target`, which rust-cache's default
+      # (`. -> target`) does NOT cover — cache it explicitly so this nightly build is not cold
+      # on every dispatch (the #989 runner-minute motive that made CI manual).
```
dutiona: [Codex Review] [LOW] The `fuzz -> target` mapping is correct, but moving `nightly` defeats this stated cross-dispatch benefit whenever the date changes. `rust-cache` includes the rustc release and commit hash in its restore key, and its own cache-effectiveness documentation says that an unpinned nightly cache is thrown away every day. With manual CI, dispatches will commonly occur on different dates and remain cold. Pin a dated nightly and update it deliberately, or narrow this comment to repeated dispatches using the same nightly.

dutiona (commented):
[Codex Review] 1 MEDIUM + 2 LOW findings. The MEDIUM is central to S6 legibility: `nm -D --defined-only` reports binary exports, not self-containment or the dynamic link surface, and the current release bins prove the distinction (0 reported exports while ELF metadata lists multiple shared-library dependencies). The LOW findings cover unreliable `cargo rustc` failure classification under `pipefail` and the daily invalidation of the unpinned-nightly cache. The `fuzz -> target` mapping itself is correct, the fuzz job has no untrusted-input interpolation, the four matrix command lines are byte-identical, the tool guards parse, and the ADR #993 resolution is coherent on the proposed merged state. No `ci.yml` run exists for this branch/head, so the new job has not yet produced a CI witness.

justfile
```diff
@@ -0,0 +1,91 @@
...
+    #!/usr/bin/env bash
+    set -euo pipefail
+    cargo build --release --bins
+    for art in target/release/memory-engine-cli target/release/memory-engine-mcp; do
+        name=$(basename "$art")
+        if [ ! -e "$art" ]; then printf '── %s: (not built) ──\n' "$name"; continue; fi
+        count=$(nm -D --defined-only "$art" 2>/dev/null | wc -l | tr -d ' ')
+        printf '── %s: %s exported dynamic symbols ──\n' "$name" "$count"
+        nm -D --defined-only "$art" 2>/dev/null | awk '{print "   " $NF}' | sort | head -30 || true
+        [ "${count:-0}" -gt 30 ] && echo "   … ($((count - 30)) more; full: nm -D --defined-only $art)" || true
```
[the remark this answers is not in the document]
dutiona: Adopted the platform guard in 5e2171a — `nm -D` is now probed before use, so on a non-Linux `nm` the recipe prints an 'unsupported here (needs Linux/ELF nm -D)' line and continues instead of hard-crashing under `set -euo pipefail`. Matches the justfile's own 'degrade gracefully instead of erroring' ethos. Thanks.

justfile
dutiona: Fixed in 5e2171a. You're right — `nm -D --defined-only` lists EXPORTS, not imports, so ~0 symbols means 'leaf executable / provides no dynamic symbols', NOT self-contained (the bins still dynamically link libssl/libc — reframed the doc to say so, pointing at `ldd`). The exported surface only becomes meaningful for the deferred #994 dylib.

justfile
dutiona: Fixed in 5e2171a. The probe now branches on cargo's real exit status (`if nsl=$(...); then`) so a genuine staticlib compile failure is never misreported as 'no native libs', and it matches with a here-string (`grep ... <<<"$nsl"`) — no `echo | grep` SIGPIPE on a large diagnostic payload.

.github/workflows/ci.yml
dutiona: Fixed in 5e2171a — narrowed the comment. `@nightly` moves daily and rust-cache keys on the rustc commit, so the cache warms repeated dispatches on the SAME nightly, not across days. (Kept `@nightly` rather than pinning, so the fuzz job tracks the current sanitizer runtime.)

PR #996, opened 2026-07-19
refactor: Wave 2 collateral quick-wins — #980 .pak double-hash · #979 me-archive surface · #969 DIM seeds · #991 docs⊇ci gate

### LINKED ISSUE #969: test(forgetting): seed test embeddings from me_test_support::fixtures::DIM instead of inline magic 4
## Problem

During the Wave 2 crate decomposition (#816), primitives that are carved into their own crates reuse `me_test_support::factory::SqliteFactory::make()` for backend construction (the Option-b harness-reuse pattern). `make()` provisions the in-memory pool at `me_test_support::fixtures::DIM` (currently `4`).

However, the tests' **embedding seed literals** stayed inline as magic `4`s — a mix of `vec![0.1; 4]` and `let embed_dim = 4; vec![0.1; embed_dim]`. Surfaced by the PR #968 (`me-forget`) super-review:

- `memory-engine/lib/me-forget/src/policy.rs` — ~10 seeding sites hardcode dim `4` (e.g. `:287`, `:350`, `:398`, `:631`, `:1004`).

The pool dimension now derives from `fixtures::DIM` in **a different crate** than the embedding literals. They are consistent today (both `4`), but a future change to `DIM` would silently desync the seed-vector length from the pool dimension, breaking `insert_fact` with a dimension-mismatch — a latent trap that the carve moved across a crate seam.

This is a **pre-existing test pattern** (byte-identical to pre-carve), not introduced by any single carve — but it now recurs in every crate the decomposition carves out that reuses `SqliteFactory` (me-forget today; me-resolve and the S4 crates next). Tracking it once, systemically.

## Proposed fix

In the carved crates' test modules, seed embeddings with the shared constant instead of an inline literal:

```rust
// before
embedding: vec![0.1; 4],
let embed_dim = 4;  embedding: vec![0.1; embed_dim],
// after
embedding: vec![0.1; me_test_support::fixtures::DIM],
```

Behavior-preserving (`DIM == 4`), so the anti-#903 test total is unaffected. Do it as a sweep across carved crates rather than per-carve-PR, to keep each carve a clean mechanical relocation.

## Scope
- [ ] `me-forget` test seeding → `fixtures::DIM`
- [ ] apply the same to me-resolve / S4 crates as they land (or a single sweep once carving completes)

Refs #816 (Wave 2 decomposition). Surfaced by PR #968 super-review.

### LINKED ISSUE #979: refactor(build): narrow me-archive's pub surface to its facade seam (post-carve polish)
## Summary

`me-archive` (carved in #978, Wave 2 #816 / S4 sub-PR 3b) exposes a broader `pub` surface than its facade seam actually requires. Surfaced by Codex during the #978 super-review; **declined in that PR on the CARVE DoD** (sub-PR bar = compiles + tests + CI hygiene, *not* public-api minimal — API-boundary polish is deferred to a post-carve polish wave) and filed here instead.

## The surface

`me-archive/src/lib.rs` declares every implementation module `pub`, which exposes helpers whose only call sites at PR-head are **inside `me-archive`**:

| item | location | who actually needs it |
|---|---|---|
| `hash_file` | `pak.rs` | intra-crate only |
| `verify_pak` | `pak.rs` | intra-crate only |
| `search_archives` | `search.rs` | intra-crate (`manage.rs`) |
| `is_within_archive_dir` | `search.rs` | intra-crate — plus one facade test's fixture-sanity assert |
| `write_pak_and_hash` | `pak.rs` | intra-crate — plus retained facade tests |

Note `me-archive` is `publish = false`, so "public" here means **workspace-internal**, not a shipped API. There is no semver consequence; this is hygiene and encapsulation.

## Suggested fix (polish wave)

Keep intra-crate helpers `pub(crate)` and expose only the orchestration + DTO surface plus the deliberate `read_pak` fuzz seam. The two items pinned `pub` purely by *retained facade tests* (`write_pak_and_hash`, `is_within_archive_dir`) want one of:
- move those low-level tests into `me-archive` (where the properties now live — this is what was already done for `pak_write_stamp_matches_read_gate`, which had been standing up a whole SQLite engine just to read back a constant); or
- add a narrowly-documented `#[doc(hidden)]` test seam.

`build_pak` was already narrowed to `pub(crate)` in #978 by exactly the first route, so the pattern is established.

## Context

Same class as the deferred items in the #935 body ledger (`from_pool` visibility, `CURRENT_SCHEMA_VERSION` re-export rendering, `#[must_use]` on `::new` constructors). Worth doing as one sweep across all the carved crates rather than piecemeal per carve.

Refs #816, #978

### LINKED ISSUE #980: refactor(storage): verify_single_archive hashes the .pak twice on the mismatch path
## Summary

`verify_single_archive` hashes a `.pak` **twice** on the corruption path. Surfaced by `gemini-code-assist` on #978; **pre-existing** (identical at base `62c5b9e`), so not folded into that PR — the carve moved the code verbatim.

## The redundancy

`verify_pak` already hashes the file internally:

```rust
pub fn verify_pak(path: &Path, expected_hash: &str) -> Result<bool> {
    let actual = hash_file(path)?;
    Ok(actual == expected_hash)
}
```

but it returns only a `bool`, discarding the hash it computed. So `verify_single_archive`'s mismatch arm has to recompute it just to format the message:

```rust
match verify_pak(pak_path, &entry.blake3_hash) {
    Ok(true) => ...,
    Ok(false) => {
        let actual = hash_file(pak_path).unwrap_or_default();   // <-- second full read + blake3
        ... format!("hash mismatch: expected {}, got {actual}", entry.blake3_hash)
    }
    ...
}
```

A `.pak` can be large (the read cap is 4 GiB), so this is a full second read + blake3 pass over the file — spent purely to build an error string.

## Impact

Bounded: it only triggers on an **actual hash mismatch** (a corrupted or tampered archive), i.e. an already-degraded path, and `verify_archives` is an explicit maintenance operation, not a hot path. Correctness is unaffected. Hence `type:refactor`, not `type:bug`.

## Suggested fix

Call `hash_file` **once** in `verify_single_archive` and compare there, dropping the `verify_pak` hop (or have `verify_pak` return the computed hash — e.g. `Result<Result<(), String>>` or `Result<(bool, String)>` — so the caller can reuse it). The former is simpler and `verify_pak`'s only other caller is the test suite; check that before choosing.

Location (post-#978): `memory-engine/lib/me-archive/src/manage.rs` (`verify_single_archive`) + `memory-engine/lib/me-archive/src/pak.rs` (`verify_pak`, `hash_file`).

Refs #978

### LINKED ISSUE #991: test(build): enforce that the docs' gate matrix equals ci.yml — the drift was silent and is now load-bearing
## The problem

`CLAUDE.md`, `AGENTS.md`, and `GEMINI.md` each carry a verification-gate matrix that is supposed to equal what `.github/workflows/ci.yml` runs. **Nothing enforces that.** They drifted, and nobody noticed.

Found during #990 (chatgpt-codex-connector, P2): CI's build job runs two commands the matrix never listed —

```
ci.yml:85  cargo check -p memory-engine
ci.yml:87  cargo check -p memory-engine --no-default-features --features backend-sqlite
```

— while all three docs claimed *"these are the **exact** commands CI runs"*. The drift was harmless only because **CI was the real gate and the matrix a convenience**: nobody had to keep them equal, so nobody did.

**#990 removes that slack.** With CI on `workflow_dispatch`, the matrix *is* the verification. An incomplete matrix is now an unenforced gate — and the two that were missing exist precisely because of **#978**, where feature unification silently killed archive-OFF coverage (the configuration downstream consumers get by default).

#990 fixed the two commands. It did **not** fix the thing that let them drift.

## The fix — make the equality mechanical

A test that reads `ci.yml`, extracts every `run: cargo …`, extracts the cargo lines from each doc's gate block, and asserts **docs ⊇ CI**. It fails the moment someone adds a CI step without documenting it, or edits the matrix out of sync.

The elegant part: it would run under **`cargo test --workspace --all-features`** — which is itself in the matrix. **The guard runs inside the thing it guards**, so it cannot be skipped by anyone who runs the gate at all. (A CI-based checker would be useless here: CI is manual, so it might never fire.)

Sketch (a `#[test]` in the workspace, or a small `utils/scripts/` check invoked from one):

```python
norm = lambda s: ' '.join(s.split())            # docs pad for column alignment; CI does not
ci_cmds = every `run: cargo …` in ci.yml, excluding the msrv job (documented separately)
doc_cmds = cargo lines from the ```bash block containing 'cargo fmt --all --check'
assert set(ci_cmds) <= set(doc_cmds)            # docs may add (e.g. `cargo deny`); never omit
```

Two gotchas already hit while verifying #990 by hand:

1. **Whitespace.** Docs write `cargo test··--workspace` (two spaces, aligned columns); CI writes one. An exact-string diff false-positives on every line. Normalise.
2. **The MSRV job** is deliberately doc'd as prose (`cargo +1.88 …`), not in the matrix block — exclude it, or encode the exception explicitly.

## Acceptance

- [ ] The check exists and runs as part of `cargo test --workspace --all-features`
- [ ] It **fails** when a `run: cargo` step is added to `ci.yml` and not to the docs (prove it by trying)
- [ ] It **passes** on current `main`
- [ ] Whitespace-insensitive; MSRV exception explicit, not accidental

## Why this shape

This repo's recurring, expensive defect is **a doc asserting a guard that no longer fires** — #987's recursion-trap docs, #986's public-api "proof", the `.pak` schema gate, and now this. The pattern that has actually worked each time is the same: **turn the convention into a compiler- or test-enforced invariant.** A rule that says "keep these in sync" decays. A test that fails when they diverge does not.

Discovered in #990 · relates to #978 (why those two gates exist) · relates to #989 (why the matrix became load-bearing).

dutiona:
## Summary

Four Wave-2 collateral quick-wins, batched per plan #925 / the S6 handoff. Behaviour-preserving except #991 (a new test); the anti-#903 suite goes 1975 → **1976**.

`Classification: refactor (user-confirmed)`

## The four

**#980 — `verify_single_archive` hashes the `.pak` once** (`me-archive/manage.rs`). The mismatch arm ran `verify_pak` (which hashes) then `hash_file` *again* to format the message — a full second read + blake3 over a file up to the 4 GiB read cap, on an already-degraded corruption path. Hash once, compare inline.

**#979 — narrow me-archive's pub surface** (`me-archive/{pak,search,lib}.rs` + facade `engine/archive.rs`). The implementation helpers (`hash_file`, `search_archives`, `is_within_archive_dir`, `write_pak_and_hash`) were `pub` only because their modules are `pub` → now `pub(crate)`; `verify_pak` removed (dead after #980). The facade's redundant `is_within_archive_dir` sanity-assert dropped (me-archive already unit-tests the guard). `read_pak` fuzz seam stays `pub`. me-archive is `publish=false`, so this is workspace-internal encapsulation, no semver surface.

**#969 — seed me-forget test embeddings from `fixtures::DIM`** (`me-forget/policy.rs`). 11 inline magic-`4` seeds → the shared `DIM` the `SqliteFactory` pool is provisioned with, so a future `DIM` change can't silently desync seed length from pool dimension. me-forget is the *only* carved crate carrying the pattern (me-types excluded — L0 DTO tests, a `DIM` import would be a dependency cycle; the facade snapshot test excluded — intrinsic 4-dim literal vectors).

**#991 — enforce the docs' gate matrix ⊇ ci.yml** (`tests/gate_matrix_parity.rs`). A `#[test]` running under `cargo test --workspace --all-features` (the guard inside the thing it guards) that fails when a `run: cargo` gate is added to `ci.yml` but not to CLAUDE.md / AGENTS.md / GEMINI.md. Whitespace-normalised; excludes the MSRV job + `cargo install` setup; `cargo deny check` runs via a CI action so it stays legitimately doc-only. **Proven** both ways (passes on main, fails on injected drift).

## Test plan

- [x] `cargo test -p me-archive` (verify_single_archive + hash_file tests)
- [x] `cargo test -p me-forget` (seeds compile + pass, count unchanged)
- [x] #991 guard proven: passes on main, fails on an injected `ci.yml` drift
- [x] Full CI-exact gate matrix green; `clippy --all-features -D warnings` clean; anti-#903 = 1982

Closes #969. Closes #979. Closes #980. Closes #991.

### DIFF
--- a/memory-engine/lib/me-archive/src/search.rs
+++ b/memory-engine/lib/me-archive/src/search.rs
@@ -85,7 +85,7 @@ impl Ord for MinScored {
 /// (`engine/archive.rs`) exercises it directly across the crate boundary (Wave 2
 /// #816 / S4, sub-PR 3b).
 #[must_use]
-pub fn is_within_archive_dir(pak_path: &str) -> bool {
+pub(crate) fn is_within_archive_dir(pak_path: &str) -> bool {
     Path::new(pak_path)
         .components()
         .all(|component| matches!(component, Component::Normal(_) | Component::CurDir))
@@ -132,7 +132,7 @@ const fn entry_is_prunable(_entry: &ArchiveManifestEntry, _query: &MemoryQuery) 
 /// Each `.pak` is opened via [`read_pak`], whose schema gate checks `me-types`'
 /// backend-independent `ARCHIVE_SCHEMA_VERSION` — the same constant the write side
 /// stamps. No backend schema version is involved (Wave 2 #816 / S4, sub-PR 3a).
-pub fn search_archives(
+pub(crate) fn search_archives(
     archive_dir: &Path,
     manifest_entries: &[ArchiveManifestEntry],
     query: &MemoryQuery,
--- a/memory-engine/lib/me-archive/src/search.rs
+++ b/memory-engine/lib/me-archive/src/search.rs
@@ -79,11 +79,10 @@ impl Ord for MinScored {
 /// for not-yet-existing files), and is purely lexical — no `canonicalize`, no
 /// TOCTOU window.
 ///
-/// `pub` (not `pub(crate)`): the **shared archive path-containment guard** — the
-/// sibling `verify_archives` (`crate::manage`) reuses this helper rather than
-/// re-deriving the check (#292), and the facade's retained regression test
-/// (`engine/archive.rs`) exercises it directly across the crate boundary (Wave 2
-/// #816 / S4, sub-PR 3b).
+/// The **shared archive path-containment guard**: `crate::manage`'s `verify_archives`
+/// reuses this helper rather than re-deriving the check (#292). `pub(crate)` — an
+/// implementation detail of me-archive, covered by this module's own traversal unit
+/// tests (the facade tests `verify_archives` end-to-end, not this helper directly; #979).
 #[must_use]
 pub(crate) fn is_within_archive_dir(pak_path: &str) -> bool {
     Path::new(pak_path)
--- a/tests/gate_matrix_parity.rs
+++ b/tests/gate_matrix_parity.rs
@@ -5,12 +5,15 @@
 //! `GEMINI.md` *is* the verification, not a convenience mirror of CI. An incomplete matrix
 //! is an unenforced gate — exactly the drift #990 found by hand (two `cargo check` gates CI
 //! ran that the docs never listed). This guard runs inside `cargo test --workspace
-//! --all-features` (itself a matrix line), so it cannot be skipped by anyone who runs the
+//! --all-features` (itself a matrix line), so it can't be skipped by anyone who runs the
 //! gate at all — unlike a CI-based checker, which is useless while CI is manual.
 //!
-//! It fails the moment someone adds a `run: cargo …` gate to `ci.yml` without documenting
-//! it in all three matrices (docs may *add* commands — e.g. `cargo deny check`, which CI
-//! runs via an action rather than `run:` — but never omit one CI runs).
+//! The CI side is parsed as a YAML *subset* rather than line-scraped, because a guard with
+//! blind spots is worse than none (it fails *open* — a drift slips through green). It walks
+//! `jobs.<job>.steps[].run`, handling the real GitHub-Actions step forms: named
+//! (`run: cargo …`), nameless (`- run: cargo …`), block scalars (`run: |`), and quoted
+//! scalars; the `#`-comment strip is applied to BOTH sides so a valid trailing comment on a
+//! CI line can't fail the guard *closed*. The `parser_*` mutation tests below lock each form.
 
 use std::collections::BTreeSet;
 use std::path::PathBuf;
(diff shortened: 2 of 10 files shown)


tests/gate_matrix_parity.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Every `cargo` gate command CI actually runs, minus two documented exceptions:
+/// - `cargo install …` — toolchain setup, not a gate (the fuzz job installs cargo-fuzz).
+/// - `… --tests --examples` — the MSRV job, deliberately documented as prose, not in the
+///   matrix block (#991).
+fn ci_gate_commands(ci_yml: &str) -> BTreeSet<String> {
+    ci_yml
+        .lines()
+        .filter_map(|l| l.trim().strip_prefix("run: "))
```
dutiona: [Codex Review] [HIGH] This line scanner silently drops cargo gates written as a YAML block scalar: run: | followed by an indented cargo command, because it records only the pipe marker and ci_cmds remains non-empty from older steps. It also drops anonymous - run: cargo steps. I reproduced both as empty extraction results, so either common Actions form lets the docs drift while this test stays green. Parse workflow steps as YAML and add mutation cases for both forms.

tests/gate_matrix_parity.rs
```diff
@@ -0,0 +1,88 @@
...
+        .map(str::trim)
+        .filter(|c| c.starts_with("cargo "))
+        .filter(|c| !c.starts_with("cargo install "))
+        .filter(|c| !c.contains("--tests --examples"))
```
dutiona: [Codex Review] [MEDIUM] The MSRV exception is selected by command text globally, not by the msrv job. A non-MSRV gate containing --tests --examples is discarded too; passing run: cargo build -p memory-engine --tests --examples reproduced an empty extraction result. Scope the exception to jobs.msrv or to explicit job-qualified commands so unrelated gates cannot bypass parity.

tests/gate_matrix_parity.rs
```diff
@@ -0,0 +1,88 @@
...
+        .map(norm)
+        .collect()
+}
+
+/// The `cargo` lines from a doc's gate-matrix fenced block (the one carrying the
+/// `cargo fmt --all --check` anchor), with trailing `# comments` and padding stripped.
+fn doc_matrix_commands(doc: &str, doc_name: &str) -> BTreeSet<String> {
+    let block = doc
+        .split("```")
```
dutiona, line 116: [Codex Review] [MEDIUM] This does not identify the actual documented gate block; split plus find can select any prose or fenced chunk containing the anchor. I reproduced both failure modes: a preceding prose mention yields zero commands, while a separate example block can supply cargo test even when the real verification block omits it. Parse fenced blocks structurally and require exactly one canonical bash gate block.

memory-engine/lib/me-archive/src/search.rs
```diff
@@ -85,7 +85,7 @@ impl Ord for MinScored {
 /// (`engine/archive.rs`) exercises it directly across the crate boundary (Wave 2
 /// #816 / S4, sub-PR 3b).
 #[must_use]
-pub fn is_within_archive_dir(pak_path: &str) -> bool {
+pub(crate) fn is_within_archive_dir(pak_path: &str) -> bool {
```
dutiona, line 87: [Codex Review] [LOW] The visibility is narrowed here, but the immediately preceding rustdoc still says pub rather than pub(crate), describes direct facade reuse, and says the facade regression test exercises this helper across the crate boundary. Those claims are now false. The same removal also leaves me-storage/src/cold_storage.rs:7-9 claiming verify_pak exists and tests/archive_test.rs:282 naming its old Ok(false) arm.

dutiona (commented):
[Codex Review] 1 HIGH, 2 MEDIUM, 1 LOW. The #980 single-hash behavior, #979 helper reachability/fuzz seam, and #969 DIM sweep passed targeted verification. #991 is not yet a reliable parity guard: common YAML forms are silently omitted, the MSRV exception is not job-scoped, and Markdown block selection can inspect the wrong block; the API narrowing also leaves stale comments. Details inline. Verified: gate_matrix_parity 1/1, archive_test 7/7, me-archive 38/38, me-forget 20/20, detached fuzz build, and rustdoc broken/private-link gate.

tests/gate_matrix_parity.rs
dutiona: Fixed in aa16879. Rewrote ci_gate_commands as a YAML-subset walk of jobs.<job>.steps[] that strips a leading `- ` and handles named / nameless / block-scalar (`run: |`) / quoted forms. Locked by parser_catches_nameless_single_line_run + parser_catches_block_scalar_run + parser_catches_quoted_run.

tests/gate_matrix_parity.rs
dutiona: Fixed in aa16879. The MSRV exception is now scoped to `jobs.msrv` via job-tracking, not a global `--tests --examples` text match. parser_scopes_msrv_exclusion_to_the_job proves an unrelated --tests --examples gate in another job is NOT excluded.

tests/gate_matrix_parity.rs
dutiona: Fixed in aa16879. doc_matrix_commands now selects fenced code blocks structurally (odd-indexed ```-split chunks) and asserts exactly one carries the fmt anchor — a prose mention or a second example block can no longer substitute.

memory-engine/lib/me-archive/src/search.rs
dutiona: Fixed in aa16879. Updated is_within_archive_dir's rustdoc to the pub(crate)/intra-crate reality, and the stale verify_pak references in me-storage/cold_storage.rs and tests/archive_test.rs.

tests/gate_matrix_parity.rs
```diff
@@ -0,0 +1,88 @@
...
+    ci_yml
+        .lines()
+        .filter_map(|l| l.trim().strip_prefix("run: "))
+        .map(str::trim)
+        .filter(|c| c.starts_with("cargo "))
+        .filter(|c| !c.starts_with("cargo install "))
+        .filter(|c| !c.contains("--tests --examples"))
+        .map(norm)
+        .collect()
+}
```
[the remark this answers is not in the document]
dutiona: Fixed in aa16879 — the parser now handles `- run:` (plus block-scalar and quoted forms); a parser_catches_nameless_single_line_run mutation test locks it. Thanks.

tests/gate_matrix_parity.rs
```diff
@@ -0,0 +1,88 @@
...
+/// The `cargo` lines from a doc's gate-matrix fenced block (the one carrying the
+/// `cargo fmt --all --check` anchor), with trailing `# comments` and padding stripped.
+fn doc_matrix_commands(doc: &str, doc_name: &str) -> BTreeSet<String> {
+    let block = doc
+        .split("```")
+        .find(|b| b.contains("cargo fmt --all --check"))
+        .unwrap_or_else(|| {
+            panic!("{doc_name}: no gate-matrix bash block (anchor `cargo fmt --all --check`) found")
+        });
+    block
```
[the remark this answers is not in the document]
dutiona: Fixed in aa16879 — adopted the odd-index code-block selection you suggested, and additionally assert exactly one anchored gate block exists.

Issue #345, opened 2026-06-01
[super-qa] medium: Safety cap constant duplicated in dedup and cluster with inconsistent enforcement policy


dutiona:
> ⚠️ Generated by `/super-qa` 2026-06-01. Severity & details from agent analysis; **not independently source-verified**. Review before action.

**Safety cap constant duplicated in dedup and cluster with inconsistent enforcement policy**

- **ID:** `consolidation/maintainability-duplicate-safety-cap-constants`
- **Severity:** medium
- **Area:** consolidation · **Type:** refactor
- **Location:** `src/consolidation/dedup.rs:30, src/consolidation/cluster.rs:36`
- **Auto-fixable:** no

## Description
Current state: `MAX_DEDUP_FACTS: usize = 50_000` (dedup.rs:30) and `MAX_CLUSTER_FACTS: usize = 50_000` (cluster.rs:36) are identical values in two files. More importantly, they are enforced differently: the dedup cap returns a sentinel and blocks watermark advancement (skip-with-consequence); the cluster cap silently returns 0 (skip-with-no-consequences, existing summaries are deleted at line 53 after the cap check). A future developer changing one cap is unlikely to know about the other, and the two enforcement policies diverge silently. Target state: lift both constants to `consolidation/mod.rs` or a private `constants` submodule; name them `MAX_FACTS_FOR_PAIRWISE_OPS` or with distinct, clear names; document the policy difference explicitly. Refactoring pattern: Consolidate Duplicate Constants / Introduce Explaining Variable. Motivation: maintainability (single source of truth), readability (explicit policy difference). Scope: 2 files. Prerequisites: none. Risk: low.

## Suggested approach
// consolidation/mod.rs (or a private constants.rs)
/// Safety cap for O(N·M) dedup. Beyond this, dedup is skipped
/// and the watermark is NOT advanced (retried next run).
pub(super) const MAX_FACTS_FOR_DEDUP: usize = 50_000;
/// Safety cap for O(N²) clustering. Beyond this, clustering is skipped
/// (existing summaries are preserved, not deleted).
pub(super) const MAX_FACTS_FOR_CLUSTERING: usize = 50_000;
// Reference these in dedup.rs and cluster.rs via super::

## References
—

PR #998, opened 2026-07-21
refactor(core): Wave 2 collateral — single-source consolidation caps (#983) + prune MemoryCtx alignment (#970)

### LINKED ISSUE #970: refactor(forgetting): align me_forget::prune to the MemoryCtx L3 signature
## Context

Wave 2 (#816) S1 locked `MemoryCtx` (me-storage) as *"the universal capability handle every primitive receives"* — the uniform signature for all L3 primitive free functions (`me-storage/src/ctx.rs`).

**me-forget** (S3 sub-PR 2, #968) was a git-mv of a **pre-Wave-2** free function whose signature predates `MemoryCtx`:

```rust
pub async fn prune(storage: &Arc<dyn StorageBackend>, graph: &RwLock<MemoryGraph>,
                   policy: &ForgetPolicy, now: DateTime<Utc>) -> Result<(PruneStats, Vec<i64>)>
```

**me-resolve** (S3 sub-PR 3, first *fresh* `impl`-method→free-fn extraction) adopts `MemoryCtx` per the S1 lock (maintainer decision 2026-07-11), establishing the uniform template:

```rust
pub async fn resolve_conflict(ctx: MemoryCtx<'_>, graph: &RwLock<MemoryGraph>,
                              arbiter: &dyn ConflictArbiter, old_id: i64,
                              new_fact: &NewFact, now: DateTime<Utc>) -> Result<ConflictResolution>
```

This leaves me-forget as the odd sibling until aligned.

## Task

Convert `me_forget::prune` to take `ctx: MemoryCtx<'_>` instead of `storage: &Arc<dyn StorageBackend>` (embed_dim/read_only unused by prune today, but carried for uniformity — same as me-resolve). Update the facade delegate `MemoryEngine::forget` to call `me_forget::prune(self.mem_ctx(), &self.graph, policy, Utc::now())` via the `mem_ctx()` helper me-resolve introduces.

Behavior-preserving; no test changes beyond the call-site. Small, mechanical — best done right after me-resolve lands (so `mem_ctx()` exists) and before S4 multiplies the pattern.

Refs #816, #939. Follow-up from the me-resolve MemoryCtx decision.

### LINKED ISSUE #983: fix(consolidation): the 50k caps are duplicated across me-consolidate and me-backend-sqlite with no agreement test
## Summary

The consolidation **50k caps** are duplicated across two crates with **no test asserting they agree**. If they ever drift apart, the pipeline and the backend silently disagree about whether a run was capped — a correctness bug with no failing gate.

Surfaced by the adversarial review of #982 (the `me-consolidate` carve). **Pre-existing** — introduced in S2 sub-PR 2b, present at base `227fd74` — so not folded into that PR, per the collateral rule.

## The duplication

| constant | location |
|---|---|
| `MAX_FACTS_FOR_DEDUP` / `MAX_FACTS_FOR_CLUSTERING` = `50_000` | `me-consolidate/src/pipeline.rs:63,76` |
| the same two caps | `me-backend-sqlite/src/consolidation/mod.rs:47,52` |

Before the carve these lived in one crate. They now sit in **two separate source trees** with nothing linking them at compile time.

## Failure mode

`load_snapshot` (backend) uses its copy to decide whether to set `over_both_caps`; `compute_plan_capped` (pipeline) uses its copy to decide whether to actually skip the pass. If the backend's cap ever drifts **below** the pipeline's:

- the backend marks a snapshot `over_both_caps` (so it stops loading the full set), while
- the pipeline believes it is under the cap and proceeds as though it has complete data.

The result is a **silent skip/no-skip mismatch** — consolidation running on a truncated view while believing it is complete. No test catches this today: nothing asserts the two constants are equal.

## Suggested fix

Single-source them. Options, cheapest first:

1. **Hoist to L0** — put the caps in `me-types` (they are policy constants, not backend or algorithm internals) and have both crates read the one definition. Removes the possibility of drift entirely.
2. **Pin with a test** — if the duplication is deliberate (e.g. the backend wants a cap it can tune independently), add an equality assertion so a drift is loud:
   ```rust
   #[test]
   fn consolidation_caps_agree_across_crates() {
       assert_eq!(me_consolidate::MAX_FACTS_FOR_DEDUP, me_backend_sqlite::consolidation::MAX_FACTS_FOR_DEDUP);
       assert_eq!(me_consolidate::MAX_FACTS_FOR_CLUSTERING, me_backend_sqlite::consolidation::MAX_FACTS_FOR_CLUSTERING);
   }
   ```
   (Requires making them `pub`; the facade or a test crate can host the assertion, since me-consolidate must not take a production dep on the backend.)

Option 1 is preferred — it makes the invariant unrepresentable rather than merely tested. This is the same lesson as #976/#978's `.pak` schema constant: a value that two sides must agree on wants **one** home, ideally in a layer both can reach.

Refs #982, #816

dutiona:
## Wave 2 collateral: two locked invariants made real (#983 + #970)

Both commits share a theme surfaced by the Wave 2 (#816) carve: *a locked invariant that was actually just a claim.* Disjoint file sets, behavior-preserving, one atomic commit each. Full external + subagent review applied (3 rounds; see the Review section).

### `fix(consolidation): single-source the 50k consolidation caps in L0 me_types::limits (#983)`

`MAX_FACTS_FOR_DEDUP` / `MAX_FACTS_FOR_CLUSTERING` (both `50_000`) were duplicated across `me-consolidate` (lock-free compute half) and `me-backend-sqlite` (below-the-seam load half) with **nothing linking the two copies at compile time**. Silent-drift hazard, no failing gate: if the backend cap ever dropped below the pipeline cap, the backend would mark a snapshot `over_both_caps` and truncate the load while the pipeline believed it had the full active set — consolidation running on a partial view while reporting success.

**Fix:** hoist both constants to **L0 `me_types::limits`** (which both consumers already depend on — no new dependency edge; `me-types` is the acyclic leaf). Cross-crate agreement is now *unrepresentable* rather than merely tested. The two caps stay separately named/documented so the deliberate #345 policy difference survives (dedup skip blocks the watermark to retry; cluster skip is a no-op). No agreement test — Option 1 of the issue makes the test unnecessary (and an equality test would wrongly freeze the two equal).

**Why `limits`, not `types::consolidation`:** the facade blanket-re-exports `pub use me_types::types`, so homing the caps there would leak them onto the facade's public API as `memory_engine::types::consolidation::MAX_FACTS_FOR_*` (a Codex review finding — no gate catches a newly-reachable public path). `me_types::limits` is facade `pub(crate)`, so the caps are workspace-internal policy defaults. The `limits` module + crate docs are updated to cover the new consolidation-caps family alongside the ingest payload caps.

### `refactor(forgetting): align me_forget::prune to the MemoryCtx L3 signature (#970)`

`me_forget::prune` was a `git mv` of a pre-Wave-2 free function that still took `storage: &Arc<dyn StorageBackend>` directly — the last L3 primitive off the S1-locked `MemoryCtx` template, in **both** signature and body. Converted to `ctx: MemoryCtx<'_>` (reads `ctx.storage`) **and** added `ctx.ensure_open()?` at entry (a review finding: prune was the lone L3 primitive not self-guarding the #742 fence — every sibling does). `embed_dim`/`read_only` carried for uniformity. Facade delegate passes `self.mem_ctx()`, keeping its own `ensure_open()` as archive-style defence-in-depth. `ctx.ensure_writable()` deliberately **not** added — that systemic write-gate question is **#972**.

**Behavior-preserving** (the facade already fenced `forget()`; the happy-path tests use an always-open fence). A dedicated regression test **`prune_fenced_by_reopen_required`** witnesses the new guard with a non-zero fence and was **mutation-proven**: it passes with `ctx.ensure_open()?` present and *fails* when that line is removed, so the guard cannot silently regress (a Codex review finding: the guard needed its own test).

### Review — gate: Full (3 rounds)

Reviewers that ran: **Codex (external, tmux)** + 3 diverse-lens subagents (correctness / architecture / docs). **agy degraded** (print-mode timeout — a valid Codex-only degraded-Full per the arc's playbook). Findings, all verified against source and fixed in-branch:
- **R1** — [MEDIUM] Codex: cap facade-leak → moved to `limits`. [MEDIUM] architecture: prune's missing entry-fence → `ensure_open()` added. (correctness + docs subagents: LGTM.)
- **R2** — leak fix confirmed closed via a consumer compile probe. [MEDIUM] the new guard had no regression test → added + mutation-proven. [LOW] stale `limits` module docs → updated.
- **R3** — round-2 CLOSED; no new BLOCKER/HIGH/MEDIUM. Remaining LOW items (PR-body/thread hygiene) addressed.

### Verification

Full 12-line gate matrix, unpiped, all green — `fmt` · `clippy --all-features -D warnings` · `build` ×2 · `test --workspace --all-features` (**1983 passed; anti-#903 1982 +1, the +1 being exactly the new `prune_fenced_by_reopen_required` regression witness — accounted by name**) · facade-alone ×2 · `doc` ×2 · `cargo deny` · `cargo +nightly fuzz build` (#993 facade-API gate). Re-run green after each review round.

Closes #983
Closes #970


memory-engine/lib/me-types/src/types/consolidation.rs
```diff
@@ -13,6 +13,35 @@ use chrono::{DateTime, Utc};
...
+///
+/// Homed here in L0 (#983) so the lock-free compute half (`me-consolidate`) and the
+/// below-the-seam load half (`me-backend-sqlite`'s `load_snapshot`) read **one**
+/// definition. Before the Wave 2 (#816) carve these caps lived in a single crate;
+/// splitting them into two source trees left the two copies with nothing linking them
+/// at compile time, so a silent drift (backend cap below pipeline cap ⇒ the backend
+/// truncates the load while the pipeline believes it has the full set) had no failing
+/// gate. Single-sourcing them makes that drift *unrepresentable* rather than merely
+/// tested.
+pub const MAX_FACTS_FOR_DEDUP: usize = 50_000;
```
dutiona: [Codex Review] [MEDIUM] These `pub const` definitions are newly consumer-visible through the facade: `memory-engine` does `pub use me_types::types` and `types::consolidation` is public, so consumers can now name `memory_engine::types::consolidation::MAX_FACTS_FOR_*`. That exposes implementation-policy defaults as facade API even though the backend import comment calls them private/internal defaults. Keep the shared definitions on a public cross-crate path outside the blanket-re-exported `types` tree, or explicitly accept and document the additive facade API.

dutiona (commented):
[Codex Review] One [MEDIUM] finding: the L0 hoist preserves values, call sites, and DAG direction, but placing the `pub const` definitions under the facade-wide `types` re-export leaks both policy defaults as new `memory_engine::types::consolidation::*` API. No blocker/high finding; no third live cap copy, stale old-signature reference, or removed pre-existing gate was found. `cargo fmt --all --check`, `git diff --check a1edc2f..HEAD`, and targeted `cargo tree` checks passed. The unrelated `add-to-main` job is red because its project action reports `Bad credentials`; the workflow-dispatch-only CI matrix has not run on this PR.

memory-engine/lib/me-types/src/types/consolidation.rs
dutiona: Fixed in e00ac27 (force-pushed). The two caps were moved out of `me_types::types::consolidation` into `me_types::limits`, which the facade re-exports as `pub(crate) use me_types::limits` — so they are cross-crate visible to me-consolidate + me-backend-sqlite but no longer reachable on the facade public API. Verified by a consumer compile probe (round-2 review): `memory_engine::types::consolidation::MAX_FACTS_FOR_DEDUP` is now unresolved while `memory_engine::types::consolidation::Snapshot` still resolves as the control. Good catch — no gate catches a newly-reachable public path.

dutiona (commented):
[super-review summary] Full gate, 3 rounds. Reviewers: Codex (external) + 3 diverse-lens subagents (correctness/architecture/docs); agy degraded (print-mode timeout). Findings, all fixed in-branch and re-gated:
- R1: facade-leak of the hoisted caps (Codex) → moved to `me_types::limits` (facade `pub(crate)`); prune missing the #742 entry-fence (architecture) → `ctx.ensure_open()?` added.
- R2: leak fix confirmed closed via consumer compile probe; the new guard lacked a regression test → added `prune_fenced_by_reopen_required` (mutation-proven: fails if the guard is removed); stale `limits` module docs → updated.
- R3: all prior findings CLOSED; no new BLOCKER/HIGH/MEDIUM.

Gate matrix green after each round; anti-#903 = 1983 (1982 + the one named regression test).

PR #1007, opened 2026-07-22
refactor(consolidation): batch importance_score updates in apply_dedup (N+1 → bulk) (#895)

### LINKED ISSUE #895: refactor(consolidation): batch importance_score updates in apply_dedup (N+1 → bulk)
Found during the #255 area:forgetting sweep (adversarial review of PR #885 / #392).

`apply_dedup` still performs the exact N+1 pattern that #392 eliminated in `prune_atomic`:

```rust
// memory-engine/lib/memory-engine/src/consolidation/dedup.rs:259-261
for &(id, score) in &computed.importance_score_updates {
    fact_store.update_importance_score(id, score)?;
}
```

Once PR #885 merges, `FactStore::update_importance_scores_bulk(&[(i64, f64)])` (single `json_each` UPDATE) exists and is a drop-in replacement here. The dedup loop runs inside `apply_dedup`'s transaction, so the bulk swap is behavior-equivalent (same atomicity), exactly as in the prune path.

This was intentionally **not** folded into PR #885 (which is scoped to #392, the prune path) — filing separately per the one-logical-concern-per-issue contract.

Part of #255 (collateral). Related: #392.

dutiona:
## What

`apply_dedup` materialized the inherited `importance_score` values with an N+1 per-row loop of `update_importance_score` (one prepared-statement exec + one B-tree write per survivor). This routes them through the existing `FactStore::update_importance_scores_bulk` — a single `json_each`-driven `UPDATE facts SET importance_score = s.value FROM json_each(?1) AS s WHERE facts.id = CAST(s.key AS INTEGER)`. It is the exact swap #392 made in the prune path (`prune_atomic`).

## Why behavior-preserving

- The bulk call stays **inside `apply_dedup`'s transaction**, so atomicity is unchanged — the whole batch commits or rolls back together, exactly as the loop did.
- The only observable deltas are on inputs a real caller cannot produce:
  - A **missing id** is now silently skipped rather than surfaced as `NotFound` (matching the prune path; facts are soft-deleted, so a missing importance-score row cannot occur for a live plan).
  - A **non-finite score** is rejected as a typed `MemoryError::Conflict` before any row is touched. `compute_dedup` only carries forward already-finite stored scores, so this is unreachable defense-in-depth.
- The `# Errors` doc on `apply_dedup` is updated to state this precisely. The separate `update_base_importance` loop is left untouched (out of scope — #895 targets only the `importance_score` N+1; no bulk base-importance helper exists yet — see PR notes for the collateral flag).

## Test

Adds `apply_dedup_bulk_updates_all_survivor_scores`: two mutually-dissimilar duplicate groups, each producing a distinct survivor that inherits its loser's higher score, so `importance_score_updates` carries **two** entries and both must land in the single bulk UPDATE. The existing `update_importance_scores_bulk` unit tests in `me-backend-sqlite` already cover the empty-slice no-op and the non-finite-rejection paths.

## Verification

Full local gate matrix run unpiped (CI is `workflow_dispatch`-only), all green: fmt, clippy `--workspace --all-targets --all-features -D warnings`, build (default + all-features), test `--workspace --all-features` (1990 passed / 0 failed / 81 ignored, +1 test vs main), facade `cargo check` (default + no-default), rustdoc (default + all-features), `cargo deny check`, `cargo +nightly fuzz build`.

Classification: Refactor (user-confirmed)

Closes #895

### DIFF
--- a/crates/me-backend-sqlite/src/consolidation/dedup.rs
+++ b/crates/me-backend-sqlite/src/consolidation/dedup.rs
@@ -43,9 +43,13 @@ use crate::store::facts::FactStore;
 ///
 /// # Errors
 ///
-/// Returns `MemoryError::Storage` on SQL failure, or `MemoryError::NotFound` from an
-/// importance update if a fact row is missing (facts are soft-deleted, so this does not
-/// fire for a merely-expired row).
+/// Returns `MemoryError::Storage` on SQL failure, or `MemoryError::NotFound` from a
+/// **base-importance** update if a fact row is missing (facts are soft-deleted, so this
+/// does not fire for a merely-expired row). The `importance_score` writes go through the
+/// bulk `update_importance_scores_bulk` (#895): a missing id is silently skipped rather
+/// than surfaced as `NotFound` (matching the prune path), and a non-finite score — which
+/// `compute_dedup` never produces, since it only carries forward already-finite stored
+/// scores — is rejected as `MemoryError::Conflict` before any row is touched.
 pub fn apply_dedup(
     tx: &Transaction,
     embed_dim: usize,
@@ -58,9 +62,11 @@ pub fn apply_dedup(
     for &(id, importance) in &computed.importance_updates {
         fact_store.update_base_importance(id, importance)?;
     }
-    for &(id, score) in &computed.importance_score_updates {
-        fact_store.update_importance_score(id, score)?;
-    }
+    // Materialize the inherited importance scores in one bulk UPDATE (#895) instead of an
+    // N+1 per-row loop — the same swap #392 made in the prune path (`prune_atomic`). We are
+    // inside the caller's transaction, so atomicity is unchanged: the whole batch commits or
+    // rolls back together, exactly as the loop did.
+    fact_store.update_importance_scores_bulk(&computed.importance_score_updates)?;
 
     // Snapshot each distinct survivor's liveness BEFORE expiring anything, so a survivor
     // this call itself expires (it may also be a loser in a merge chain) is not confused
--- a/crates/me-consolidate/src/pipeline/dedup.rs
+++ b/crates/me-consolidate/src/pipeline/dedup.rs
@@ -765,6 +765,59 @@ mod tests {
         );
     }
 
+    /// #895: `apply_dedup` writes the inherited `importance_score`s through a single
+    /// bulk `update_importance_scores_bulk` UPDATE instead of an N+1 per-row loop. This
+    /// exercises the bulk path with **more than one** score write in a single apply — two
+    /// mutually-dissimilar duplicate groups, each producing a distinct survivor that
+    /// inherits its loser's higher score — so `importance_score_updates` carries two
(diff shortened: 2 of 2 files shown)


dutiona:
**Review gate: Adaptive → Light floor, degraded to inline self-review (recorded decision)**

This ran in a non-interactive, headless subagent context with no `Agent`/subagent-spawn tool and no interactive TTY for the Codex-tmux / agy-resume reviewer loop. Per super-review's degradation table (Agent absent → perform the review inline; non-interactive fan-out → Adaptive gate, Light floor), the review was conducted as a rigorous inline critical self-review. This is legitimate for lighter-work Refactor (not Feature/Security/multi-file). Degradation stated loudly: no external `[Codex Review]`/`[Antigravity Review]` voices ran.

Self-review findings — LGTM, no blocker/high/medium:
- Faithful mirror of the #392 prune-path precedent (`prune_atomic` did the identical `update_importance_scores_bulk` swap).
- **Atomicity preserved:** the bulk `UPDATE` executes on the `&Transaction`-backed `FactStore` inside `apply_dedup`'s tx — whole batch commits/rolls back together, exactly as the loop did.
- Empty `importance_score_updates` → helper's early `Ok(())` == the zero-iteration loop (no-op).
- `running_scores` is a `HashMap`, so the collected `(id, score)` pairs have unique ids — no dup-key/last-wins divergence from the loop.
- `update_base_importance` writes only the `importance` column; the bulk helper writes only `importance_score` — independent columns, so ordering between the base loop and the bulk write is immaterial.
- Two behavior deltas, both unreachable for real callers and now documented in `# Errors`: a missing id → silent skip (was `NotFound`), matching prune; a non-finite score → typed `Conflict` before any row is touched (`compute_dedup` only forwards already-finite stored scores).
- Added test `apply_dedup_bulk_updates_all_survivor_scores` drives N=2 through the bulk path and passes.

dutiona:
[Antigravity Review] Cross-model review (agy `Gemini 3.1 Pro high`, one-shot) — **LGTM**. Correct, transactionally safe, well-covered.

- **Happy path**: behavior-preserving — same `(id, score)` set, same final `importance_score` values.
- **Atomicity verified**: the bulk `update_importance_scores_bulk` executes on the borrowed `&Transaction` (`self.conn`) without an internal commit, so it joins the caller's transaction — all-or-nothing preserved, rolls back with the outer tx on failure.
- **Behavior deltas [LOW, accepted + documented]**: (a) silent-skip on a missing id (vs the loop's `NotFound`) is unreachable in practice (`compute_dedup` only emits existing ids; soft-deleted rows still update) and matches the #392 prune precedent; (b) non-finite score → `Conflict` (was a DB `NOT NULL` `Storage` error) is strictly better design (policy rejection vs constraint failure). The `# Errors` update accurately describes both.
- **Test**: `apply_dedup_bulk_updates_all_survivor_scores` is a valid N=2 witness guarding the bulk SQL against mis-keyed JSON / partial writes; edge cases are structurally unreachable from `dedup()`, so integration coverage is sufficient.

